PR #18, opened 2021-09-02
Created initial proposal

jamesefhawkins:
This is a test of our new communication style for an issue @kpthatsme raised.

It's a very early product concept worth discussing/concluding on.

### DIFF
--- a/autocapture.md
+++ b/autocapture.md
@@ -1,0 +1,36 @@
+This is a rough proposal that fits into the category of complex product decisions. I will polish this markdown as a PR.
+
+# Proposal: stop supporting autocapture
+
+We currently support autocapture of front end events.
+
+This helps us with:
+
+* Users can get events in more quickly
+* It captures lots of useful pageview data
+* Perhaps generates initial sign ups who we retain for _other_ reasons
+
+Hypothesis for why this is hurting us:
+
+Complexity.
+
+* It generates lots of auto labelled events, which will lead to confusing experiences for users or incorrect decision-making
+* It doesn't solve the whole problem - you still can't get email addresses for example with it.
+  * This makes the solution _more_ confusing: you now need to do manual events AND train users to use autocapture too
+* It makes mobile a 2nd class citizen, since it isn't supported with autocapture
+* Technical debt maintaining it as a feature
+
+# Should we investigate first?
+
+There will be two outcomes of a completed investigation:
+
+1. Users love it and it helps retention
+2. Whilst it may/may not help sign ups, it reduces retention = discard
+
+With 2, we would discard it, so we're left with questioning what would happen with 1. Even if users love it, I think it could be just a substitute for a lack of guidance over which events should come in manually and we should try onboarding people with more guidance on this. It is rational to test this idea out.
+
+# What about the opposite: couldn't we double down on it?
+
+We could go much further with autocapture so it is a complete solution.
+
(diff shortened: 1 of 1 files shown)


autocapture.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Proposal: stop supporting autocapture
+
+We currently support autocapture of front end events.
+
+This helps us with:
+
+* Users can get events in more quickly
+* It captures lots of useful pageview data
+* Perhaps generates initial sign ups who we retain for _other_ reasons
```
mariusandra, line 11: Also:
- Autocaptured clicks power the toolbar heatmap.

autocapture.md
```diff
@@ -0,0 +1,36 @@
...
+
+Hypothesis for why this is hurting us:
+
+Complexity.
+
+* It generates lots of auto labelled events, which will lead to confusing experiences for users or incorrect decision-making
+* It doesn't solve the whole problem - you still can't get email addresses for example with it.
+  * This makes the solution _more_ confusing: you now need to do manual events AND train users to use autocapture too
+* It makes mobile a 2nd class citizen, since it isn't supported with autocapture
```
mariusandra, line 20: Support for React Native is possible, yet has been deprioritized. Heap does it though, and we could do the same. Context [here](https://github.com/PostHog/posthog/issues/1174) and [here](https://heap.io/blog/engineering/how-we-leveraged-asts-and-babel-to-capture-everything-on-react-native-apps).

autocapture.md
```diff
@@ -0,0 +1,36 @@
+This is a rough proposal that fits into the category of complex product decisions. I will polish this markdown as a PR.
+
+# Proposal: stop supporting autocapture
+
+We currently support autocapture of front end events.
+
+This helps us with:
+
+* Users can get events in more quickly
+* It captures lots of useful pageview data
```
mariusandra, line 10: Don't underestimate how valuable this is for e.g. documentation sites like kea. With no setup I can see what links are being clicked and what aren't.

autocapture.md
```diff
@@ -0,0 +1,36 @@
...
+1. Users love it and it helps retention
+2. Whilst it may/may not help sign ups, it reduces retention = discard
+
+With 2, we would discard it, so we're left with questioning what would happen with 1. Even if users love it, I think it could be just a substitute for a lack of guidance over which events should come in manually and we should try onboarding people with more guidance on this. It is rational to test this idea out.
+
+# What about the opposite: couldn't we double down on it?
+
+We could go much further with autocapture so it is a complete solution.
+
+We'd need to capture backend events and support mobile. I suspect the reduction in workload setting up events will be reaplced by workloading mapping the underlying database to PostHog. I suspect this will just move the technical work from setting up backend event tracking to tagging what backend events mean in our front end.
```
mariusandra, line 36: Other ideas:
- Could we provide a toggle to enable/disable autocapture in the UI, just like for session recordings. Easier to turn off if not needed or outgrown.
- Autocapture with the allow-listed `data-attr` tracking as we have now can provide pretty solid instrumentation for very little effort.
- What about renaming "autocapture" to "capture clicks"? Explicit is better than implicit?

Twixes: It's great how simple it is to get _some_ interesting data into PostHog, even if it's not as valuable as custom events. I also think the name fits the feature very well as is ("capture clicks" would be a bit misleading actually).
IMO what would solve the issue being discussed is that ability to disable autocapture session-recording style. It could maybe be shown in onboarding too (if we bring onboarding back) – on by default, but easy to disable at any moment.

jamesefhawkins:
I don't think this was the right thing to do - we have new product priorities. Hence closing.

PR #19, opened 2021-09-06
Create Request for Comments template

Twixes:
This is a go at creating a Request for Comments formula for proposals inside the company, based on https://github.com/PostHog/posthog.com/pull/1881 (and @macobo's comment https://github.com/PostHog/posthog.com/pull/1881#discussion_r702612585).


rfc/$template.md
```diff
@@ -0,0 +1,18 @@
+# RFC: *topic of the proposal*
```
macobo, line 1: Suggestion: Add YYYY-MM-DD to the file name template as well - makes it easier to find recent RFCs

macobo (commented):
Why not posthog.com :thinking:

Twixes:
@jamesefhawkins I lightly transformed https://github.com/PostHog/meta/pull/18 into an RFC according to this template, as an example.

rfc/$template.md
Twixes: Great idea, time-sortability FTW

Twixes:
I think we could put this posthog.com, but that would require a section separate to both the Docs (which are user-facing) and the Handbook (which is mostly for internal use too, but kept constantly relevant, while these RFCs by nature would be only archival after implementation).

macobo (approved):
:heart: We can always move things later if needed :)

timgl:
Would be in favour of calling it something other than RFC. I don't think it's super well known outside of bigger companies and it's another acronym to learn. Either we should spell it out (Request for Proposal) or just call it "Proposal".

piemets:
@Twixes I'd recommend adding an FAQs section at the end (which you I think have kind of done anyway with the first proposal).

> Would be in favour of calling it something other than RFC. I don't think it's super well known outside of bigger companies and it's another acronym to learn. Either we should spell it out (Request for Proposal) or just call it "Proposal".

Agree - using acronyms is not in line with our [writing style](https://posthog.com/handbook/company/communication). I had to look up what RFC is (as I'd only heard of it in the Rugby Football Club context...)

Twixes:
Makes sense @piemets @timgl Agreed RFC may be too engineer-y, let's go for Proposal then.

Twixes:
Hm, currently "Proposal" is one of the sections of the template though. Maybe we could just use the full name "Request for Comments" and avoid the acronym as much as possible? I do like how "Request for Comments" conveys the propose-discuss-finalize nature of this.
(BTW we should be discussing this in an inline comment thread instead of loosely here, for instance on the first line containing "RFC:")

piemets (approved):

PR #30, opened 2021-10-05
RFC: 2021-10-05-group-analytics

macobo:
Take a look, share your thoughts!

### DIFF
--- a/requests-for-comments/2021-10-05-group-analytics.md
+++ b/requests-for-comments/2021-10-05-group-analytics.md
@@ -1,0 +1,196 @@
+# Request for Comments: *topic of the proposal*
+
+- **Created:** 2021-10-05
+- **Owner:** @macobo
+- **GitHub issues:** https://github.com/PostHog/posthog/issues/2247
+
+> While commenting on this, please comment on specific lines if possible, to thread the discussion.
+
+## Overview
+
+This RFC proposes a solution which would allow PostHog users to do analysis not only by users but also by groups (e.g. organizations, projects).
+
+Let's make it possible 
+
+## Background
+B2B products don’t deal directly with users, but with other companies and hence have more complicated needs around what kinds of metrics are most important for them.
+
+
+Lack of this feature has been a major driver for users wanting data exports in different analytics products, forcing users to roll their own analytics on top of data warehouses to get at their key metrics.
+
+### Example usecases
+
+
+- Show me how many active “organizations” there are
+- Show only pageviews from organizations who are paying us (filtering by group properties)
+- Have an overview of groups and their relations with users
+- Feature flags: roll out feature group-by-group
+
+For PostHog the product, the 3 groups we would use are:
+
+- Instances (e.g. cloud, posthog.someproduct.org, …)
+- Organizations
+- Projects
+
+## Semantics & Constraints
+
+### 1. Users can belong to multiple groups
+
+Some examples:
+
+1. Contractor type work (e.g. for farm management agronomists work for multiple farms at once)
+2. Billing happens on a different level (e.g. our projects vs organization)
+3. Users clients are global companies with a lot of local “branches” where execs need access to everything
+4. Internal users have access to debug issues on multiple accounts
+### 2. User is active in only “group” at any point in time
+
+Whenever an event is made, we can (and should) make the assumption that it belongs to a single group (e.g. organization). This is needed to avoid “double-counting” in stats.
+
+Example: If Yakko has been added to 20 different organizations and is active, we don’t want to count all of them as active.
+
+### 3. Limited amount of groups
+
+From speaking with clients, their internal amount of different “groups” is limited. 
+
+This constraint allows us to implement things slightly more efficiently and have a nicer UX compared to unbounded amount of groups.
+
+### 4. Feature flags in groups should be consistent
+
+For the users who are using feature flags, they want to ensure a consistent experience within a group during rollout. 
+
+This is to avoid situations where some users see a new feature and some do not.
+
+This also factors into their later analysis - for example, they’d like to treat two different “organizations” both as a single datapoint even if one has 1500 users and the other has 1.
+
+## Proposal
+
+### Groups are event-level
+
+Users can define up to N (5?) group types. For us these would be `instance` , `project` and `organization`.
+
+Each event we ingest can have one group per group type associated. 
+
+We will have N new columns in the `events` table, containing the group ids - first one for the first group type of the project, second for the second one etc. 
+
+### Group properties API
+
+We will expose a new API for creating and adding properties groups. This API can be exposed on the client level, e.g.
+
+```python
+    import posthog
+    
+    posthog.groups.add_properties('organization', 'org:5', { name: 'Swedish Bakery', subscription_level: 'gold' })
+```
+
+There will not be any functionality for merging two groups, as that behavior is user-specific.
+
+### Capture API
+
+Capturing events in client libraries will allow users to pass group metadata alongside with an event.
+
+Here’s how it would look like in python:
+
+```python
+    import posthog
+    
+    posthog.capture(
+      user_id,
+      'purchase'
+      { plan: 'silver' },
+      groups={
+        'organization': 'org:6'
+      }
+    )
+```
+
+`groups` metadata will be sent to us under event properties under new the `$groups` key.
+
+Our data ingestion in plugin-server will perform the neccessary mappings from group ids to column indexes or other keys as needed.
+
+
+### In-App: filtering by group properties in insights
+
+In-app, users will be able to do analytics by groups and filter by them.
+
+To give an example:
+
+- Throughout insights, we’ll allow filtering insights with group properties. This will be implemented as a new tab per group type in the filters modal.
+- In trends, users can also select “active organizations” after first capturing events for an “organization” 
+- In funnels, users can select to analyze by a group type (e.g. “instance”). Different users within the group can have completed different steps of the funnel.
+- In retention, users can select “organization” over “user” for the unit to analyze retention by
+- And so on…
+
+
+### In-App: profile pages for groups
+
+We’ll also expose pages for different groups in-app, with:
+
+
+1. Listing all your different group types
+2. Listing and allowing searching all different groups of a type (e.g. organizations)
+3. Group page with
+    1. Group properties
+    2. Events from that group
+    3. Related users and groups (calculated via users’ other events)
+
+Initially, the design and behavior will be taken from the person pages and iterated later on.
+
+
+### Autocapture & posthog-js
+
+To make autocapture work with groups and to make posthog-js usage a bit simpler, we can provide the following API:
+
+```js
+    // Normal posthog usage
+    posthog.identify(userId)
+    // New proposed method, to be called right after .identify
+    posthog.group('instance', 'posthog.yourapp.com', { instance_prop: 5})
+```
+
+After calling this method, we will automatically associate events from that session with that group until either:
+
+- `posthog.group` is called with the same group type
+- `posthog.reset()` is called.
+### Feature flags
+
+Users should be able to add group-level property filters to feature flags and be able to ensure rollouts at a group level (?).
+
+However feature flags will not be part of the initial scope of the project, see below.
+
+## Open questions
+### What pricing tier this belongs to
+
+Does this sound like a collaboration-related feature we should immediately upsell? This is what mixpanel and amplitude do.
+
+### Feature flags and posthog-js timing
+
+Feature flags should work with group filters eventually. One tricky part there is behavior of posthog-js and ensuring we “know” what groups are currently active at the time of the request.
+
+### How would “managing” groups go, if users manage to mess up?
+
+TBD, can focus on this after the initial MVP
+
+## Other details
+
+### Initial scope
+
+This project is quite large. Rather than update all of our client libraries, ingestion, queries, feature flags and insights pages at once, the plan would be to do a gradual rollout in phases, gathering feedback from stakeholders at each step.
+
+The first step would be an internal MVP, where we can experiment with the feature ourselves behind a feature flag.
+
+Some tasks involve implementing new logic in posthog-js and posthog-python client libraries, get queries working for trends/funnels, implement very simple group profile pages.
+
+Work on feature flags specifically will be done later due to the extra complexity involved in making sure feature flags would work in posthog-js as expected.
+
+### Competitor solutions:
+
+Two other analytics products advocate group analytics:
+
+- Mixpanel https://mixpanel.com/group-analytics/
+- Amplitude https://help.amplitude.com/hc/en-us/articles/115001765532-Accounts
+
+The two approaches are different: Mixpanels groups are user-level, while Amplitudes approach is similar to this proposal with groups being on events.
+
+Other resources:
+
+- https://www.avo.app/docs/data-design/groups


requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Other details
+
+### Initial scope
+
+This project is quite large. Rather than update all of our client libraries, ingestion, queries, feature flags and insights pages at once, the plan would be to do a gradual rollout in phases, gathering feedback from stakeholders at each step.
+
+The first step would be an internal MVP, where we can experiment with the feature ourselves behind a feature flag.
+
+Some tasks involve implementing new logic in posthog-js and posthog-python client libraries, get queries working for trends/funnels, implement very simple group profile pages.
```
marcushyett-ph, line 181: It would be great if you could be a bit more specific about what you think will be possible / not possible with the MVP so we can make sure we can get the most value out of it to validate it effectively.

Would be keen to see if the MVP meets the basic needs of the growth team so we can validate on stuff we're working on right away, @kpthatsme @samwinslow?

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,196 @@
...
+### Feature flags
+
+Users should be able to add group-level property filters to feature flags and be able to ensure rollouts at a group level (?).
+
+However feature flags will not be part of the initial scope of the project, see below.
+
+## Open questions
+### What pricing tier this belongs to
+
+Does this sound like a collaboration-related feature we should immediately upsell? This is what mixpanel and amplitude do.
```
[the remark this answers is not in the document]
timgl: Agree

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,196 @@
...
+Here’s how it would look like in python:
+
+```python
+    import posthog
+    
+    posthog.capture(
+      user_id,
+      'purchase'
+      { plan: 'silver' },
+      groups={
```
timgl, line 100: If we make this a `$groups` property in properties instead we could have this live for all libraries instantly. Downside is less strong typing (though we could build that in)

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Proposal
+
+### Groups are event-level
+
+Users can define up to N (5?) group types. For us these would be `instance` , `project` and `organization`.
+
+Each event we ingest can have one group per group type associated. 
+
+We will have N new columns in the `events` table, containing the group ids - first one for the first group type of the project, second for the second one etc.
```
timgl, line 73: I suspect you've thought long & hard about why we're doing this on the events table and not on persons. Would be nice to understand why you've come to this solution over doing it person based.

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,196 @@
...
+
+This is to avoid situations where some users see a new feature and some do not.
+
+This also factors into their later analysis - for example, they’d like to treat two different “organizations” both as a single datapoint even if one has 1500 users and the other has 1.
+
+## Proposal
+
+### Groups are event-level
+
+Users can define up to N (5?) group types. For us these would be `instance` , `project` and `organization`.
```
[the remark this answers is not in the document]
macobo: To say 15 - very little. To 100 or above or unbounded it would get dicy - we'd likely want to rethink the schema. It would also be an indication of an unknown usecase.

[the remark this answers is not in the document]
marcushyett-ph: Yep - I can't imagine currently a use-case that would require >10 group types - from what we've come across.

requests-for-comments/2021-10-05-group-analytics.md
macobo: We'd continue working on the feature even after this sprint so want to avoid going too low level here - it's really not helpful. The rough order is:

1. Ingestion + js/py client library changes
2. Trends/funnels
3. Groups pages
4. Other insights modules

A bit later:
5. Other client libraries
6. Feature flags

requests-for-comments/2021-10-05-group-analytics.md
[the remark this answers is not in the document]
macobo: Thanks, I missed that - https://github.com/PostHog/company-internal/issues/417

requests-for-comments/2021-10-05-group-analytics.md
macobo: This is a fair point. However:
1. want to keep the client API as clear as possible in python/elsewhere
2. we need to also add the identify APIs for the groups to libraries anyways, tiny change in addition to that

requests-for-comments/2021-10-05-group-analytics.md
macobo: I hoped it was already clear based on the constraints but added a section at the end under `Other details`.

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,196 @@
...
+
+
+### In-App: filtering by group properties in insights
+
+In-app, users will be able to do analytics by groups and filter by them.
+
+To give an example:
+
+- Throughout insights, we’ll allow filtering insights with group properties. This will be implemented as a new tab per group type in the filters modal.
+- In trends, users can also select “active organizations” after first capturing events for an “organization”
```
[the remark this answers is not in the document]
macobo: Added a clarification

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,196 @@
...
+4. Internal users have access to debug issues on multiple accounts
+### 2. User is active in only “group” at any point in time
+
+Whenever an event is made, we can (and should) make the assumption that it belongs to a single group (e.g. organization). This is needed to avoid “double-counting” in stats.
+
+Example: If Yakko has been added to 20 different organizations and is active, we don’t want to count all of them as active.
+
+### 3. Limited amount of groups
+
+From speaking with clients, their internal amount of different “groups” is limited.
```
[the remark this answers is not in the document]
macobo: Clarified. I can see terminology causing issues, suggestions on improving this doc?

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,214 @@
...
+1. Contractor type work (e.g. for farm management agronomists work for multiple farms at once)
+2. Billing happens on a different level (e.g. our projects vs organization)
+3. Users clients are global companies with a lot of local “branches” where execs need access to everything
+4. Internal users have access to debug issues on multiple accounts
+
+### 2. User is active in only “group” at any point in time
+
+Whenever an event is made, we can (and should) make the assumption that it belongs to a single group (e.g. organization). This is needed to avoid “double-counting” in stats.
+
+Example: If Yakko has been added to 20 different organizations and is active, we don’t want to count all of them as active.
```
yakkomajuri, line 48: :D

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,214 @@
...
+### 1. Users can belong to multiple groups
+
+Some examples:
+
+1. Contractor type work (e.g. for farm management agronomists work for multiple farms at once)
+2. Billing happens on a different level (e.g. our projects vs organization)
+3. Users clients are global companies with a lot of local “branches” where execs need access to everything
+4. Internal users have access to debug issues on multiple accounts
+
+### 2. User is active in only “group” at any point in time
```
yakkomajuri, line 44: worth specifying this is one group per "group type"?

e.g. I _can_ be in `org: PostHog` and `project: HogFlix` at the same time

yakkomajuri (commented):
❤️  

do let me know if you end up needing some help with plugin server/libs

excited for this feature!

tiina303 (commented):
This is exciting, great write-up.

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,214 @@
...
+
+```python
+    import posthog
+
+    posthog.capture(
+      user_id,
+      'purchase'
+      { plan: 'silver' },
+      groups={
+        'organization': 'org:6'
```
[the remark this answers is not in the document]
EDsCODE: +1 to above. Also, in general unclear on how the group information is stored for persons

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,214 @@
...
+
+We will expose a new API for creating and adding properties groups. This API can be exposed on the client level, e.g.
+
+```python
+    import posthog
+
+    posthog.groups.add_properties('organization', 'org:5', { name: 'Swedish Bakery', subscription_level: 'gold' })
+```
+
+There will not be any functionality for merging two groups, as that behavior is user-specific.
```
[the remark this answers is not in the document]
macobo: To clarify: Merging users is a thing we only do for users due to needing to support cases where user uses a different browser. This usecase does not exist for groups.

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,214 @@
...
+
+
+### In-App: filtering by group properties in insights
+
+In-app, users will be able to do analytics by groups and filter by them.
+
+To give an example:
+
+- Throughout insights, we’ll allow filtering insights with group properties. This will be implemented as a new tab per group type in the filters modal.
+- In trends, users can also select “unique organizations” after first capturing events for an “organization”
```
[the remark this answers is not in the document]
macobo: Yes, these are all examples. Each group type will show up as an option.

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Autocapture & posthog-js
+
+To make autocapture work with groups and to make posthog-js usage a bit simpler, we can provide the following API:
+
+```js
+    // Normal posthog usage
+    posthog.identify(userId)
+    // New proposed method, to be called right after .identify
+    posthog.group('instance', 'posthog.yourapp.com', { instance_prop: 5})
```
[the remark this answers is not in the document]
macobo: Since group types are event-level not user level this scenario doesn't affect things (unless you're using $identify event in your analytics). 

It shouldn't be part of identify - groups affect _events_ not users. See longer section down below on why.

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,214 @@
...
+
+This is to avoid situations where some users see a new feature and some do not.
+
+This also factors into their later analysis - for example, they’d like to treat two different “organizations” both as a single datapoint even if one has 1500 users and the other has 1.
+
+## Proposal
+
+### Groups are event-level
+
+Users can define up to N (5?) group types. For us these would be `instance` , `project` and `organization`.
```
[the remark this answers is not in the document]
macobo: Group types are created when events for that group type are captured.

requests-for-comments/2021-10-05-group-analytics.md
macobo: Added a clarifying note.

macobo: Added a clarificaiton

requests-for-comments/2021-10-05-group-analytics.md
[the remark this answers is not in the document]
marcushyett-ph: This topic is always going to be tough with respect to terminology. Groups & Group Types should be sufficient.

requests-for-comments/2021-10-05-group-analytics.md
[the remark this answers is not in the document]
tiina303: Hmm... not sure about that ... What about teams being merged in companies, something that can happen in large companies

requests-for-comments/2021-10-05-group-analytics.md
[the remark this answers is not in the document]
tiina303: > What do we do when a new group type appears? 

Should we put this in the RFC?

> If no such column found but there are free columns for that project, we put it in the next one

Typos in a code can create columns, which means the user can be left with odd testing names for the columns. So we should add a way to be able to rename the columns. Doesn't need to be in the MVP. 

> If out of columns, we ignore

This will likely come to as a surprise to users at one point & cause confusion. Is there a way to assign a group later? If we had `group_assignment_error` or something property added to the event, we could at least discover the problem. Doesn't need to be in the MVP. But maybe add "TBD reporting the failure"

requests-for-comments/2021-10-05-group-analytics.md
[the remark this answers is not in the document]
macobo: Then what about teams being split in companies? ;)

The ramification of supporting "merging" is messing with historical data. This:
1. Messes with historical data - e.g. when looking at history of reports the results will _change_ based on the merge
2. It creates a lot of extra complexity on our end
3. We didn't hear about this usecase from any users either
4. Can be met by new events being set to a new group.

We can always add extra layers etc later - but this isn't relevant right now.

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,218 @@
...
+    posthog.group('instance', 'posthog.yourapp.com', { instance_prop: 5})
+```
+
+After calling this method, we will automatically associate events from that session with that group until either:
+
+- `posthog.group` is called with the same group type
+- `posthog.reset()` is called.
+### Feature flags
+
+Users should be able to add group-level property filters to feature flags and be able to ensure rollouts at a group level (?).
```
kpthatsme, line 159: Another (separate or complimentary) approach here may be support for feature flags created _against_ group types instead of users. I.e. you specify @ flag creation what kinds of groups this flag is being rolled to.

requests-for-comments/2021-10-05-group-analytics.md
kpthatsme: Slightly different suggestion on prioritization-

I feel like feature flag support should be moved up – it is core functionality and our solution today is a non starter for orgs that take testing seriously (since they can't guarantee the same experience for teams).

I'd suggest prioritizing feature flags after groups pages support. 

Re: other client libraries – I think the most important thing here is for us to create the HTTP api so folks can still roll their own solution if needed. There's a lot of work to do for client libraries in general.

samwinslow: I'm excited about group analytics in general!

Eventually, I'd like to see what kinds of queries we'll be able to build — a powerful way to look at data would be by "unique groups", such as we can already do with unique users. I could see us using that a lot.

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,218 @@
...
+
+For the users who are using feature flags, they want to ensure a consistent experience within a group during rollout.
+
+This is to avoid situations where some users see a new feature and some do not.
+
+This also factors into their later analysis - for example, they’d like to treat two different “organizations” both as a single datapoint even if one has 1500 users and the other has 1.
+
+## Proposal
+
+### Groups are event-level
```
paolodamico, line 70: One thing is not clear to me from a UX perspective is how the user will understand the group hierarchy conceptually. For us an instance can have multiple orgs, an org multiple teams, ... but our users may have a different setup with a different hierarchy. I want to make sure we have a lot of clarity around the mental model of this from a user's perspective.

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,218 @@
...
+
+```python
+    import posthog
+
+    posthog.groups.add_properties('organization', 'org:5', { name: 'Swedish Bakery', subscription_level: 'gold' })
+```
+
+There will not be any functionality for merging two groups, as that behavior is user-specific.
+
+### Capture API
```
paolodamico, line 90: I'm guessing a good chunk of our users will have only a 1:1 matching to their groups, the most basic case is just users belonging to an org. For these users, could we make some special handling so they don't have to add the extra properties that don't change on every request/page load?

requests-for-comments/2021-10-05-group-analytics.md
```diff
@@ -0,0 +1,218 @@
...
+To give an example:
+
+- Throughout insights, we’ll allow filtering insights with group properties. This will be implemented as a new tab per group type in the filters modal.
+- In trends, users can also select “unique organizations” after first capturing events for an “organization”
+- In funnels, users can select to analyze by a group type (e.g. “instance”). Different users within the group can have completed different steps of the funnel.
+- In retention, users can select “organization” over “user” for the unit to analyze retention by
+- And so on…
+
+
+### In-App: profile pages for groups
```
paolodamico, line 127: I suggest involving @clarkus very early on. Probably worth giving this more thought into what we want to expose to users, perhaps there's a simpler way to display this all together. We would also probably want to add some groups information into the Person page.

requests-for-comments/2021-10-05-group-analytics.md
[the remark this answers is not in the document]
paolodamico: Worth clarifying if this is a Scale or Enteprise feature as we now have both.

macobo:
Merging this in as work has started. Clarifying feedback points as well!

requests-for-comments/2021-10-05-group-analytics.md
macobo: Will do once there's something to iterate! :)

requests-for-comments/2021-10-05-group-analytics.md
macobo: We asked about this in the interviews: the TL;DR is that it didn't seem to be an urgent issue. Users often did have a hierarchical structure to their groups which they knew from heart and didn't need help on. 

But as with everything, we might need to iterate this after shipping.

Issue #4, opened 2021-07-13
Community Conversations: Questions for Leo


shehuphd:
foo.bar

PR #39, opened 2022-01-27
Future of our person model

macobo:
For hoglets: This PR is a follow-up to https://github.com/PostHog/product-internal/pull/240, outlining one area of improvement.

I suggest reading the proposal via rich diff, hit the following button in Files tab: 
image

### DIFF
--- a/requests-for-comments/2022-01-27-future-of-person-model.md
+++ b/requests-for-comments/2022-01-27-future-of-person-model.md
@@ -1,4 +1,4 @@
-# Request for Comments: *topic of the proposal*
+# Request for Comments: Future of person model
 
 - **Created:** 2022-01-27
 - **Owner:** @macobo, @marcushyett-ph 
--- a/requests-for-comments/2022-01-27-future-of-person-model.md
+++ b/requests-for-comments/2022-01-27-future-of-person-model.md
@@ -1,0 +1,105 @@
+# Request for Comments: *topic of the proposal*
+
+- **Created:** 2022-01-27
+- **Owner:** @macobo, @marcushyett-ph 
+
+## Problem: Scaling queries horizontally
+
+In posthog, having insights load fast is key for a great experience. However for large projects tracking >10M active persons per month query performance degrades fast.
+
+This can't be solved by scaling horizontally, because a lot of our queries do aggregation on persons - e.g. filtering by user properties, counting daily active persons and calculating funnels. 
+
+To do these queries we need to JOIN `person` and `person_distinct_id` tables. This in clickhouse is really slow (can cause >30x slower queries) and requires these tables to be replicated on every node, setting an upper bound to scalability. Teams with >10M users often run into memory usage problems as well.
+
+This RFC proposes both a technical and product solution to this problem: Store `person_id` and `person_properties` on the main events table.
+
+However, doing so will require breaking changes to the product. These changes are likely to cause some issues for customers, but these have either a negligible impact (affecting around 0.03% of events) or can be mitigated through documentation and education.
+
+## Proposed solution
+
+From a technical standpoint: Store `person_id` and `person_properties` on the main events table.
+
+From a product standpoint storing person_id and person_properties on events requires making the following changes:
+1. After the first session we will no longer merge events prior to login with the main person’s events.
+
+2. Person properties become immutable: Changing properties applies only on new events and past data is not updated.
+
+### Customer Impact
+
+#### Benefits
+For both changes the benefits are as follows:
+Significant short term performance gains
+Ability to support the largest tiers of customer now and in the future
+Standard analysis such as events from paying users over time will stop yielding misleading results when people start / stop paying (as they do today)
+
+
+#### 1. After the first session we will no longer merge events prior to login with the main person’s events.
+How we'll map persons to events going forward:
+1. On initial visit, users device gets assigned an an (anonymous) distinct id `anonid1`. Posthog internally associates person_id `p1` to `anonid1`, which is saved together with their events
+2. When after signing up `posthog.identify('userid')` is called, we associate the same person_id `p1` with the new distinct id `userid`
+3. If the user later visits the site with a new device, a new anonymous distinct id `anonid2` is created. Posthog internally associates these events with a new person id `p2` as no mapping for that id exists.
+4. If after signing in `posthos.identify('userid')` is called, we associate events going forward with person_id `p1`
+
+As a result, events done between steps 1, 2 and 4 will be tracked as from person `p1` but events between steps 3 and 4 will be under person`p2`. As a result no events are lost but some will be considered as coming from a different (anonymous) person.
+
+This affects use cases where users access their product (Logged out) on multiple devices regularly. However, based on our analysis the impact is only on ~0.03% events, some marketing use-cases are more significantly affected.
+
+As a mitigation, if necessary, we would consider a buffer mechanism to reduce affected events to ~0.003%, we can provide further details on this solution, however we would ideally avoid implementing this as it increases the complexity of the system quite significantly.
+
+#### 2. Person properties become immutable: Changing properties applies only on new events and past data is not updated.
+
+Example: Count of active users broken down by `paying` user property each month in 2021.
+
+After this change, if `paying` property of a user in Dec 2021 is changed to `false`, count of `paying` users does not change for e.g. March 2021. The  user will be considered non-paying until their `paying` property changes again2.
+
+It also means users could be counted as both `paying=true` and `paying=false` during a given month when they change state.
+
+User properties will change over time on events, some customers may rely on the existing behavior (where the latest property is assumed for all prior events), however the new behavior is considered much more accurate by our customer feedback.
+
+Our proposed mitigation or most use-cases (e.g. events for users who have paid at some point) cohorts can be used to achieve the same results, we will document this for affected users
+
+
+### Technical implications
+
+### Benefits
+1. Faster queries: remove the need to do JOINs for any insight queries.]
+2. Unlock: Our query results no longer change unless new data has come in - we make it possible to later implement time-series caching to speed up repeated queries over long time frames (>10x improvement expected on repeated calculations).
+4. We can sample by person_id in queries, which will make it possible to build a quick way to speed up queries in huge (1Bn+ events) installations with no additional query time resource cost.
+5. Queries will require less memory and no longer OOM (making it cheaper and more reliable to operate posthog)
+3. We can shard by person_id, simplifying the sharding story a lot and removing the need for cross-shard communication when sharding.
+
+Note that some of these changes have been benchmarked [here](https://github.com/PostHog/posthog/issues/7962)
+
+
+#### Challenges
+
+6. We will need to set up some strong consistency for message ordering in plugin-server and kafka, to make sure messages are processed in the right order per user.
+7. We will store more data in the events table. See below for more details.
+8. Deleting user properties becomes more expensive due to being stored in a redundant way.
+
+<details><summary>Details on cost of storage challenges</summary>
+
+
+Storing person properties in events means we’ll be storing more data.
+
+The exact impact of this depends on the dataset. Some factors about our setup which might not make this into a huge deal:
+
+1. The data is low cardinality and clickhouse is great at compression => we should get really good compression ratios
+2. When [Clickhouse implements semistructured data support](https://github.com/ClickHouse/ClickHouse/pull/23932), compression ratios will get even better.
+
+I’ve done a measurement on a test benchmarking cluster:
+- Original table was 7.56 GiB on disk, 99.13 GiB uncompressed
+- New table (with person_properties/person_id joined in) was 9.16 GiB on disk, 167.88 GiB uncompressed
+- person_properties column was 1.13 GiB on disk, 51.21 GiB uncompressed with a compression ratio of 0.022
+- For reference, `properties` column was 4.49 GiB on disk, 76.28 GiB uncompressed with a compression ratio of 0.059
+
+This measurement was done on a dataset that would compress relatively well (few people), however looking at the uncompressed size, we’d expect at most a 2x disk size increase in most cases.
+
+</details>
+
+## Conclusion
+Scalability of Posthog will be critical for our success, this project lays the foundation for high scalability, removes a critical scaling bottleneck and unblocks future improvements for even larger deployments. There are both technical and product risks associated with this project but they are far outweighed by the long-term benefits.
+
+## Open questions
+How do we handle historical data imports?
+How do we implement the change on cloud in a cost-effective way?
--- a/requests-for-comments/2022-01-27-future-of-person-model.md
+++ b/requests-for-comments/2022-01-27-future-of-person-model.md
@@ -21,7 +21,6 @@ From a technical standpoint: Store `person_id` and `person_properties` on the ma
 
 From a product standpoint storing person_id and person_properties on events requires making the following changes:
 1. After the first session we will no longer merge events prior to login with the main person’s events.
-![image](https://user-images.githubusercontent.com/85295485/151368075-a7ab9dc7-ee8d-4cad-a454-2524161973f1.png)
 2. Person properties become immutable: Changing properties applies only on new events and past data is not updated.
 
 ### Customer Impact
@@ -35,7 +34,7 @@ Standard analysis such as events from paying users over time will stop yielding 
 
 #### 1. After the first session we will no longer merge events prior to login with the main person’s events.
 
-![image](https://user-images.githubusercontent.com/148820/151367498-39637fc6-27ef-4058-b479-1651a4121875.png)
+![image](https://user-images.githubusercontent.com/85295485/151368075-a7ab9dc7-ee8d-4cad-a454-2524161973f1.png)
 
 How we'll map persons to events going forward:
 1. On initial visit, users device gets assigned an an (anonymous) distinct id `anonid1`. Posthog internally associates person_id `p1` to `anonid1`, which is saved together with their events
@@ -43,7 +42,7 @@ How we'll map persons to events going forward:
 3. If the user later visits the site with a new device, a new anonymous distinct id `anonid2` is created. Posthog internally associates these events with a new person id `p2` as no mapping for that id exists.
 4. If after signing in `posthos.identify('userid')` is called, we associate events going forward with person_id `p1`
 
-As a result, events done between steps 1, 2 and 4 will be tracked as from person `p1` but events between steps 3 and 4 will be under person`p2`. As a result no events are lost but some will be considered as coming from a different (anonymous) person.
+As a result, events done between steps 1, 2 and 4 will be tracked as from person `p1` but events between steps 3 and 4 will be under person`p2`. As a result no events are lost but some will be considered as coming from a different (anonymous) person. In the above image this corresponds to the green period.
 
 This affects use cases where users access their product (Logged out) on multiple devices regularly. However, based on our analysis the impact is only on ~0.03% events, some marketing use-cases are more significantly affected.
 
--- a/requests-for-comments/2022-01-27-future-of-person-model.md
+++ b/requests-for-comments/2022-01-27-future-of-person-model.md
@@ -104,5 +104,6 @@ This measurement was done on a dataset that would compress relatively well (few 
 Scalability of Posthog will be critical for our success, this project lays the foundation for high scalability, removes a critical scaling bottleneck and unblocks future improvements for even larger deployments. There are both technical and product risks associated with this project but they are far outweighed by the long-term benefits.
 
 ## Open questions
-How do we handle historical data imports?
-How do we implement the change on cloud in a cost-effective way?
+
+- How do we handle historical data imports?
+- How do we implement the change on cloud in a cost-effective way?
--- a/requests-for-comments/2022-01-27-future-of-person-model.md
+++ b/requests-for-comments/2022-01-27-future-of-person-model.md
@@ -104,5 +104,5 @@ Scalability of Posthog will be critical for our success, this project lays the f
 
 ## Open questions
 
-- How do we handle historical data imports?
-- How do we implement the change on cloud in a cost-effective way?
+- How do we handle events coming in out of order (both short time differences from network etc and historical data imports)?
+- How do we implement the changeover on cloud and self-hosted instances?
--- a/requests-for-comments/2022-01-27-future-of-person-model.md
+++ b/requests-for-comments/2022-01-27-future-of-person-model.md
@@ -34,6 +34,9 @@ Standard analysis such as events from paying users over time will stop yielding 
 
 
 #### 1. After the first session we will no longer merge events prior to login with the main person’s events.
+
+![image](https://user-images.githubusercontent.com/148820/151367498-39637fc6-27ef-4058-b479-1651a4121875.png)
+
 How we'll map persons to events going forward:
 1. On initial visit, users device gets assigned an an (anonymous) distinct id `anonid1`. Posthog internally associates person_id `p1` to `anonid1`, which is saved together with their events
 2. When after signing up `posthog.identify('userid')` is called, we associate the same person_id `p1` with the new distinct id `userid`
--- a/requests-for-comments/2022-01-27-future-of-person-model.md
+++ b/requests-for-comments/2022-01-27-future-of-person-model.md
@@ -21,7 +21,7 @@ From a technical standpoint: Store `person_id` and `person_properties` on the ma
 
 From a product standpoint storing person_id and person_properties on events requires making the following changes:
 1. After the first session we will no longer merge events prior to login with the main person’s events.
-
+![image](https://user-images.githubusercontent.com/85295485/151368075-a7ab9dc7-ee8d-4cad-a454-2524161973f1.png)
 2. Person properties become immutable: Changing properties applies only on new events and past data is not updated.
 
 ### Customer Impact


requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+1. On initial visit, users device gets assigned an an (anonymous) distinct id `anonid1`. Posthog internally associates person_id `p1` to `anonid1`, which is saved together with their events
+2. When after signing up `posthog.identify('userid')` is called, we associate the same person_id `p1` with the new distinct id `userid`
+3. If the user later visits the site with a new device, a new anonymous distinct id `anonid2` is created. Posthog internally associates these events with a new person id `p2` as no mapping for that id exists.
+4. If after signing in `posthos.identify('userid')` is called, we associate events going forward with person_id `p1`
+
+As a result, events done between steps 1, 2 and 4 will be tracked as from person `p1` but events between steps 3 and 4 will be under person`p2`. As a result no events are lost but some will be considered as coming from a different (anonymous) person. In the above image this corresponds to the green period.
+
+This affects use cases where users access their product (Logged out) on multiple devices regularly. However, based on our analysis the impact is only on ~0.03% events, some marketing use-cases are more significantly affected.
```
timgl, line 47: Could you elaborate on the marketing use-cases that would be affected? Most marketing websites don't `posthog.identify` users at all and primarily rely on cookies. I guess you're maybe thinking about tracking re-activation of users?

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+3. We can shard by person_id, simplifying the sharding story a lot and removing the need for cross-shard communication when sharding.
+
+Note that some of these changes have been benchmarked [here](https://github.com/PostHog/posthog/issues/7962)
+
+
+#### Challenges
+
+6. We will need to set up some strong consistency for message ordering in plugin-server and kafka, to make sure messages are processed in the right order per user.
+7. We will store more data in the events table. See below for more details.
+8. Deleting user properties becomes more expensive due to being stored in a redundant way.
```
timgl, line 80: Do you see the migration as a potential challenge? How long did it take to set up the test table for us for example?

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+
+</details>
+
+## Conclusion
+Scalability of Posthog will be critical for our success, this project lays the foundation for high scalability, removes a critical scaling bottleneck and unblocks future improvements for even larger deployments. There are both technical and product risks associated with this project but they are far outweighed by the long-term benefits.
+
+## Open questions
+
+- How do we handle historical data imports?
+- How do we implement the change on cloud in a cost-effective way?
```
timgl, line 108: Why cost effective? I thought we had quite a lot of spare storage on our clusters and the increase doesn't sound too crazy. Is it CPU intensive?

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+This measurement was done on a dataset that would compress relatively well (few people), however looking at the uncompressed size, we’d expect at most a 2x disk size increase in most cases.
+
+</details>
+
+## Conclusion
+Scalability of Posthog will be critical for our success, this project lays the foundation for high scalability, removes a critical scaling bottleneck and unblocks future improvements for even larger deployments. There are both technical and product risks associated with this project but they are far outweighed by the long-term benefits.
+
+## Open questions
+
+- How do we handle historical data imports?
```
timgl, line 107: I'm not sure how big of an issue this is. My hunch is that most people who do historical exports don't import person properties or `$identify` events correctly anyway. Also, in the current implementation those historical events would have the current person properties attached to them, so there wouldn't be a regression.

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+
+In posthog, having insights load fast is key for a great experience. However for large projects tracking >10M active persons per month query performance degrades fast.
+
+This can't be solved by scaling horizontally, because a lot of our queries do aggregation on persons - e.g. filtering by user properties, counting daily active persons and calculating funnels. 
+
+To do these queries we need to JOIN `person` and `person_distinct_id` tables. This in clickhouse is really slow (can cause >30x slower queries) and requires these tables to be replicated on every node, setting an upper bound to scalability. Teams with >10M users often run into memory usage problems as well.
+
+This RFC proposes both a technical and product solution to this problem: Store `person_id` and `person_properties` on the main events table.
+
+However, doing so will require breaking changes to the product. These changes are likely to cause some issues for customers, but these have either a negligible impact (affecting around 0.03% of events) or can be mitigated through documentation and education.
```
timgl, line 16: Do you have links to the analysis done here? Curious to see which users and what kind of events would be affected.

requests-for-comments/2022-01-27-future-of-person-model.md
macobo: Good question. This was also mentioned under open questions.

Data copying wise - no. As part of https://github.com/PostHog/posthog/issues/5684 I figured out a way to do the copying 20x faster than what we've done so far, which is good enough.

Infrastructure-wise - maybe. Due to the extra storage we might need to add nodes to our cluster, which has budget ramifications etc we'd need to figure out.

Implementation-wise - definetly! Actually ingesting data in a consistent way so events get up-to-date person properties is going to be a challenge due to similar reasons it's a challenge right now. :)

timgl (approved):
Strongly in favour of doing this, should be a better and more consistent experience for customers + unlock way higher scale.

requests-for-comments/2022-01-27-future-of-person-model.md
macobo: A sample one is this story:

1. User signs up
2. Some time later user becomes inactive
3. App sends user an email
4. User opens link in a re-activation/marketing email on a new device, signed out
5. User signs in and becomes active

If you wanted to track users who opened the link due to this change you will need some custom code on your end.

cc @marcushyett-ph you might have a few others.

requests-for-comments/2022-01-27-future-of-person-model.md
marcushyett-ph: Yep - [doc is a little messy](https://docs.google.com/document/d/1RMOoEH-eq8vp2fHqta_JEqgF6Tbd6s7ARhacDpQ5DIo/edit#heading=h.r6voxlxduftz) but theres a data report section on the impact analysis and a section including customer interviews.

requests-for-comments/2022-01-27-future-of-person-model.md
macobo: Partially answered in https://github.com/PostHog/meta/pull/39#discussion_r793623444

The sticking point _might_ be disk. We have some leeway right now (In practice disk is ~50% full which includes session recordings) but the copying process + extra data requirements of the new schema might push us to the point where we want more nodes or rethink storage.

requests-for-comments/2022-01-27-future-of-person-model.md
timgl: Makes sense! I guess I was specifically worried about self hosting. I think there's a reasonable way to implement this whilst running both tables at the same time?

macobo: Yeah, I think so:
1. We can leverage async migrations to do it in a non-intrusive way
2. We could add the person_id and person_properties columns to the existing table and backfill, but would need logic in the app to know when a query needs to join vs when it needs to read from the table. Some  questions around how to handle the difference in person semantics in the product will arise as a result but we can figure them out if this project kicks off :)

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+4. We can sample by person_id in queries, which will make it possible to build a quick way to speed up queries in huge (1Bn+ events) installations with no additional query time resource cost.
+5. Queries will require less memory and no longer OOM (making it cheaper and more reliable to operate posthog)
+3. We can shard by person_id, simplifying the sharding story a lot and removing the need for cross-shard communication when sharding.
+
+Note that some of these changes have been benchmarked [here](https://github.com/PostHog/posthog/issues/7962)
+
+
+#### Challenges
+
+6. We will need to set up some strong consistency for message ordering in plugin-server and kafka, to make sure messages are processed in the right order per user.
```
yakkomajuri, line 78: Is this a definite necessity? This is a tough one, even if we put implementation aside.

There are many things at play here:

1. Users might want to send us old events 
2. Users may accidentally send us old events (e.g. through retries, network issues, etc)
3. Data imports/exports
4. Plugin server processing events out of order makes things easier performance-wise

#4 is mostly implementation, but the other 3 are inherent concepts that clash with this

tiina303 (commented):
🙌  Awesome proposal. Just some questions about how we're thinking of handling certain things/situations.

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+Our proposed mitigation or most use-cases (e.g. events for users who have paid at some point) cohorts can be used to achieve the same results, we will document this for affected users
+
+
+### Technical implications
+
+### Benefits
+1. Faster queries: remove the need to do JOINs for any insight queries.]
+2. Unlock: Our query results no longer change unless new data has come in - we make it possible to later implement time-series caching to speed up repeated queries over long time frames (>10x improvement expected on repeated calculations).
+4. We can sample by person_id in queries, which will make it possible to build a quick way to speed up queries in huge (1Bn+ events) installations with no additional query time resource cost.
+5. Queries will require less memory and no longer OOM (making it cheaper and more reliable to operate posthog)
```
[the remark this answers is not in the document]
macobo: This benefit refers back to the opening section - currently queries from clients with 10M+ persons will OOM due since the whole persons dataset needs to be loaded into memory for joins.

By removing the JOIN we also reduce memory usage significantly and there's no more conceptual roadblock. Yes, badly implemented queries _might_ still OOM but that's out of scope for this proposal.

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+
+Example: Count of active users broken down by `paying` user property each month in 2021.
+
+After this change, if `paying` property of a user in Dec 2021 is changed to `false`, count of `paying` users does not change for e.g. March 2021. The  user will be considered non-paying until their `paying` property changes again2.
+
+It also means users could be counted as both `paying=true` and `paying=false` during a given month when they change state.
+
+User properties will change over time on events, some customers may rely on the existing behavior (where the latest property is assumed for all prior events), however the new behavior is considered much more accurate by our customer feedback.
+
+Our proposed mitigation or most use-cases (e.g. events for users who have paid at some point) cohorts can be used to achieve the same results, we will document this for affected users
```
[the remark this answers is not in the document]
macobo: > Were there any specific properties/examples for when latest property was desired?

From speaking to customers: not really. Changing properties like "email" is not a real usecase that affects day-to-day usage of the product and in practice point-in-time properties hits real usecases (like e.g. payment breakdown) much better.

That said if you have counter-examples from your usage of posthog 100% worth posting them here.

requests-for-comments/2022-01-27-future-of-person-model.md
macobo: It's an operation we currently provide via the UI and comes up relatively often around compliance requirements, hence brought up here. 

That said it's not in the critical path of the app and not a huge blocker when implementing it.

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+
+However, doing so will require breaking changes to the product. These changes are likely to cause some issues for customers, but these have either a negligible impact (affecting around 0.03% of events) or can be mitigated through documentation and education.
+
+## Proposed solution
+
+From a technical standpoint: Store `person_id` and `person_properties` on the main events table.
+
+From a product standpoint storing person_id and person_properties on events requires making the following changes:
+1. After the first session we will no longer merge events prior to login with the main person’s events.
+2. Person properties become immutable: Changing properties applies only on new events and past data is not updated.
```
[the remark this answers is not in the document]
macobo: Out of scope for this proposal - goal is to align us (& the community, this being on the public issue tracker) with the breaking product changes, once agreed we can start speccing out the implementation with a much smaller circle.

requests-for-comments/2022-01-27-future-of-person-model.md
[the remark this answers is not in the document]
tiina303: My specific examples for email:
* needing emails to know who's still self-hosting posthog postgres for the marketing team to reach out (definitely wanted the latest emails only). Though I think here I was using some event property, but I can see others having this use case.
* there was a problem with in the app, that caused something weird to happen for certain users - we identify an event how to find those users (e.g. users running a broken version of posthog in self hosted that can end up in a crash loop or potentially set some things in a bad way and requires manual action from the instance operator) - same here I'd want latest emails.

Agree that they are likely rare and that point-in-time properties are usually what users want. Just thought it might be helpful for us to have a short snippet in the docs for everyone for how to get latest property values for anything (i.e. cohort & then potentially random insight on that cohort)

requests-for-comments/2022-01-27-future-of-person-model.md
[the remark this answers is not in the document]
tiina303: > Yes, badly implemented queries might still OOM

badly set resource requirements too on self hosted

> but that's out of scope for this proposal.

Agreed. 

I seem to have worded my point badly. I was wondering if there's a bigger benefit here of "memory usage more predictable/less spiky" which to me is a bigger selling point as that would make self hosted users and our life easier for managing their instance compared to "no longer OOM for specific customers on specific instances"

requests-for-comments/2022-01-27-future-of-person-model.md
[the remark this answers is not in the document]
tiina303: It changes things for the instance operator though quite significantly - for the past events does their storage cost increase (if we add user props to all past events) or not (if we do some query time magic with user props table for older events), so I think it's relevant. Specifically you said [here](https://github.com/PostHog/meta/pull/39/files#r793690132)

> would need logic in the app to know when a query needs to join vs when it needs to read from the table

which makes me assume the plan was not to backfill, but I think we should as that would make everything easier. So we don't need the migration plan, but we need to know the impact of the migration especially for self-hosted users but also for cloud for our cost management.

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Customer Impact
+
+#### Benefits
```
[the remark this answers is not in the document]
EDsCODE: Extending on 5, do we have analysis on how often users are trying to update person properties atm? I agree with the approach but this could be a crippling experience if we disrupt X% of user experiences with updating person properties

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+This RFC proposes both a technical and product solution to this problem: Store `person_id` and `person_properties` on the main events table.
+
+However, doing so will require breaking changes to the product. These changes are likely to cause some issues for customers, but these have either a negligible impact (affecting around 0.03% of events) or can be mitigated through documentation and education.
+
+## Proposed solution
+
+From a technical standpoint: Store `person_id` and `person_properties` on the main events table.
+
+From a product standpoint storing person_id and person_properties on events requires making the following changes:
+1. After the first session we will no longer merge events prior to login with the main person’s events.
```
kpthatsme, line 23: Am I interpreting this right? Is this saying that after making this change we'll no longer be able to merge an anoynmous user's timeline into the identified users?

So for example if I come to posthog.com as an anonymous user, distinct ID A and signup as distinct ID kunal@posthog.com – distinct ID A's will always look like a separate person from kunal@posthog.com?

macobo: It depends.

The section below under customer impact goes into more detail.

To elaborate in different words though: during initial session (when kunal@ hasn't ever been identified) it behaves as you expect. The difference comes in on subsequent sessions e.g. on a different device where events pre login would get associated to a different person.

Let me know if this or the other explanation below make sense!

kpthatsme: Thanks - let's go a little deeper:

In today's world...

I come on incognito window from my phone:
- I have distinct ID A assigned. 
- I do a bunch of things before logging in
- I login, events performed under Distinct ID A can now be found looking at the timeline for kunal@posthog.com


I then come on from an incognito window from my laptop:
- I have distinct ID B assigned
- I do a bunch of things prior to login
- I login, events performed under Distinct ID B can now be found at the timeline for kunal@

So now looking at the event timeline for kunal@ - I see all the events that were performed from both distinct IDs A and B. 

Which pieces of this would break (if any) in the proposed change?

requests-for-comments/2022-01-27-future-of-person-model.md
macobo: Agreed 100% - ingestion server changes are the meat of implementing this.

I intentionally left the details up in the air in this proposal implementation-wise and was planning to follow up with a proper plan if we decide to accept these breaking product changes. Please also take a look at these and give your thoughts on whether they're acceptable trade-offs for you.

However to answer the question:

Complexity needs to live somewhere. We're constrained by being unable to "update" data in clickhouse (even via replacing/collapsing) so this complexity naturally falls on the ingestion side of things.

I intentionally haven't defined what it means to do $set here - e.g. if you $set in an event from March 2021 what effects this should have in future events. I believe the approach we should go for is that there'll still be a set of "current" user properties (as we have currently) that will apply for _new_ events that get ingested.

Regarding ordering messages, the high-level thinking it's doable by:
- Partitioning our kafka topic by distinct_id
- Grouping messages each batch by distinct_id, enforcing ordering per distinct_id in the plugin server.

That said, I haven't put too much thought into this and out-of-order delivery. This is an area I was hoping to recruit your + @tiina303 help on if this moves on as data consistency is a key worry.

Do the higher level principles make sense? Everything said here we can iterate on should we accept the breaking changes and would like to avoid this turning into a full technical spec. :)

requests-for-comments/2022-01-27-future-of-person-model.md
marcushyett-ph: @kpthatsme it's also worth reading [this doc from our friends at another YC analytics company](https://help.amplitude.com/hc/en-us/articles/115003135607-Track-unique-users-in-Amplitude) - as far as we can tell their merging system has a similar behavior - but they might have done a better job at explaining it

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### 1. After the first session we will no longer merge events prior to login with the main person’s events.
+
+![image](https://user-images.githubusercontent.com/85295485/151368075-a7ab9dc7-ee8d-4cad-a454-2524161973f1.png)
+
+How we'll map persons to events going forward:
+1. On initial visit, users device gets assigned an an (anonymous) distinct id `anonid1`. Posthog internally associates person_id `p1` to `anonid1`, which is saved together with their events
+2. When after signing up `posthog.identify('userid')` is called, we associate the same person_id `p1` with the new distinct id `userid`
+3. If the user later visits the site with a new device, a new anonymous distinct id `anonid2` is created. Posthog internally associates these events with a new person id `p2` as no mapping for that id exists.
+4. If after signing in `posthos.identify('userid')` is called, we associate events going forward with person_id `p1`
```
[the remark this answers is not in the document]
macobo: > For merging properties I assume we'll want to do that based on timestamps

Historical backfills are unscoped as part of this proposal - I was planning on doing a follow-up "implementation" spec should we accept this proposal focusing on the product changes. As mentioned in https://github.com/PostHog/meta/pull/39#discussion_r794407227 I would love your and Yakkos help with that.

> What about merging people (either from UI or via the alias command)? Is that planned to be treated the same way as logged-out user?

Not sure I understood the question here, but yes, the key concept is that we wouldn't update historical data on merges/identifies/aliases and these changes only apply going forward.

requests-for-comments/2022-01-27-future-of-person-model.md
marcushyett-ph: Yep its primarily around re-activation of existing users on different devices (so the size of the use-case is generally very small). By marketing I was referring primarily to scenarios where the logged out experience (usually marketing site) is part of the end-to-end funnel.

requests-for-comments/2022-01-27-future-of-person-model.md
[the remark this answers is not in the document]
macobo: These are all good points however I'd argue that for the sake of this proposal, they're mostly related to technical trade-offs. Customer impact here meaning end-users who are using the posthog to do the analysis, not the operators.

> the initial events before logging in aren't tied to the same person

This interpretation isn't quite correct and understanding this is key for this proposal. To avoid writing the same topic twice, see the customer impact solution for graph + explanation + this table in my response to a similar question to kunal: https://github.com/PostHog/meta/pull/39#discussion_r794400029

Read through the proposal again (from a product, not technical standpoint) - would this be an acceptable loss for you as a posthog user?

> 2. Increased disk usage - due to redundancy we now need more storage
> 3. More data to copy during events table future migrations due to redundancy in storing everything on events making migrations slower, requiring more disk overhead
> 4. Some use cases now require 2 steps 1) cohort 2) look up most recent value (e.g. most up to date emails for folks who did X in some timeframe in the past)
> 5. If a critical user property (e.g. is_paying_customer) event went missing (network error whatever) user can currently just set it on the person and all queries will have it (there is the drawback that it's now set for all time, but say we don't care), however in the new system when that property is on events we can't add it later (well we could edit all events, but I assume that's not planned - or well could be done via async migrations if the need arises commonly). Not sure if we care...
> Extending on 5, do we have analysis on how often users are trying to update person properties atm? I agree with the approach but this could be a crippling experience if we disrupt X% of user experiences with updating person properties

I'd argue these are implementation concerns - let's figure them them out should we accept the 2 breaking changes in this proposal.

Resolving this thread for the moment since it's discussing out-of-scope issues and to make the discussion easier to follow. Unresolve if you agree!

requests-for-comments/2022-01-27-future-of-person-model.md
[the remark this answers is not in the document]
macobo: I think we're largely on the same page here - potato potata.

Reason I prefer current wording is that the problem users are facing right now is that some of their queries will fail due to running out of memory on a single individual query due to needing to JOIN `persons` table. Clickhouse memory usage for queries is inherently spiky due to people running queries very rarely from a systems standpoint.

requests-for-comments/2022-01-27-future-of-person-model.md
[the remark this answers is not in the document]
macobo: Will scope out a proper technical plan for this (hopefully in cooperation with you) if we accept the two breaking changes to the product!

requests-for-comments/2022-01-27-future-of-person-model.md
macobo: Tiina brings out the reasons I brought this up. But figuring out how to solve these concerns is unaffected by this proposal - we can figure it out during technical planning! :)

requests-for-comments/2022-01-27-future-of-person-model.md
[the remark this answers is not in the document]
macobo: No plan yet (as part of this proposal) - let's figure this out as part of technical planning!

requests-for-comments/2022-01-27-future-of-person-model.md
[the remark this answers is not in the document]
macobo: Getting latest values definitely makes sense and is needed to even power the Persons page! 

Let's explore this more during technical planning - as you mentioned elsewhere, this need will come up elsewhere as well (e.g. calculating cohorts).

requests-for-comments/2022-01-27-future-of-person-model.md
yakkomajuri: I'm not as worried about a) being able to enforce order on the events we do have and b) historical exports as much as I'm worried about e.g. two people in a group are using the app at the same time, one has internet woes. Or, even with distinct IDs - events from the web and server side might arrive in a different order as they're subject to different networking conditions.

So yeah, things like that. Either way, I'm not holding back the proposal on this point, was just wondering how much this whole system relies on fully ordered ingestion. If the answer is "this system relies extremely heavily on completely in-order ingestion at all costs" then I think we need a deeper discussion, else happy to move.

requests-for-comments/2022-01-27-future-of-person-model.md
mariusandra: Could we have some way to "backtrack" to the anonymous events/session post login?

For example, by adding the property `anon_distinct_id: "anonid2"` to events `11` and `12` in the table above. 

This would allow us to link session recordings with the pre-login part, and potentially allow some flexibility in the future when someone asks for a funnel one part of it is logged out events, and the other part consists of logged in events.

requests-for-comments/2022-01-27-future-of-person-model.md
tiina303, line 108: Follow-up to our discussion move "We will need to set up some strong consistency for message ordering in plugin-server and kafka, to make sure messages are processed in the right order per user." from challenges to open questions, specifically I would have it here be:
```suggestion
- How do we handle events coming in out of order (both short time differences from network etc and historical data imports)?
- How do we implement the changeover on cloud and self-hosted instances?
```
^ nitted the second one because we might ask feedback from users who are self hosting, this will make clear it's an open question

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+
+As a result, events done between steps 1, 2 and 4 will be tracked as from person `p1` but events between steps 3 and 4 will be under person`p2`. As a result no events are lost but some will be considered as coming from a different (anonymous) person. In the above image this corresponds to the green period.
+
+This affects use cases where users access their product (Logged out) on multiple devices regularly. However, based on our analysis the impact is only on ~0.03% events, some marketing use-cases are more significantly affected.
+
+As a mitigation, if necessary, we would consider a buffer mechanism to reduce affected events to ~0.003%, we can provide further details on this solution, however we would ideally avoid implementing this as it increases the complexity of the system quite significantly.
```
mariusandra, line 49: (reposting a new thread from an earlier comment)

Could we have some way to "backtrack" to the anonymous events/session post login?

For example, by adding the property `anon_distinct_id: "anonid2"` to events `11` and `12` in the table above. 

This would allow us to link session recordings with the pre-login part, and potentially allow some flexibility in the future when someone asks for a funnel one part of it is logged out events, and the other part consists of logged in events.

jamesefhawkins:
Have read through - this feels like it'll mainly cause issues around reactivation... which is a very late stage (ie Facebook size) company problem where you've already saturated the market and aren't hunting for new users. The area it'd harm doesn't reflect what our existing customers focus on - activation for example. That's demonstrated through the tiny percentage impact it'd have in the data.

We should document this behavior very clearly in the docs.

It's also a significant data parity win not changing historical graphs. Again, this needs documenting somewhere too very clearly so we don't have users thinking this is a bug when they expected the reverse behavior - ie graphs that are historic would change.

The top reason to do this from a customer view is performance issues are frequent and worth solving.

requests-for-comments/2022-01-27-future-of-person-model.md
tiina303: TLDR: what if we delayed ingestion storing the anonymous user events somewhere and ingesting after identify event with the logged_in person ID. Then we'd still get our perf and simplicity improvements for queries without losing pre-logged in events being tied to the same person.

Long version:
Assumptions: Pre-login we don't know the logged_in person_id, hence events will be sent with a new uuid

There are only 3 options (let me know if you can think of any more):
### 1. join persons 
This is what we do now & has all the perf problems for queries. If we are not querying over a single person_id for unique users queries results in us joining persons. Adding any properties etc results in the same need to join persons. I'd bucket any ideas with running queries across CH/postgres here too & I think they all have the same perf problem.

#### 2. update events in clickhouse 
Updating the person_id from the anonymous one to the logged_in person_id: We can't really mutate events in CH, what we could do is run an async migration on the shard (essentially copy all data to a new table modifying the distinct_id's accordingly & then rename tables like 0002 async migration). This is probably expensive in terms of resources even done for a single shard; that migration would put us temporarily in a state of delayed ingestion (while we copy the last part of the table and rename we'd be queueing up new events in kafka similar to 0002).
#### 3. delay ingestion until we know the logged_in person_id: 

There are two options here which might not be all that bad of an idea:
1. delay by using some intermediary storage for logged_out users: we could dump all anonymous events somewhere keyed by uuid and then consume them once we see the identify event (I was thinking kafka, but we can't get messages by key, maybe by some separate table in ClickHouse) and we we can write the events to ClickHouse with the logged_in person_id. I was thinking up to certain delay, e.g. 1 h or 1 day could be longer in theory too, but note that during that delay we wouldn't see those events in queries (which could be acceptable). I wonder if this would actually not be a terrible idea?
2. Re-ingest the events with the new person_id and mark the other uuid as a deleted person (do we filter out deleted people somehow in queries or in any case filtering out is potentially cheaper than joins). We can occasionally run an async migration that clears them out.

-----
 
Additionally there's a problem tied to person properties and how they would get updated, but we could just decide on a global semantics that person properties can only be used on logged_in users (in any case this is probably less important and the funnels working etc is more important).

requests-for-comments/2022-01-27-future-of-person-model.md
tiina303: TLDR: what if we delayed ingestion storing the anonymous user events somewhere and ingesting after identify event with the logged_in person ID. Then we'd still get our perf and simplicity improvements for queries without losing pre-logged in events being tied to the same person.

Long version:
Assumptions: Pre-login we don't know the logged_in person_id, hence events will be sent with a new uuid

There are only 3 options (let me know if you can think of any more):
### 1. join persons 
This is what we do now & has all the perf problems for queries. If we are not querying over a single person_id for unique users queries results in us joining persons. Adding any properties etc results in the same need to join persons. I'd bucket any ideas with running queries across CH/postgres here too & I think they all have the same perf problem.

#### 2. update events in clickhouse 
Updating the person_id from the anonymous one to the logged_in person_id: We can't really mutate events in CH, what we could do is run an async migration on the shard (essentially copy all data to a new table modifying the distinct_id's accordingly & then rename tables like 0002 async migration). This is probably expensive in terms of resources even done for a single shard; that migration would put us temporarily in a state of delayed ingestion (while we copy the last part of the table and rename we'd be queueing up new events in kafka similar to 0002).
#### 3. delay ingestion until we know the logged_in person_id: 

There are two options here which might not be all that bad of an idea:
1. delay by using some intermediary storage for logged_out users: we could dump all anonymous events somewhere keyed by uuid and then consume them once we see the identify event (I was thinking kafka, but we can't get messages by key, maybe by some separate table in ClickHouse) and we we can write the events to ClickHouse with the logged_in person_id. I was thinking up to certain delay, e.g. 1 h or 1 day could be longer in theory too, but note that during that delay we wouldn't see those events in queries (which could be acceptable). I wonder if this would actually not be a terrible idea?
2. Re-ingest the events with the new person_id and mark the other uuid as a deleted person (do we filter out deleted people somehow in queries or in any case filtering out is potentially cheaper than joins). We can occasionally run an async migration that clears them out.

-----
 
Additionally there's a problem tied to person properties and how they would get updated, but we could just decide on a global semantics that person properties can only be used on logged_in users (in any case this is probably less important and the funnels working etc is more important).

requests-for-comments/2022-01-27-future-of-person-model.md
macobo: Thank you!

requests-for-comments/2022-01-27-future-of-person-model.md
paolodamico: I would argue it's pretty critical that we're able to properly delete user properties everywhere. Not only is it a huge compliance red flag, I think brand-wise it could hurt that we're the privacy friendly analytics, and you can't delete potentially highly sensitive data.

There also seems to be a non-trivial amount of [current usage](https://app.posthog.com/insights/MNop0r80/edit#filters=%7B%22insight%22%3A%22TRENDS%22%2C%22interval%22%3A%22month%22%2C%22display%22%3A%22ActionsLineGraph%22%2C%22actions%22%3A%5B%7B%22id%22%3A%2214240%22%2C%22name%22%3A%22person%20deleted%20click%22%2C%22type%22%3A%22actions%22%2C%22order%22%3A0%7D%5D%2C%22events%22%3A%5B%7B%22id%22%3A%22person%20viewed%22%2C%22type%22%3A%22events%22%2C%22order%22%3A1%2C%22name%22%3A%22person%20viewed%22%7D%5D%2C%22properties%22%3A%5B%5D%2C%22filter_test_accounts%22%3Afalse%2C%22new_entity%22%3A%5B%5D%2C%22date_from%22%3A%22-90d%22%2C%22date_to%22%3Anull%7D) (~1k persons deleted per month), plus we've seen cases from Enterprise customers where this becomes a total blocker.

requests-for-comments/2022-01-27-future-of-person-model.md
paolodamico: Assuming at first we don't implement a solution to properly associate green events ("green" from diagram) to the proper user, is there a way we can mitigate impact to DAUs/WAUs/MAUs analytics? My biggest concern is that your active users metric becomes inaccurate because of this.

requests-for-comments/2022-01-27-future-of-person-model.md
```diff
@@ -0,0 +1,108 @@
...
+Note that some of these changes have been benchmarked [here](https://github.com/PostHog/posthog/issues/7962)
+
+
+#### Challenges
+
+6. We will need to set up some strong consistency for message ordering in plugin-server and kafka, to make sure messages are processed in the right order per user.
+7. We will store more data in the events table. See below for more details.
+8. Deleting user properties becomes more expensive due to being stored in a redundant way.
+
+<details><summary>Details on cost of storage challenges</summary>
```
paolodamico, line 82: Would it be feasible to do a detailed cost analysis that we can share with customers? My thinking is, we'll need more disk space (more $$) but less processing power (less $$), would this lead to a net increase/decrease in infrastructure cost?

paolodamico (approved):
Proposal makes sense to me. Particularly the trade-off of having reliable metrics over time (e.g. a user that was free and moved to paid at some point) vs. having some anonymized secondary usage unidentified, seems okay. Added comments in-line.

requests-for-comments/2022-01-27-future-of-person-model.md
macobo: I think so. Will figure this out as we get to implementation! :)

requests-for-comments/2022-01-27-future-of-person-model.md
macobo: To make it clear, I'm not proposing dropping it: rather my comment was to point out it's not as performance-critical as say ingestion is and we can make this happen even in the new system by accepting the extra load it generates.

requests-for-comments/2022-01-27-future-of-person-model.md
tiina303: > mitigate impact to DAUs/WAUs/MAUs analytics

perhaps filtering on logged in users would help?

requests-for-comments/2022-01-27-future-of-person-model.md
rcmarron: This is pretty interesting. In a way, this is what the `session_id` is today.

paolodamico:
Just cross-posting some user context/pain from [Slack](https://posthogusers.slack.com/archives/C01GLBKHKQT/p1646132983808019)

PR #40, opened 2022-03-08
RFC: Full SAML support

paolodamico:
See PR for details.

### DIFF
--- a/requests-for-comments/2022-03-08-full-saml.md
+++ b/requests-for-comments/2022-03-08-full-saml.md
@@ -1,29 +1,35 @@
 # RFC: Full SAML support
 
 - **Created:** *2022-03-08*
-- **Owner:** @paolodamico (and later @Twixes)
+- **Owner:** @mariusandra (previously @paolodamico)
 - **GitHub issues:** [#5647](https://github.com/PostHog/posthog/issues/5647)
 
 > While commenting on this, please comment on specific lines if possible, to thread the discussion.
 
 ## Overview
 
-Currently we support SAML only on self-hosted. More users have requested support for this feature on the Cloud version too. Further, the self-hosted version does not support multitenancy SAML. This RFC contains a technical proposal for how to implement full multitenant SAML support.
+Currently we support SAML only on self-hosted. More users have requested support for this feature on the Cloud version too. This RFC contains a technical proposal for how to implement full multitenant SAML support.
+
+On multitenant environments SAML can present a risk because these servers are asserting to the identity of a user, and we can't verify that a specific server has permission to assert the identity of a specific user (e.g. I could create a malicious SAML server that says I'm `admin@posthog.com` with no permissions over that login). To address this, we'll introduce domain verification through DNS so that users configuring SAML can also validate they have permissions over a specific domain.
 
 ## Background
 
 From the original SAML issue on GitHub,
 > This is a compliance feature that a lot of large-scale organizations require for full widespread rollout. When working at a large org, having a centralized place for provisioning and managing users can make the onboarding/offboarding process smoother and help reduce security risks. There's a great [Slack thread](https://posthogusers.slack.com/archives/CT7HXDEG3/p1628859965082200) by a community member about the importance of SAML and [this study](https://www.onelogin.com/press-center/press-releases/97-percent-of-saas-vendors-backing-saml-based-single-sign-on) by OneLogin in which they find that up to 97% of SaaS support have support (or plans) for SAML. 
-
-
 ## Proposal
 
-1. SAML configuration becomes organization-based instead of instance-based. We create a separate model for `OrganizationConfiguration` where we track all SAML parameters (ACS URL, Entity ID, X509 certificate and attribute names [4]), and of course drop all configuration from instance environment variables.
-2. For the next release, when updating, if SAML configuration is set and the instance has a single organization, we set the same SAML parameters for that organization.
-3. Drop `SAML_ENFORCED` in favor of a `SSO_ENFORCEMENT` attribute that supports `saml`, `google-oauth2`, `github` & `gitlab` to solve [#7535](https://github.com/PostHog/posthog/issues/7535) too (only one per organization).
-4. Drop `SAML_DISABLED` support to make maintenance easier and avoid having multiple build flows. Drawback is that the FOSS version which does not support SAML will still require the SAML dependencies.
-5. If SAML is enabled for at least one organization in the instance, the login page shows only the email input at first. Upon entering an email, we do an API request as shown below (where `enforced_sso` contains information on whether to enforce log in with a particular SSO provider [either SAML, Google, ...] or `null` if the user has no SSO enforced **or the user does not exist**).
-    ```
+### Self-hosted
+1. Self-hosted instances remain with single-tenant SAML support (i.e. only one SAML provider for each instance), but we drop `SAML_ENFORCED` in favor of a `SSO_ENFORCEMENT` instance-level attribute that supports `saml`, `google-oauth2`, `github` & `gitlab`. We redirect automatically to the provider's login page if SSO is enforced (solves [#7535](https://github.com/PostHog/posthog/issues/7535)).
+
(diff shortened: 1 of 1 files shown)


requests-for-comments/2022-03-08-full-saml.md
```diff
@@ -0,0 +1,48 @@
...
+        "email": "test@posthog.com"
+    }
+
+    Response:
+    {
+        "enforced_sso": "saml"
+    }
+    ```
+
+    > ⚠️ Security concern. If the user has any SSO enforced, it will leak information about a specific user being registered in the PostHog instance.
```
paolodamico, line 38: @Twixes maybe we should rate limit the API endpoint? Or is there another approach that doesn't leak info?

requests-for-comments/2022-03-08-full-saml.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+## Proposal
+
+1. SAML configuration becomes organization-based instead of instance-based. We create a separate model for `OrganizationConfiguration` where we track all SAML parameters (ACS URL, Entity ID, X509 certificate and attribute names [4]), and of course drop all configuration from instance environment variables.
+2. For the next release, when updating, if SAML configuration is set and the instance has a single organization, we set the same SAML parameters for that organization.
+3. Drop `SAML_ENFORCED` in favor of a `SSO_ENFORCEMENT` attribute that supports `saml`, `google-oauth2`, `github` & `gitlab` to solve [#7535](https://github.com/PostHog/posthog/issues/7535) too (only one per organization).
+4. Drop `SAML_DISABLED` support to make maintenance easier and avoid having multiple build flows. Drawback is that the FOSS version which does not support SAML will still require the SAML dependencies.
```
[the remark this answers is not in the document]
guidoiaquinti: Yes as this will simplify our CI/CD flow and codebase quite a bit.

requests-for-comments/2022-03-08-full-saml.md
```diff
@@ -0,0 +1,48 @@
...
+5. If SAML is enabled for at least one organization in the instance, the login page shows only the email input at first. Upon entering an email, we do an API request as shown below (where `enforced_sso` contains information on whether to enforce log in with a particular SSO provider [either SAML, Google, ...] or `null` if the user has no SSO enforced **or the user does not exist**).
```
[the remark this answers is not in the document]
mariusandra: we also need to ensure we don't break password managers with this

requests-for-comments/2022-03-08-full-saml.md
```diff
@@ -0,0 +1,48 @@
...
+7. Bugs [#5759 Respect redirect URL when logging in with SAML](https://github.com/PostHog/posthog/issues/5759) & [#5839 Fix team member invites with SAML](https://github.com/PostHog/posthog/issues/5839) should be addressed too.
+8. Generally, users with SSO enforced will likely only belong to a single organization (because you're probably using your work email). However, there's the edge case of a user belonging to multiple organizations with different SSO policies. Here's how we'll address them.
+    - If the user attempts to join an organization that has a different SSO policy from the one present in their current organization(s), we won't allow the user to join the new organization and show a warning explaining why.
+    - When enabling SSO enforcement for an organization, we'll show a warning to the admin signalling any users who belong to multiple organizations and be very clear that the SSO requirement will not be enforced for them unless they leave their other organizations.
```
mariusandra, line 43: > "we'll show a warning to the admin signalling any users who belong to multiple organizations"

Might this be considered as leaking data we shouldn't?

requests-for-comments/2022-03-08-full-saml.md
```diff
@@ -0,0 +1,48 @@
...
+6. If the user belongs only to an organization with SSO enforced (or multiple organizations with the same SSO enforcement policy), we won't allow logging in with password, changing password or log in with any other SSO mechanism. If the user requests a password reset we'll only include the information about which SSO provider to use but no link to reset the password.
+7. Bugs [#5759 Respect redirect URL when logging in with SAML](https://github.com/PostHog/posthog/issues/5759) & [#5839 Fix team member invites with SAML](https://github.com/PostHog/posthog/issues/5839) should be addressed too.
+8. Generally, users with SSO enforced will likely only belong to a single organization (because you're probably using your work email). However, there's the edge case of a user belonging to multiple organizations with different SSO policies. Here's how we'll address them.
+    - If the user attempts to join an organization that has a different SSO policy from the one present in their current organization(s), we won't allow the user to join the new organization and show a warning explaining why.
```
mariusandra, line 42: If I understand well, the issue we're trying to avoid is somebody malicious logging in as `bob@posthog.com` with a self-hosted or compromised SAML server... that will make us trust that they're actually who they say they are, and hence give them access to other orgs.

- This might be technically trickier, but what about showing another login prompt then? The goal being that you can log in to each org individually. Or should we log you out of one org before logging in to another?
- Should we also show e.g. a lock next to the organization in the org dropdown?
- Should we even restrict domains, so that only users with `@posthog.com` domains allowed for org X, and no other user anywhere in the app can use this domain? (edit: saw the last line in the RFC... so that's a "not yet")

mariusandra (commented):
Overall looks good, though I think we should spec out the security implications a bit more (comments inline)

requests-for-comments/2022-03-08-full-saml.md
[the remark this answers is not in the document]
paolodamico: ofc, generally this doesn't break password managers (try logging into Okta from our 1Password shared account), though it will require an extra click.

requests-for-comments/2022-03-08-full-saml.md
paolodamico: Now @mariusandra

mariusandra: @guidoiaquinti is the one whose input I'd listen to here. I haven't worked at a scale where this was a real threat I had to worry about. Guido, any best practices or suggestions regarding API limits from ye olde days that we could adopt as well? We are talking about a public endpoint.

requests-for-comments/2022-03-08-full-saml.md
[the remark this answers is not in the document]
paolodamico: Already done & merged.

requests-for-comments/2022-03-08-full-saml.md
paolodamico: Yeah makes sense @guidoiaquinti, let's do domain-based.

paolodamico:
Alright took a completely different approach here. This I believe addresses the security concerns too, but we end up with a mammoth of complexity here, which doesn't look good. I'm scheduling a sync call to discuss tomorrow with you.

requests-for-comments/2022-03-08-full-saml.md
```diff
@@ -0,0 +1,53 @@
...
+## Proposal
+
+### Self-hosted
+1. Self-hosted instances remain with single-tenant SAML support (i.e. only one SAML provider for each instance), but we drop `SAML_ENFORCED` in favor of a `SSO_ENFORCEMENT` instance-level attribute that supports `saml`, `google-oauth2`, `github` & `gitlab`. We redirect automatically to the provider's login page if SSO is enforced (solves [#7535](https://github.com/PostHog/posthog/issues/7535)).
+
+### PostHog Cloud
+1. On PostHog Cloud create a separate model for `OrganizationConfiguration` where we track all SAML parameters (ACS URL, Entity ID, X509 certificate and 4 attribute names) at the organization-level.
+2. Before SAML can be enabled, admins have to verify ownership of a specific domain by adding a TXT record to their domain with a random 32-byte secret. An organization can verify multiple domains, but a domain may belong to only one organization. `OrganizationDomain` model for this.
```
paolodamico, line 26: As this will be an enterprise feature, another simpler approach here would be to do what Notion does, we manually assign authorized domains to each organization (and validate manually).

mariusandra (approved):
Great changes! LGTM.

PR #56, opened 2022-06-02
Adding session analysis to insights

rcmarron:
This is in an RFC that proposes a specific approach to how we add session analysis to our insights product.

For additional context on the problem being solved and the customer needs, please read: https://github.com/PostHog/product-internal/pull/307

### DIFF
--- a/requests-for-comments/2022-06-01-session-analysis.md
+++ b/requests-for-comments/2022-06-01-session-analysis.md
@@ -1,0 +1,118 @@
+# Sessions in insights
+
+The goal of this document is to propose a specific approach to how we add session analysis to our insights product as defined in the [Product Context document](https://github.com/PostHog/product-internal/pull/307/files),
+
+For more details on the problem we're trying to solve and the customer needs, please reference that document.
+
+**Note:** The goal of this project is not to build a feature that allows users to find and explore specific sessions. Rather this feature is targeted at doing analysis on sessions in aggregate
+
+## Requirements
+
+We need to figure out how to allow users to do “session based analysis” inside our product. As defined in the document above, this means allowing users to do analysis on the following:
+
+1. Number of sessions where specific events occur over time
+2. Average session length over time
+3. Average number of sessions per user over time
+4. Session length distributions
+5. Count of a specific event per session over time
+6. Limit a funnel conversion to a single session
+7. Limit a path conversion to a single session
+
+## Possible Solutions
+
+At it’s highest level, there are 2 options here:
+
+1. Fitting session analysis into our existing insights features (mainly trends)
+2. Building out a new sessions feature tailor made for this analysis
+
+(tl;dr we should do option 1 because we can lean on lots of existing behavior of trends)
+
+The big trade-offs between these two options are:
+
+- Option 1 allows for lots of analysis that you “get for free” because of the power of trends. For example the following are all easily possible:
+    - Plot the number of sessions over time along with the count of users. Then use formulas to divide the 2 and get “sessions per user”
+    - Plot average session length where a users does Z broken down by browser, and notice that your feature is broken on IE.
+    - Plot the average session length where users do X and compare it to the average session length where users do Y. Notice that users that do X, stick around longer.
+    - Run an experiment where the goal is to increase the session length 🤯
+- Potentially, option 2 would allow for more complex sessions to be defined. Right now, the language of trends is about defining events, not sessions. As a result, when this language is translated to sessions, it means, “Give me sessions where this specific event occurs”. This restricts users from creating complex session definitions such as “sessions where X happens and the Y happens within 2 minutes”.
+
+My vote is that we **do option 1 and build this analysis into the existing trends product**. The benefits of building these features into trends greatly outweighs the complex sessions a user could define in option 2. Also, there are future paths where we could allow users to define complex sessions within option 1.
+
+## How should these added analyses be built into trends?
+
+Looking at the various analyses that we want to allow, it seems like the requirements above could fit relatively cleanly into our existing insights UX. Below is a proposal for how we would enable all of the above defined requirements.
+
+### 1. Number of sessions where specific events occur over time
+
+Proposal: Inside trends, we add “Unique Session” as an ‘event aggregation’ option.
+
+![image](https://user-images.githubusercontent.com/4813045/171521247-45dfc5d4-d2c5-4ca4-9fec-f81cc0ace5e8.png)
+
+
+### 2. Average session length over time
+
+Proposal: We allow users to do property aggregations based on “session length”
+
+![image](https://user-images.githubusercontent.com/4813045/171521195-121300f6-48a9-488c-956c-f176243865a7.png)
+
+Details:
+- In addition to showing event properties in the taxonomic filter when a property aggregation is selected, we also show “session properties”. Within session properties, we would show session length.
+- If a session property is the target of the property aggregation, then the query would be aggregated by sessions.
+
+### 3. Average number of sessions per user over time
+
+Proposal: Users accomplish this using formulas. They would make a series that is (a) Session count and (b) user count, and then divide (a) by (b).
+
+![image](https://user-images.githubusercontent.com/4813045/171521323-74e94328-1b21-45c8-925f-d9cc3aab6725.png)
+
+### 4. Session length distributions
+
+Proposal: We add session length as an option in the breakdown properties. Additionally, we allow users to “auto-bucket” breakdowns that are on numeric properties. 
+
+![image](https://user-images.githubusercontent.com/4813045/171521347-8f3fd16c-28ab-49ac-9baa-c02097d4c42c.png)
+
+Details: The “auto-bucket” functionality would require more work to figure out exactly how this should work + what the UX should be.
+
+### 5. Count of a specific event per session over time
+
+Proposal: Users accomplish this using formulas. They would make a series that is (a) Session count and (b) total count, and then divide (b) by (a).
+
+![image](https://user-images.githubusercontent.com/4813045/171521365-2b353bb0-7634-45fa-84e3-1463eb351fc0.png)
+
+### 6. Limit a funnel conversion to a single session
+
+Proposal: In funnels, we add “Unique sessions” as an aggregation option. Setting this value would enforce that the sequence of events must occur within a specific session.
+
+![image](https://user-images.githubusercontent.com/4813045/171521401-3b7a234c-24b4-45cd-a3d0-15ab4243efb0.png)
+
+
+### 7. Limit a path conversion to a single session
+
+Proposal: I’d consider this to be a lower priority, but we should update the logic in paths to respect the concept of a session. Today, it uses a 30 min gap between events to split paths - which is essentially the same as sessions, but subtly different.
+
+One issue that we will have to figure out is being backward compatible to data that was generated prior to session support.
+
+## Work breakout:
+
+To accomplish the above, we would need to do the following work:
+
+- Trends:
+    - Add “unique sessions” as an event aggregation option in trends
+    - Add the concept of “session properties”, and specifically the “session length” property to the taxonomic filter.
+    - Allow for session properties to be used as a property aggregation in trends.
+    - Allow for session properties to be used in breakdowns
+    - Add the ability to “auto bucket” when breaking down by numeric properties
+    - Note: while not included in the requirements above, it’d be cool to allow filtering based on session length everywhere. (e.g. total count of pageviews where the corresponding session length > 1 min)
+- Funnels
+    - Add sessions as a aggregation target
+- Paths
+    - Use sessions instead of the 30 min gap (need to consider backward compatibility here)
+
+## Questions + notes
+
+1. Do we need a sessions table?
+    1. If we add session length as a property option, those queries will have to generate a list of session durations on the fly. This query can’t be limited to specific event types because it has to query across all events to ensure it finds the first and last event in the session. As a result it might be a hefty query. To optimize this, we could create a sessions table in clickhouse (probably using AggregatingMergeTree) ****that contains the session_id and it’s length
+    2. My vote is that we don’t over optimize early and try to do without this table at first. We can always add it later.
+2. There will be a mismatch in recording duration and session duration
+    1. Because event based sessions and recording based sessions are based on different events, there will not be a perfect match when comparing the start and end time of each. This could cause some confusing user experiences.
+    2. One solution here could be leaning on a sessions table as described above, and having it reconcile the difference between these two data sources to ensure consistency.


rcmarron:
@neilkakkar I didn't directly address the "is a session a group" question above, but I think the described approach aligns with your thinking.

The one detail that I wasn't certain about is how the "session length" property would work. It seems like it's something like a "calculated group property"? I think we can treat that as an implementation detail though.

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+
+The big trade-offs between these two options are:
+
+- Option 1 allows for lots of analysis that you “get for free” because of the power of trends. For example the following are all easily possible:
+    - Plot the number of sessions over time along with the count of users. Then use formulas to divide the 2 and get “sessions per user”
+    - Plot average session length where a users does Z broken down by browser, and notice that your feature is broken on IE.
+    - Plot the average session length where users do X and compare it to the average session length where users do Y. Notice that users that do X, stick around longer.
+    - Run an experiment where the goal is to increase the session length 🤯
+- Potentially, option 2 would allow for more complex sessions to be defined. Right now, the language of trends is about defining events, not sessions. As a result, when this language is translated to sessions, it means, “Give me sessions where this specific event occurs”. This restricts users from creating complex session definitions such as “sessions where X happens and the Y happens within 2 minutes”.
```
neilkakkar, line 37: True, but the example is a funnel aggregated by sessions, so should be possible in the new view. We need another example of a complex session that isn't possible using any existing insights lol.

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+    - Add sessions as a aggregation target
+- Paths
+    - Use sessions instead of the 30 min gap (need to consider backward compatibility here)
+
+## Questions + notes
+
+1. Do we need a sessions table?
+    1. If we add session length as a property option, those queries will have to generate a list of session durations on the fly. This query can’t be limited to specific event types because it has to query across all events to ensure it finds the first and last event in the session. As a result it might be a hefty query. To optimize this, we could create a sessions table in clickhouse (probably using AggregatingMergeTree) ****that contains the session_id and it’s length
+    2. My vote is that we don’t over optimize early and try to do without this table at first. We can always add it later.
+2. There will be a mismatch in recording duration and session duration
```
neilkakkar, line 116: hmmm, tricky! "Session" Recordings sounds like a misnomer now 😂. Trying to reconcile these two when they're going to be completely different sounds like a losing battle...

So, question: Do users find it valuable to align these? What problem does this solve? Or is it more of a consistency issue?

neilkakkar:
> I didn't directly address the "is a session a group" question above, but I think the described approach aligns with your thinking.

It does align! This looks great!

I'm curious though why we're sidestepping this question?

requests-for-comments/2022-06-01-session-analysis.md
neilkakkar: valid example!

Aside: I'm not sure it requires something dedicated to solve this. For example, this sounds like a trend graph, with property aggregation of average, on the property session length, and filtered by a cohort of sessions created using the funnel.

Basically, the missing primitive is 'cohorts of groups', which will solve this problem for all groups, not just sessions.

But yeah, agree that these kinds of complex queries support aren't required yet.

its-all-connected-everything-is-connected

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+- Funnels
+    - Add sessions as a aggregation target
+- Paths
+    - Use sessions instead of the 30 min gap (need to consider backward compatibility here)
+
+## Questions + notes
+
+1. Do we need a sessions table?
+    1. If we add session length as a property option, those queries will have to generate a list of session durations on the fly. This query can’t be limited to specific event types because it has to query across all events to ensure it finds the first and last event in the session. As a result it might be a hefty query. To optimize this, we could create a sessions table in clickhouse (probably using AggregatingMergeTree) ****that contains the session_id and it’s length
+    2. My vote is that we don’t over optimize early and try to do without this table at first. We can always add it later.
```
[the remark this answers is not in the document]
EDsCODE: I think this would get tricky with backwards compatibility 😬

requests-for-comments/2022-06-01-session-analysis.md
EDsCODE: I think there's value in the continuity of analysis. It seems that we're grafting in sessions to insights so that analysis from insights to sessions is super smooth and we're choosing that priority over session aggregates to recordings. Ideally we will allow for both ways of analysis, if you're on recordings and checking out specific sessions with some characteristics, it'd be awesome to then be able to ask "how many sessions like this have happened over time" directly from recordings. Although, maybe the time alignment isn't the biggest deal here

EDsCODE (commented):
I'm still torn on how on whether or not sessions analysis should go in insights. It seems most natural to connect it to recordings. Implementationally it does seem more satisfying to build it in to most of the syntax we have (into insights + reusing groups) but definitely worried that's an engineering thinking trap. For example, is it intuitive for users to expect sessions to be analogous to a group and for the breakdown by functionality to create the time distribution graph? That seems like we're hiding an easter egg for what should be a primary feature. 

We do have default properties we capture on our default events, but adding those properties is a convenience because users could technically use our client libraries and add it themselves. Session length doesn't seem like a "property" that a user would be able to instrument in themselves so it's unnatural to expect it to be hiding in properties. Also, are we allowing for users to add custom session properties somehow?

requests-for-comments/2022-06-01-session-analysis.md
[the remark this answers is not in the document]
neilkakkar: I'd call it a completely new feature? 🤔 - do we currently have sessions exposed in app?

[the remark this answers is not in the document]
EDsCODE: I suppose not although just feelsbad if the graph is empty before a certain time

neilkakkar:
Hmm, sounds like we'd want to distinguish two things. Is this feature:

1. A recordings companion, to analyse / aggregate "sessions" in recordings? In which case, we should be using the exact same events & crafting everything to work specific to this use case. Or,
2. A new feature for enabling time-based analysis on user-defined / our default-defined groups, the prime example being 'user activity with < 30 mins of sleep time', which sometimes is synonymous with recordings.

EDsCODE:
IMO, we need to find a path for both. I'm also all for continuing down this implementation as there's already been a good amount of thought put into it that I trust. The burden will fall upon designing and communicating the feature clearly

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+    - Note: while not included in the requirements above, it’d be cool to allow filtering based on session length everywhere. (e.g. total count of pageviews where the corresponding session length > 1 min)
+- Funnels
+    - Add sessions as a aggregation target
+- Paths
+    - Use sessions instead of the 30 min gap (need to consider backward compatibility here)
+
+## Questions + notes
+
+1. Do we need a sessions table?
+    1. If we add session length as a property option, those queries will have to generate a list of session durations on the fly. This query can’t be limited to specific event types because it has to query across all events to ensure it finds the first and last event in the session. As a result it might be a hefty query. To optimize this, we could create a sessions table in clickhouse (probably using AggregatingMergeTree) ****that contains the session_id and it’s length
```
EDsCODE, line 114: I used to do this madness after calcluating the 30minute window so with session_id alone, the query should actually be ok

[Basic query example](https://metabase.posthog.net/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7InR5cGUiOiJuYXRpdmUiLCJuYXRpdmUiOnsicXVlcnkiOiJTRUxFQ1QgZGF0ZV9kaWZmKCdzZWNvbmQnLCBtaW5fdGltZSwgbWF4X3RpbWUpIGFzIHNlY29uZHMsIG1heCh0aW1lc3RhbXApIG1heF90aW1lLCBtaW4odGltZXN0YW1wKSBtaW5fdGltZSBGUk9NIGV2ZW50cyB3aGVyZSB0ZWFtX2lkID0gMiBhbmQgTk9UICRzZXNzaW9uX2lkID0gJycgQU5EIHRpbWVzdGFtcCA-ICcyMDIyLTA1LTAxIDIzOjU5OjU5JyBHUk9VUCBCWSAkc2Vzc2lvbl9pZCIsInRlbXBsYXRlLXRhZ3MiOnt9fSwiZGF0YWJhc2UiOjJ9LCJkaXNwbGF5IjoidGFibGUiLCJ2aXN1YWxpemF0aW9uX3NldHRpbmdzIjp7fX0=)

neilkakkar: not bad!

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+One issue that we will have to figure out is being backward compatible to data that was generated prior to session support.
+
+## Work breakout:
+
+To accomplish the above, we would need to do the following work:
+
+- Trends:
+    - Add “unique sessions” as an event aggregation option in trends
+    - Add the concept of “session properties”, and specifically the “session length” property to the taxonomic filter.
+    - Allow for session properties to be used as a property aggregation in trends.
```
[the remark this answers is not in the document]
rcmarron: Right now, we only support event properties for property aggregations, there aren't group properties there. For this feature specifically, we would need to extend this to include session length. I think the decision on whether this means we're adding all group properties here or not depends on how we're implementing session length (specifically, is session length a "group property" or something we're calculating on the fly)

image

requests-for-comments/2022-06-01-session-analysis.md
rcmarron: > Although, maybe the time alignment isn't the biggest deal here

I agree with this. Also, with PageView and PageLeave events, these two durations should be very close to each other most of the time.

requests-for-comments/2022-06-01-session-analysis.md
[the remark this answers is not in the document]
rcmarron: This seems really similar to a `first_seen_at` and `last_seen_at` property. I think several users have [asked for this](https://github.com/PostHog/posthog/issues/9683) on persons (and probably other groups). If we were to implement it, I think you're right that it would have to be special cased - because it doesn't make sense with person-on-events. A few other considerations are probably:
* How would we do a backfill on it?
* How would we handle out-of-order events or (event backfills from 3rd parties as a prime example). I think we would just do a max/min function for this one.

I'd be very curious to get the thoughts of someone from the ingestion team on this one. @tiina303 or @yakkomajuri

rcmarron:
> I'm curious though why we're sidestepping this question?

I just wanted to make sure the approach made sense, and to have that discussion separately from the "is this just a group" discussion.

---

Sounds like we're all generally aligned on the concepts above, but there are a few details to be further ironed out. Specifically, we should figure out:

### 1. If we're treating a session as a group, how far do we take this concept?

On one end of the spectrum, we just add `$group_6`, and it's a special cased as "session", and on the other end of the spectrum, sessions are a concept that we treat very similarly to groups, but has its own code paths.

A couple specific considerations that come to mind:
* We already have a `$session_id` property in events, and this has been tracking sessions for some time. How should this be reconciled with the rest of the groups feature?
* There are parts of the front-end where we definitely don't want sessions to appear with the other groups. For example, in retention analysis. So at least on the front-end, sessions will be special-cased.
* In our system, a person is a "grouping of events" outside of "groups". What distinguishes what should be a "group" vs. what's just a "grouping of events".
* In the client libraries, sessions are definitely a different concept than a group (because they're managed by the inactivity gap), but I could imagine a world where users want to add their own "session properties". 

My gut tells me that the right approach here is for sessions to be treated very similarly to a group, but that it's a special case - similar to persons. With that said, I'm not very familiar with the groups work, so I'd definitely lean on other's opinions here.

@macobo @neilkakkar 

### 2. How the "session duration" property will work.

For much of the analysis above, we need to add the concept of session duration. From what I can tell, we have a couple options here:

1. We treat the session duration as a "calculated property" - it would be calculated on the fly and joined into the queries as needed.
2. We create a "sessions" table that has the duration "pre-calculated" - maybe using an `AggregatingMergeTree`
3. We track the session duration as a proper session property as @neilkakkar had described above

My opinion is that we should explore option 3, but my gut tells me that the backfill, person-on-events, and out-of-order event complexities described above likely mean it's not trivial and will require support from the ingestion team (who seem to be fighting battles right now).

If that's the case, I think we can start with option 1 and move to option 3 when appropriate.

### 3. The UX + implementation details of "auto-binning"

We have a new concept of "auto binning" described above. We need to figure out what the UX for how this interaction would work. The basic idea of the feature is that if the user breaks down by a numeric or datetime property, we would "auto bin" the results. So instead of breaking down by 1000 values, those values are grouped into a specific number of bins and those bins are shown to the user. A few specific questions:
1. Should we auto-bin by default anytime a breakdown by a numeric/datetime property is done
2. Should we make the number of bins configurable?
3. Should users be able to "undo" the binning
4. How do we make it obvious that the results are being binned?

I'd like to lean on @clarkus for help with this one.

EDsCODE:
RE 2
The aggregatingmergetree is pretty much made for this purpose so I would suggest we explore the feasibility of that. It's a materialized table too so we can easily up and down it without much concern.

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
+# Sessions in insights
+
+The goal of this document is to propose a specific approach to how we add session analysis to our insights product as defined in the [Product Context document](https://github.com/PostHog/product-internal/pull/307/files),
+
+For more details on the problem we're trying to solve and the customer needs, please reference that document.
+
+**Note:** The goal of this project is not to build a feature that allows users to find and explore specific sessions. Rather this feature is targeted at doing analysis on sessions in aggregate
+
+## Requirements
```
macobo, line 9: Question: Could we also add a client motivation section? What are the questions users are trying to answer. Without this, this list of requirements seems incomplete.

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+
```
macobo, line 10: We haven't defined what a "session" is which is key to success here.

- How is the "border" between sessions drawn? Is it a timeout, dependent on devices, etc?
- Does it only include events from the same library?
- Can sessions retroactively change if events are being backfilled?

requests-for-comments/2022-06-01-session-analysis.md
macobo: This is answered in https://github.com/PostHog/product-internal/pull/307/files

requests-for-comments/2022-06-01-session-analysis.md
macobo, line 10: We haven't defined what a "session" is which is key to success here.

- How is the "border" between sessions drawn? Is it a timeout, dependent on devices, etc?
- Does it only include events from the same library?
- Can sessions retroactively change if events are being backfilled?

cc @marcushyett-ph since this is as much of a product question than a technical one.

requests-for-comments/2022-06-01-session-analysis.md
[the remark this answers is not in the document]
macobo: Instead of special-casing this, let's build out general groups support for both. By "internally" treating sessions as groups we're able to enhance both use-cases.

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+Proposal: I’d consider this to be a lower priority, but we should update the logic in paths to respect the concept of a session. Today, it uses a 30 min gap between events to split paths - which is essentially the same as sessions, but subtly different.
+
+One issue that we will have to figure out is being backward compatible to data that was generated prior to session support.
+
+## Work breakout:
+
+To accomplish the above, we would need to do the following work:
+
+- Trends:
+    - Add “unique sessions” as an event aggregation option in trends
```
macobo, line 100: Would get for free as groups-as-sessions.

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+    - Add the concept of “session properties”, and specifically the “session length” property to the taxonomic filter.
```
macobo, line 101: Would get for free as groups-as-sessions.

requests-for-comments/2022-06-01-session-analysis.md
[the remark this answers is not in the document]
macobo: Let's implement these initially as 'computed' properties rather than a separate denormalization initially. If performance is bad we can change ingestion to suit this, but queries are 100x easier to iterate than ingestion.

requests-for-comments/2022-06-01-session-analysis.md
macobo: We call them "recordings" now so I believe a note in documentation somewhere explaining this is ~fine.

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+
+### 5. Count of a specific event per session over time
+
+Proposal: Users accomplish this using formulas. They would make a series that is (a) Session count and (b) total count, and then divide (b) by (a).
+
+![image](https://user-images.githubusercontent.com/4813045/171521365-2b353bb0-7634-45fa-84e3-1463eb351fc0.png)
+
+### 6. Limit a funnel conversion to a single session
+
+Proposal: In funnels, we add “Unique sessions” as an aggregation option. Setting this value would enforce that the sequence of events must occur within a specific session.
```
macobo, line 84: Free with groups-as-sessions.

macobo:
> If we're treating a session as a group, how far do we take this concept?

IMO this is not complicated at all.

The two important properties as sessions-on-groups are:
- it allows us to minimize ingestion-level changes. 
- It allows code reuse for endpoints, aggregation methods and more

This helps us avoid re-building things for similar features and get this out the door faster. The rest of the overlap is more incidental.

For the rest of special-casing, we basically add a new "property" to each group type for "is_session" or equivelent which:
- Which we can use to determine which "column" to read (if we want to use a different column than $group_X) in property.py and elsewhere.
- We can use to hide certain groups on features features.

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+- Paths
+    - Use sessions instead of the 30 min gap (need to consider backward compatibility here)
+
+## Questions + notes
+
+1. Do we need a sessions table?
+    1. If we add session length as a property option, those queries will have to generate a list of session durations on the fly. This query can’t be limited to specific event types because it has to query across all events to ensure it finds the first and last event in the session. As a result it might be a hefty query. To optimize this, we could create a sessions table in clickhouse (probably using AggregatingMergeTree) ****that contains the session_id and it’s length
+    2. My vote is that we don’t over optimize early and try to do without this table at first. We can always add it later.
+2. There will be a mismatch in recording duration and session duration
+    1. Because event based sessions and recording based sessions are based on different events, there will not be a perfect match when comparing the start and end time of each. This could cause some confusing user experiences.
```
timgl, line 117: Does it make sense to exclude backend events from sessions? There's a risk some scheduled backend events will screw up session stats, but some folks will only have backend events

requests-for-comments/2022-06-01-session-analysis.md
macobo: Also kind of covered in the sped: https://github.com/PostHog/product-internal/pull/307/files

requests-for-comments/2022-06-01-session-analysis.md
```diff
@@ -0,0 +1,118 @@
...
+
+![image](https://user-images.githubusercontent.com/4813045/171521195-121300f6-48a9-488c-956c-f176243865a7.png)
+
+Details:
+- In addition to showing event properties in the taxonomic filter when a property aggregation is selected, we also show “session properties”. Within session properties, we would show session length.
+- If a session property is the target of the property aggregation, then the query would be aggregated by sessions.
+
+### 3. Average number of sessions per user over time
+
+Proposal: Users accomplish this using formulas. They would make a series that is (a) Session count and (b) user count, and then divide (a) by (b).
```
andyvan-ph, line 64: Some driveby feedback from a user here. I feel like 'sessions per user' is too important a metric to be reying on formulas. 

It makes sense logically, but I can easily imagine some users will need help disovering the solution to how you get this metric, when it ought to be obvious and require no cognitive load on the user.

clarkus (approved):
Looks like a good start to me. The proposed UX changes for exposing session properties and aggregation make sense to me.

requests-for-comments/2022-06-01-session-analysis.md
neilkakkar: Fair point. More generally, 'users per organisation', 'paymentIDs per user', 'paymentIDs per organisation', i.e. count / math operation on one group type, aggregated by another group type would be a really nice to have.

I think it's possible, but raises a few code complications (which would get less complicated once person-on-events exists), so I agree with the decision to push it down to the future, specially since there's a "hack" available to get the same results, and there's an open question of how the UI for this should work.

PR #57, opened 2022-06-10
RFC: Nail Exports

benjackwhite:
TODO:
- [x] Release Slack App in Slack Directory 

Remaining issues:
- [ ] Exporting CSVs is not enabled everywhere
- [ ] Exporting large CSVs perhaps still doesn't work
- [ ] Embedding / sharing of images
- [ ] Using shared images in meta for Sharing html

### DIFF
--- a/requests-for-comments/2022-06-10-nail-exports.md
+++ b/requests-for-comments/2022-06-10-nail-exports.md
@@ -61,8 +61,8 @@ Similar to SharedDashboards we currently have, it makes sense to have a unique r
 
 ---
 ### Images
-| insights ✅ | dashboards ✅ | person ❌ | cohorts ❌ | groups ❌ |
-|------------|--------------|----------|-----------|----------|
+| insights ✅ | dashboards ✅ | events ❌ | person ❌ | cohorts ❌ | groups ❌ |
+|------------|--------------|----------|-----------|-----------|----------|
 
 Images are important either for the user (e.g. to include in a powerpoint presentation) or for use by other export methods such as subscriptions
 
@@ -73,8 +73,8 @@ Images are important either for the user (e.g. to include in a powerpoint presen
 
 ---
 ### Subscriptions - email:
-| insights ✅ | dashboards ✅ | person ❌ | cohorts ❌ | groups ❌ |
-|------------|--------------|----------|-----------|----------|
+| insights ✅ | dashboards ✅ | events ❌ | person ❌ | cohorts ❌ | groups ❌ |
+|------------|--------------|----------|-----------|-----------|----------|
 
 Subscriptions via email are essentially Image exports but delivered via email
 
@@ -84,8 +84,8 @@ Subscriptions via email are essentially Image exports but delivered via email
 
 ---
 ### Subscriptions - slack:
-| insights ✅ | dashboards ✅ | person ❌ | cohorts ❌ | groups ❌ |
-|------------|--------------|----------|-----------|----------|
+| insights ✅ | dashboards ✅ | events ❌ | person ❌ | cohorts ❌ | groups ❌ |
+|------------|--------------|----------|-----------|-----------|----------|
 
 Subscriptions via slack are essentially Image exports but delivered via slack. We may want to consider additional information to be included
 
@@ -96,13 +96,14 @@ Subscriptions via slack are essentially Image exports but delivered via slack. W
 
 ---
 ### CSV / Excel:
-| insights ✅ | dashboards ✅ | person ✅ | cohorts ✅ | groups ✅ |
-|------------|--------------|----------|-----------|----------|
+| insights ✅ | dashboards ✅ | events ✅ | person ✅ | cohorts ✅ | groups ✅ |
+|------------|--------------|----------|-----------|-----------|----------|
 
-Resources can be exported directly to a CSV
+Resources can be exported directly to a CSV. This may be commonly done adhoc (e.g. from some in-memory filter configured on the frontend.
 
 #### Technical considerations
 - [ ] Exporting 10000s of rows can be costly or even impossible syncronously - it needs to be possible to do this in the background and send to the user when complete
+  - [ ] This could also incur very high costs. Perhaps we still need some upper bounds.
 - [ ] TTLs: Exported data can include very sensitive information. Ideally we should only keep the exported file for long enough to satisfy the user's request
 
 ---
--- a/requests-for-comments/2022-06-10-nail-exports.md
+++ b/requests-for-comments/2022-06-10-nail-exports.md
@@ -112,3 +112,17 @@ Resources can be exported directly to a CSV. This may be commonly done adhoc (e.
 |------------|--------------|----------|-----------|-----------|----------|
 
 The API should offer all of the same data access as the web app
+
+
+# Sprint breakdown
+## Sprint 1
+
+- [x] Subscriptions - Email - Insights
+- [ ] Subscriptions - Email - Dashboards
+- [ ] Image - Dashboards / Insights improvements
+
+## Sprint 2
+- [ ] Subscriptions - Slack - Insights
+- [ ] Subscriptions - Slack - Dashboards
+- [ ] CSV - everything
+- [ ] Embeds - single embed style for any export
--- a/requests-for-comments/2022-06-10-nail-exports.md
+++ b/requests-for-comments/2022-06-10-nail-exports.md
@@ -1,5 +1,11 @@
 # Request for comments: Nail Exports - Ben White
 
+## NOTE: This RFC was closed on 2022-07-14 with these agreed remaining follow up issues
+- [ ] Exporting CSVs is not enabled everywhere
+- [ ] Exporting large CSVs perhaps still doesn't work
+- [ ] Embedding / sharing of images
+- [ ] Using shared images in meta for Sharing html
+  
 ## Problem statement
 
 More context here https://github.com/PostHog/product-internal/pull/308/
--- a/requests-for-comments/2022-06-10-nail-exports.md
+++ b/requests-for-comments/2022-06-10-nail-exports.md
@@ -8,33 +8,33 @@ More context here https://github.com/PostHog/product-internal/pull/308/
 ### What can be Exported
 
 * Insights
-  - Visualisation - web, image
-  - Data - API, csv
+  - ✅ Visualisation - web, image
+  - ✅ Data - API, csv
 * Dashboards
-  - Visualisation - web, image (or collection of images)
-  - Data - API, csv
+  - ✅ Visualisation - web, image (or collection of images)
+  - ✅ Data - API, csv
 * Events
-  - the actual underlying data 
-* Session Recordings
-* Persons, Cohorts, Groups
+  - ✅ the actual underlying data 
+* (🚫 out of scope) Session Recordings
+* ✅ Persons, Cohorts, Groups
 
 ### How can they be Exported
-* Embeds 
+* ✅ Embeds 
   - Via an iframe or directly linked to a dedicated webpage for viewing the element 
-* Images
+* ✅ Images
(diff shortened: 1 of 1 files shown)


requests-for-comments/2022-06-10-nail-exports.md
```diff
@@ -0,0 +1,110 @@
...
+
+---
+### CSV / Excel:
+| insights ✅ | dashboards ✅ | person ✅ | cohorts ✅ | groups ✅ |
+|------------|--------------|----------|-----------|----------|
+
+Resources can be exported directly to a CSV
+
+#### Technical considerations
+- [ ] Exporting 10000s of rows can be costly or even impossible syncronously - it needs to be possible to do this in the background and send to the user when complete
```
macobo, line 102: Will we create one file or several in that case? 

Also note given exporting raw data is costly (esp on cloud) and creates some data control problems - should we have controls/limits around that?

requests-for-comments/2022-06-10-nail-exports.md
```diff
@@ -0,0 +1,110 @@
...
+  - Possibly needs multiple dimension options
+* Subscriptions - email:
+  - User configured regular reports via email including exported image of the relevant item
+* Subscriptions - slack:
+  - Slack notification including the exported image(s)
+* CSV / Spreadsheet:
+  - Relevant data exported to a CSV / Spreadsheet
+* Custom Components (React)
+  - Integrated directly into customer's code, calling Posthog API for the data 
+* Alerts (variation on subscriptions):
```
macobo, line 32: Note this feature seems unrelated to the rest of this RFC and not sure how this is achieved. Should this be tracked/solved separately?

benjackwhite: Only mentioned it here as one of the things we are trying to achieve with this RFC is to help pick a technical solutions that will support future use-cases and not back is into multiple unique solutions per-case 👍  
```suggestion
* Alerts (variation on subscriptions): _(out of scope for current exporting work)_
```

requests-for-comments/2022-06-10-nail-exports.md
benjackwhite: It's a good point. Commited that point to the md

requests-for-comments/2022-06-10-nail-exports.md
```diff
@@ -96,13 +96,14 @@ Subscriptions via slack are essentially Image exports but delivered via slack. W
...
+| insights ✅ | dashboards ✅ | events ✅ | person ✅ | cohorts ✅ | groups ✅ |
+|------------|--------------|----------|-----------|-----------|----------|
 
-Resources can be exported directly to a CSV
+Resources can be exported directly to a CSV. This may be commonly done adhoc (e.g. from some in-memory filter configured on the frontend.
 
 #### Technical considerations
 - [ ] Exporting 10000s of rows can be costly or even impossible syncronously - it needs to be possible to do this in the background and send to the user when complete
+  - [ ] This could also incur very high costs. Perhaps we still need some upper bounds.
 - [ ] TTLs: Exported data can include very sensitive information. Ideally we should only keep the exported file for long enough to satisfy the user's request
```
macobo, line 107: Should we include data control options around this (e.g. this requiring some scope of user access)? Exporting all raw data via UI might be potential security risk at some organizations.

requests-for-comments/2022-06-10-nail-exports.md
```diff
@@ -0,0 +1,114 @@
...
+
+---
+### CSV / Excel:
+| insights ✅ | dashboards ✅ | events ✅ | person ✅ | cohorts ✅ | groups ✅ |
+|------------|--------------|----------|-----------|-----------|----------|
+
+Resources can be exported directly to a CSV. This may be commonly done adhoc (e.g. from some in-memory filter configured on the frontend.
+
+#### Technical considerations
+- [ ] Exporting 10000s of rows can be costly or even impossible syncronously - it needs to be possible to do this in the background and send to the user when complete
```
kappa90, line 105: I would opt for a default behavior - any csv gets sent via email. Easier to implement, more unified experience regardless of csv size.

benjackwhite: Downside of this would be:
1. Self-hosted doesn't always have email configured
2. Alot of csv exports _will_ be small enough for a quick easy export. I'm thinking we already have the async flow for image exports so it wouldn't be hard to manage it that way as well.

Maybe if the export takes so long that an email is required, we should reconsider if we even support it...?

requests-for-comments/2022-06-10-nail-exports.md
```diff
@@ -0,0 +1,114 @@
...
+* Images
+  - Server-side rendering of the shared webpage (using chromedriver) to an image (.png)
+  - Possibly needs multiple dimension options
+* Subscriptions - email:
+  - User configured regular reports via email including exported image of the relevant item
+* Subscriptions - slack:
+  - Slack notification including the exported image(s)
+* CSV / Spreadsheet:
+  - Relevant data exported to a CSV / Spreadsheet
+* Custom Components (React)
```
marcushyett-ph, line 33: How is this different from Embeds?

requests-for-comments/2022-06-10-nail-exports.md
```diff
@@ -0,0 +1,114 @@
...
+  - Integrated directly into customer's code, calling Posthog API for the data 
+* Alerts (variation on subscriptions): _(out of scope for current exporting work)_
+  - When certain thresholds are crossed or changes happen, a user is notified with similar channels to that of Subscriptions
+* API:
```
marcushyett-ph, line 37: Is this out of scope in terms of what we're planning on building right now? (i.e. already there / good enough)

requests-for-comments/2022-06-10-nail-exports.md
```diff
@@ -0,0 +1,114 @@
...
+
+Images are important either for the user (e.g. to include in a powerpoint presentation) or for use by other export methods such as subscriptions
+
+#### Technical considerations
+- [ ] Some things are hard/impossible to render nicely in small sizes (e.g. Funnels).
+- [ ] 
+
+
+---
+### Subscriptions - email:
```
marcushyett-ph, line 75: I would definietly prioritize Subscriptions over some of the others like Embeds at least for the next sprint

requests-for-comments/2022-06-10-nail-exports.md
benjackwhite: The Custom react components? From an implementation perspective completely different - embeds are an easy-ish one-liner iframe, react (or vue, svelte etc.) components means creating a separate npm module, custom code etc.

requests-for-comments/2022-06-10-nail-exports.md
benjackwhite: Which bit? (you highlighted 4 lines)
Alerts is definitely out of scope
API is only there as a holistic reminder that we have the API as a way of exporting data

marcushyett-ph: Not sure how 4 lines got highlighted, I only clicked one.

Was just highlighting API, so you answered my question :)

requests-for-comments/2022-06-10-nail-exports.md
```diff
@@ -0,0 +1,114 @@
...
+| insights ✅ | dashboards ✅ | events ❌ | person ❌ | cohorts ❌ | groups ❌ |
+|------------|--------------|----------|-----------|-----------|----------|
+
+Subscriptions via email are essentially Image exports but delivered via email
+
+#### Technical considerations
+- [ ] The exported Image needs to be publicly accessible using some sort of token
```
pauldambra, line 82: We _could_ embed the image as a data url. Then it doesn't have to publicly accessible from us which simplifies hosting.

But also means we can't track usage so that's probably a silly suggestion :)

requests-for-comments/2022-06-10-nail-exports.md
pauldambra: I'm imagining

* someone clicks to export
* we tell them we'll generate it and let them know
* and in that toast link them to a page where they can see their historical exports
* they'd have status like "expired" e.g. we deleted it but you can regenerate it
* and have download links
* then we can also email people to let them know a download is available if email is configured 
* but they have an onward journey if it isn't

mariusandra:
Been thinking about insights, iframes, subscriptions and exports... and came to the conclusion that in order to truly nail exports, we need a common interface for exports, subscriptions and shares, with clear audit controls.

To have publicly sharable links, they need to use access tokens, which we must store somewhere. We already have a system for shared dashboard access tokens, and exported asset access tokens. Now I feel we need to build a third one for shared insights... or find ways to consolidate. 

Could we consolidate?

I'm imagining an interface something like [Paul's idea here](https://github.com/PostHog/meta/pull/57#discussion_r894630985), but expanded to cover all our export types.

Said visually, it'd be a cross between the subscription management/editing views:

image
image

and this:

image

... with data similar to this:

image

It would show where all the data from my posthog instance has been exported to, by which means, and which exports are active. If I'd like, I could revoke access.

Instead of the separate "share", "export", "subscribe" buttons, there'd be one "share" link that'd let you add new exports (shares?), or manage old ones.

When adding a new export, you first get a choice between
- e-mail one-offs
- e-mail subscriptions
- embedded iframes
- shared "full screen" dashboard URLs
- one-off image downloads of various sizes
- stable public URLs of images of various sizes (I can imagine the slack integration using these)
- stable public URLs for "json" data
- whatever else we come up with

Could this also be connected to CSV exports in some way... 🤔 

Any thoughts @benjackwhite @pauldambra @clarkus ?

In the context of embedded insights, I definitely feel like consolidating on a common export format is a better idea that adding a yolo `share_token` field into `posthog_dashboarditem` 😆

pauldambra:
If we only want to show someone what has been done then we could use the activity log

That is on insights (behind a flag) but not dashboards. 

you could log 
* X exported
* X shared
* subscription created for X
* subscription managed for X
* subscription triggered/processed for X

benjackwhite:
@mariusandra updated the PR with a comment description related to the discussion we had. What do you think? Feel free to propose the alternative you discussed (maybe as a commit suggestion) if you like

marcushyett-ph:
+1 to @clarkus definitely keen to have a detailled chat on this.

requests-for-comments/2022-06-10-nail-exports.md
```diff
@@ -0,0 +1,144 @@
...
+    | relation | relation  | the actual binary content | file\|s3://link | png\|pdf\|csv | time by which the content should be accesible |
+  - **Subscription**
+    | insight  | dashboard | title                | target_type  | target_value   | {...rules}                              |
+    |----------|-----------|----------------------|--------------|----------------|-----------------------------------------|
+    | relation | relation  | name of subscription | email\|slack | a@b.co \| #dev | columns todo with subscription calendar |
+  
+  - **SharingConfiguration**
+    | insight  | dashboard | token                   | permissions                    | iframe_origins | deleted                            |
+    |----------|-----------|-------------------------|--------------------------------|----------------|------------------------------------|
+    | relation | relation  | unique short-ish string | list_of_emails\|password\|sso? | only-me.com    | soft delete (like enable/disabled) |
```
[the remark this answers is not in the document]
mariusandra: Works for me 👍

benjackwhite:
> There's one thing we talked about which isn't here: image embeds. The goal is that I could put a regularly updating png chart inside my github markdown document.

If we came up with some schedule rule to auto-screenshot shared insights / dashboards, this would have the side benefit of making previewing / Slack unfurling way easier...

benjackwhite:
Thanks for the discussion everyone! I'm merging this as we have come as far as we need to with this RFC PR. Some follow up work is listed in the document and any new issues / conversations should be created as dedicated issues 🙌

PR #58, opened 2022-07-14
RFC 2022-07-14-design-system

benjackwhite:


requests-for-comments/2022-07-14-design-system.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Possible rules we could set (remember - this is a proposal to be commented on!)
+
+* Components never layout themselves - no margin or padding
+* Components never expose hacks (e.g. style properties)
+* Anything that isn't a Lemon Component should be styled only with utility classes
+
+### Technical questions
+* Do we continue building this free-form or do we look into using some base Design System library?
+* Do we consider something like Tailwind to aid with tooling, autocomplete, Figma-linking etc?
```
pauldambra, line 38: +1 to Tailwind

requests-for-comments/2022-07-14-design-system.md
```diff
@@ -0,0 +1,46 @@
...
+
+Q3 Objective: `50% reduction in custom and inline CSS styles.`
+
+A clear goal is that Designing and Developing the frontend should be faster, perhaps through it being more restrictive. Developers touching custom CSS files should be a rare occurence, both because the components provide all the flexibility required **and** because Designers are also not deviating from the agreed Design System options unless truly necessary.
+
+
+## Context
+*What are our competitors doing, what are the technical constraints, what are customers asking for, what does the data tell us, are there external motivations (e.g. launch week, enterprise contract)?*
+
+> TODO: Link to some good resources on pitfalls of UI library / Design System creation
```
pauldambra, line 24: https://www.coop.co.uk/experience-library/
https://digitalblog.coop.co.uk/2022/03/28/its-an-experience-library-because-design-system-implies-its-only-for-designers/
https://design-system.service.gov.uk/

pauldambra (commented):
a little brain dump after first read

requests-for-comments/2022-07-14-design-system.md
mariusandra: Lessons learned by DigitalOcean:
https://www.gabe.pizza/notes-on-component-libraries/

requests-for-comments/2022-07-14-design-system.md
mariusandra: I haven't used Tailwind in a professional setting, so these fears might be unwarranted, but...

Can we get by without using Tailwind, but with using our own subset of tailwind-inspired custom classes? I fear that adding all of Tailwind will either add as much CSS as antd, or require a custom CSS tree shaker plugin to get small bundles. The latter will require some non-standard approaches, as we use many JS bundlers (esbuild, webpack), and plan to use even more (vite?). Thus I'm greatly in favour of not adding more complexity to the build system.

requests-for-comments/2022-07-14-design-system.md
```diff
@@ -0,0 +1,46 @@
...
+### Possible rules we could set (remember - this is a proposal to be commented on!)
+
+* Components never layout themselves - no margin or padding
+* Components never expose hacks (e.g. style properties)
+* Anything that isn't a Lemon Component should be styled only with utility classes
+
+### Technical questions
+* Do we continue building this free-form or do we look into using some base Design System library?
+* Do we consider something like Tailwind to aid with tooling, autocomplete, Figma-linking etc?
+* Do we try and push developers to the right solution (eslint warnings for usage of `style={}`)
```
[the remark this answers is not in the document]
mariusandra: +1 to CSS warnings on `style` tags. We can likely add them for antd components as well (e.g. `<Row />`):

- https://github.com/jsx-eslint/eslint-plugin-react/blob/HEAD/docs/rules/forbid-dom-props.md
- https://github.com/jsx-eslint/eslint-plugin-react/blob/HEAD/docs/rules/forbid-elements.md

mariusandra (commented):
Added some notes as well. There's still a longer list of raw notes in [this shared doc](https://docs.google.com/document/d/12vqTSOgkj5bxAH6tx67uWrYrPfvn7xVzmzCQnbT9pwY).

kappa90:
Regarding Tailwind, @benjackwhite was suggesting this other library, [Windi)(https://windicss.org/), which is supposed to alleviate the tree shaking problems of Tailwind ([see here](https://github.com/windicss/windicss/discussions/176))

Regarding the custom compilers, I tried looking into it but recognized it I don't know well enough how the current system works to move everything to Vite. Would be worth discussing if we still need these custom compilers?

requests-for-comments/2022-07-14-design-system.md
benjackwhite: The important point for me is that we follow Tailwind's API regardless of whether we use Tailwind or not.
As it is so well known (and the docs are great) it would be silly to have utility classes that are custom because it's way too much mental overhead. 

Tailwind itself has lots of great stuff such as auto-completion, breakpoints `sm:foo md:bar` etc. which would be worth getting in but isn't necessarily massively dependent on it.

There are plenty of ways to use Tailwind effectively so it may simply make sense to "try it out" and see what the damage is.

marcushyett-ph:
@benjackwhite [these are the most used elements with custom styles](https://metabase.posthog.net/question/391-elements-to-make-reusable), probably some of these should be combined into components as we prioritize here.

Would be great to get as specific as we can on the components we want to prioritize (especially for next sprint).

pauldambra:
I've used Tailwind with PostCSS and the treeshaking was great... But, yep, that does mean a new tool in the pipeline

ELanning:
As a complete outsider who has been involved in two design systems (one of them being [open source](https://github.com/Faithlife/styled-ui)), I only have a few thoughts.

1. TailWind is cool, but taking a long term perspective (5 to 10+ years), it will either be deprecated or not the new hotness. When this is the case, future maintainers will not be too keen on dealing with a niche legacy CSS system. CSS/LESS/SCSS are probably more safe from this effect than say, TailWind or styled-components. In the intermediate, TailWind is still being actively upgraded and released. I think TailWind 3.0 came out fairly recently. This is another "thing" that needs to be maintained and possibly frequently upgraded.
2. Under "who are we building for" I believe one answer is _other developers_. Developers want to get their job done without making changes to the design system repo ([closed for modification, open for extension](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle)). This is mentioned in the `gabe.pizza` link of developers coming in and adding "one off" props for their specific use case. I believe the reason this typically happens is owners of the design system repo being dogmatic about disallowing access to `styles`. Good abstractions allow "escape hatches", in React this is `refs`. Here, it would be access to `styles`. I recommend allowing access to `styles` (I like to do so via [`styled-system`](https://github.com/styled-system/styled-system), but pick your poison). Then every once in awhile taking a look at all places the `styles` are used and see if it can be refactored to be a prop in the design system.

PR #62, opened 2022-09-08
RfFC: Notifications

pauldambra:
See https://github.com/PostHog/posthog/issues/11368 for user interview context


requests-for-comments/2022-09-08-notifications.md
```diff
@@ -0,0 +1,81 @@
+# Request for comments: TITLE & OWNER
```
pauldambra, line 1:
```suggestion
# Request for comments: Nptifications & @pauldambra 
```

requests-for-comments/2022-09-08-notifications.md
```diff
@@ -0,0 +1,81 @@
+# Request for comments: Notifications & @pauldambra 
+
+## Problem statement
+*Who are we building for, what are their needs (include example cases), why is this important?*
+
+Users don't know when to come back to PostHog. Something important can happen and they might miss it. This can be important things in the data, or that another user has edited your insight.
```
neilkakkar, line 6: > that another user has edited your insight.

Q: Is it important to know about this when it happens? What do notifs add here over the audit log?

requests-for-comments/2022-09-08-notifications.md
```diff
@@ -0,0 +1,81 @@
...
+
+* number of triggers created
+* number of notifications sent
+* stickiness of dashboards and insights
+
+## Context
+*What are our competitors doing, what are the technical constraints, what are customers asking for, what does the data tell us, are there external motivations (e.g. launch week, enterprise contract)?*
+
+## Design 
+*What are the key user experience and technical design decisions / trade-offs?*
```
neilkakkar, line 23: RFC Feedback: I'm having a hard time parsing this (the diagram makes some sense after staring at it for a while, but lacks context like why this, what other alternatives we considered, and how it solves the notification problem)

requests-for-comments/2022-09-08-notifications.md
```diff
@@ -0,0 +1,81 @@
...
+    CS-->Slack
+    CS-->Email
+    Web ---|show notification list| NS
+    CS -->NS[(notification store)]
+```
+
+There are some designs in figma for comment https://www.figma.com/file/Y9G24U4r04nEjIDGIEGuKI/PostHog-Design-System-One?node-id=3112%3A1296
+
+### user considerations
+
```
neilkakkar, line 42: how real time do the notifications need to be?

Ex: are users okay if the notification occurs a day later? 

Would love to see what users want here, based on time delays.

requests-for-comments/2022-09-08-notifications.md
neilkakkar: Would be also nice to read about what challenges (technical & design wise) this solution might pose, that we're explicitly now accepting.

requests-for-comments/2022-09-08-notifications.md
```diff
@@ -0,0 +1,81 @@
...
+### Notifications people mentioned
+
+* chart goes above or below a limit
+    * expected to set this on an insight
+* for a single event if the value of a property is over a threshold
+* on particular calculated event
+* notification when same event occurs for a single user/group over time (cohort entry/exit?)
+* notification when experiment reaches significance
+
+### Tech considerations
```
neilkakkar, line 59: I'm guessing the core here is re-using subscriptions infra to send notifications as well to slack, not just images?

And the new bit is the notification store for in-app notifications, which are also triggered via celery?

Q: Can we re-use this to send custom in-app notifications? In the future, we might just want to use this to point users to something new ourselves ( see a very WIP dump: https://github.com/PostHog/product-internal/pull/346/files - i.e. it's a notification from posthog vs for a trigger you setup )

Just mentioning this so we know to keep in-app notifications not tightly coupled with the rest. It's unclear to me from the diagram that this is the case.

requests-for-comments/2022-09-08-notifications.md
neilkakkar, line 6: Hmm, thinking about this a bit more, 

> Users don't know when to come back to PostHog

The fact that they prefer `slack > in-app > email` seems to suggest to me that coming back isn't the full problem here? (if it were, why is there a higher preference to in-app over email notifications?)

But more like 2 different problems based on urgency: (not sure, hypothesising)
1. I want to know when something urgent has happened in `slack / custom webhook` so I can come to PostHog and investigate. Like alerts / when some event is over threshold / someone shared on HN and blog spiked.
2. I want to know what all has changed since I last logged in. These are in-app notifications about things I care about, like others editing my insights, my experiment becoming significant, etc. etc.

requests-for-comments/2022-09-08-notifications.md
```diff
@@ -0,0 +1,81 @@
...
+### user considerations
+
+* how to create the triggers for notifications
+* where to see configured triggers
+* linking from notification to appropriate onward journey
+
+* expectation of performance
+  * e.g. evaluate whether insight series have crossed a threshold could be very expensive
+
+### Notifications people mentioned
```
timgl, line 50: An in-app notification center doesn't seem the best place for any of these notifications to get to people. Do we actually need this, or is this just some emails we can send?

requests-for-comments/2022-09-08-notifications.md
clarkus: They add timeliness. The audit log is passive - you have to go check it to find new information. Notifications can push important information directly to the user regardless of their context in the product. Additionally notifications can span multiple contexts, so they can communicate a broader range of information in very specific contexts. Audit logs are often specific to a single feature, so they only cover that scope.

clarkus: > I want to know when something urgent has happened in slack / custom webhook so I can come to PostHog and investigate. Like alerts / when some event is over threshold / someone shared on HN and blog spiked.

Notifications can happen in multiple places concurrently. This is where preferences are a huge benefit. You might want some notifications to only go to email, while others you might want to go to slack (because they're more important). You might not want to send notifications for some scenarios. There's a lot of nuance here, so preferences help make the feature flexible enough to account for a broad range of use cases. 

> I want to know what all has changed since I last logged in. These are in-app notifications about things I care about, like others editing my insights, my experiment becoming significant, etc. etc.

This is one place where notifications might not be the best solution. If every change is communicated as a distinct event and has to be cleared from a list, it's going to be really tedious to catch up on changes quickly. This is where something like an activity log can be really useful. Something at a project-level that summarizes important changes since the last time you viewed it. There could be a lot of changes between sessions, so a centralize component might handle that higher volume of changes better. 

Another consideration would be digest notifications - a single notification that covers a broader area of changes. This is great if the user only wants to be notified of changes once day. This case also maps pretty well to generating an email or posting a digest message to slack.

requests-for-comments/2022-09-08-notifications.md
clarkus: The challenge with notifications are:
- Scope - Users belong to organizations, organizations can have multiple projects, users can be at any level in the hierarchy at any given time - what's the appropriate scope of notifications based on what we know about the user? If the user is in a project, would we notify them of events from another project? Only a subset of users might get notifications about billing or org management level changes. 
- Establishing what all justifies a notification. There's a really broad area to cover with our product, not all things will be of equal importance to a user, not all things are available depending on your plan / payment status / org role / etc. 
- Nailing the preferences - getting just enough configuration to cover the use cases discovered in interviews. This can be a really complex thing to figure out.
- Probably some other stuff... I will keep thinking about it.

requests-for-comments/2022-09-08-notifications.md
clarkus: It depends. Certain events might be so important that users want to be notified immediately. Others might want to rollup notifications into a digest that they only get once a day.

pauldambra (commented):
...

requests-for-comments/2022-09-08-notifications.md
pauldambra: Have split messaged and in-app notifications.....

requests-for-comments/2022-09-08-notifications.md
pauldambra: > RFC Feedback: I'm having a hard time parsing this (the diagram makes some sense after staring at it for a while, but lacks context like why this, what other alternatives we considered, and how it solves the notification problem)

I'll make another pass later. Hard drawing alone :)

pauldambra:
Note to self: need to add how users will understand why they're seeing notifications (whether in-app or slack/email)

Should they be able to opt-out and/or report a notification to us

pauldambra (commented):
break the diagram into steps

weyert:
Two use cases for this I can think of this would be useful:
  - Get notified when the feature flag value or filters have changed, to allow a backend service to be informed that it needs to reevaluate a feature flag [1]
  - Regulatory reasons, see https://github.com/PostHog/posthog/issues/3073#issuecomment-766922156 ability to store the change log somewhere outside of Posthog Cloud would help to ensure we keep the audit trail e.g. in BigQuery / Cloud Storage etc

[1] For example, when I have a ops feature flag to enable telemetry (tracing) or sampling ratio, I could get informed it has changed and reevalluate the feature flag on the backend or in a sidecar and then send a `SIGHUP` POSIX signall to the service. In theory we could also an agent protocol instead of POSIX signals, e.g. https://github.com/open-telemetry/opamp-spec/blob/main/specification.md

requests-for-comments/2022-09-08-notifications.md
```diff
@@ -0,0 +1,181 @@
...
+* when to send notifications? 
+  * See "Complexity in whether to send a message" below
+  * some users might want a particular trigger right away, others might want it in a roll-up email
+    * do we allow that complexity
+* where to show notifications?
+  * if we have both an in-app notification center for passive/low-urgency notifications do we also show active/high-urgency notifications there
+
+## Active/Urgent Notifications
+
+Sent to slack or email
```
weyert, line 136: Ability to call a url would be also be beneficial

pauldambra:
We're not working on this for a while (i think) so closing this for now

PR #66, opened 2022-11-03
RFC: Data exploration

mariusandra:
Finally took some time to think about data exploration.

The [existing MVP](https://github.com/PostHog/posthog/pull/11981) was cobbled together by letting the frontend call different endpoints with a similar set of filters. This RFC attempts to get to the core of the issue by proposing changes to the filters themselves. I'm very curious to hear if this is in any way in line with what @Twixes was thinking of with refactoring queries, and/or how could these efforts be combined.

I might be totally missing the mark here and proposing a cool solution for the sake of it, so please criticise this to shreds :).

Also, this is very backend-heavy. If you think this would benefit from more screenshots or frontend explanations, I can expand further. I tried to keep this somewhat scoped to the immediate next steps.


requests-for-comments/2022-11-02-data-exploration.md
```diff
@@ -0,0 +1,600 @@
...
+interface CorrelationNode extends Node {
+  type: "correlation";
+  correlation?: CorrelationOptions;
+  from?: FunnelNode;
+}
+```
+
+## Backend
+
+TODO: what is needed in the backend to make this possible?
```
EDsCODE, line 500: Drive by thought: The nested filter is already dramatically better than our current filter monster. 

I think the crux of the complexity is how we'll be handling queries of arbitrary matchups of concepts such as correlation of trends or breakdown on paths? Making sure that we can support arbitrary complexity in querying. Part of the reason the insights are so bespoke is because the queries can end up being formulated very differently. Really interested to hear the ideas on how to make queries composable as I think that will inform the feasibility of nested filters

requests-for-comments/2022-11-02-data-exploration.md
```diff
@@ -0,0 +1,600 @@
...
+
+- You can't drill down on insight person modals (e.g. filter the list further, break down by cohorts, etc).
+- Our tables are basic. E.g. The events table supports some customisation, but not enough. What about aggregations? Tables below insights contain different data in insight vs table mode. Persons tables are very basic. Math and formulas?
+- There are chart components that can't be added to dashboards (e.g. retention line chart)
+- Various code reasons (explained below) of why it's a bad idea to keep appending stuff to `FilterType`
+- Multiple endpoints accept various types of filters with various capabilities (`/insights/funnel/correlations` vs `/insights/trend/` vs `/events` vs `/persons/trends`), hard to know what is supported where.
+
+## What?
+
+This is a proposal to change the way we ask for data from the PostHog API. This will unblock the new "data exploration" view, and also:
```
neilkakkar, line 21: I'm curious why you think changing the way we ask for data is the core component of data exploration here?

It seems to suggest that what's hard about this project is "asking for what you want" & how to express that. While useful to have a more uniform query type, I'm not convinced that's where all the complexity lies here.

Rather, it _seems_ to me to be a nice to have (at this stage of the project). I think the complexity lies in trying to combine diverse filters together in a query.

I might be completely off the mark here (or in the end the solution might just become what you propose), but I think this MVP should focus more on validating / figuring out how diverse backend filters can interact together.

requests-for-comments/2022-11-02-data-exploration.md
```diff
@@ -0,0 +1,600 @@
...
+  }
+}
+```
+
+## Proposal: nestable typed filter objects
+
+Instead of a bespoke and scary monolithic filter object, I propose to split this up into **nestable typed filter objects**.
+
+Basically, I'd like to build something that can pass as an AST - Abstract Syntax Tree. This means we could convert it 
+to any other notation down the line (think HogQL perhaps?).
```
timgl, line 330: This is super key. We could end up running on top of other data warehouses like bigquery. Also, one of the long term risks to the business is how we rely on PostHog. If in 5 years time everyone moves over to the next hot thing, we can relatively easily switch.

requests-for-comments/2022-11-02-data-exploration.md
mariusandra: Thanks for asking! This is exactly what the RFC is for: to discuss if we're tackling the right problems, and in the right order. However this is also still "brainstorming mode", where everything's possible and there are no really stupid answers... so treat this more as an "improv comedy" (more "yes and", less "yes but") 😅.

> I'm curious why you think changing the way we ask for data is the core component of data exploration here?

Well that's because if you start with data, then everything else falls into place easier. 😆 

If in this RFC, as the first step, I would have shown just a lot of screenshots of "[universal search](https://github.com/PostHog/posthog/issues/7963)", the mockups from our offsite design sprint, or went into the specifics on how a `breakdown on paths` differs from `breakdown on funnels` at the clickhouse SQL level, we would have gotten lost in detail. I'd like to get the big abstractions out of the way first, *even if we implement them last*. This is a markdown file asking for exactly these comments after all, not 8 man-hours of code :).

The big picture if we don't want to accrue technical debt before the next refactor in 2 years: **the data exploration view should be declarative, composable and embeddable.**

Now let's figure out how all our existing filters clash with this abstraction, and how we can combine existing filters together in a query. I don't have all the answers, but I'll start exploring this next, along with some frontend ideas.

requests-for-comments/2022-11-02-data-exploration.md
timgl: I dug into our queries, and basically all of them follow the same, sort of map-reduce-reduce format:
1. Select all events that are relevant
    - Selects timestamp, aggregation target (person_id, distinct_id) + ‘other’
        - Funnel strict: step_{0,1,…}, latest_{0,1,…} (timestamp), prop_vals (breakdown)
        - Trends: breakdown_value
        - Lifecycle: created_at
        - Paths: path_item
        - Retention: breakdown_values (though this is breadth first)
            - Retention has two: target_event_query and returning_event_query
        - Stickiness: nothing
    - Normally grouped by those selects
2. Reduce to aggregation target, time period + ‘other’
    - Funnel strict: step_(1,2,3)_average_conversion_time_inner, prop (for breakdown)
    - Funnel trends: steps completed
    - Trends: breakdown_value, session_duration
    - Lifecycle: status
    - Path: path_time
    - Stickiness: num_intervals
3. Reduce to counts for the graph

Right now 2 and 3 are often completely intertwined. If we separate them it should be relatively easy to do this. For example:

**Get a list of people at a specific funnel**
The example on line 411 would turn into:
`Select * from persons where person_id IN (select person_id from (FUNNEL OUTPUT STEP 2) where step_1 = 1 and prop = ‘$browser’)`

**Trends query on anyone that dropped off on step 2**
`Select count(1), timestamp from events where person_id IN (select person_id from (FUNNEL OUTPUT STEP 2) where step_1 = 1`

requests-for-comments/2022-11-02-data-exploration.md
```diff
@@ -0,0 +1,600 @@
...
+        ]
+      }
+    ]
+  }
+}
+```
+
+## Proposal: nestable typed filter objects
+
+Instead of a bespoke and scary monolithic filter object, I propose to split this up into **nestable typed filter objects**.
```
timgl, line 327: This is much less work than we think it is, and I don't think we need to start from scratch. I think we can use [these serializers](https://github.com/PostHog/posthog/blob/master/posthog/api/insight_serializers.py#L143) that I created for API documenting, and just create one for each insight type that we're proposing, plus a handful more shapes for the new type of queries.

The added benefit is that we can strongly type this + automatically create an OpenAPI library so querying in the frontend becomes trivial.

requests-for-comments/2022-11-02-data-exploration.md
macobo: > Right now 2 and 3 are often completely intertwined. If we separate them it should be relatively easy to do this. For example:

I might be misunderstanding the proposal but in ClickHouse queries it's often important to combine the two for performance reasons. Otherwise you can end up with badly performing queries due to streaming unneeded data over the network or getting bitten by the lack of query optimizer in clickhouse.

For some examples of this, check out these nifty handbook pages:
- https://posthog.com/handbook/engineering/clickhouse/data-storage#no-query-planner
- https://posthog.com/handbook/engineering/clickhouse/replication#example-query-limit-filter-and-aggregate

pauldambra (commented):
I think you could make a strong argument for doing (part of) this without caring about data exploration. The TS `FilterType` is more than 80 lines long. Only splitting this so there is a separate interface per interaction would clarify the API

Folk are integrated against the existing API so rolling this out (incrementally) will require some redirection/re-documenting/deprecation management. Perhaps the first step is a branch-by-abstraction/strangler pattern approach that puts something in between the existing API and the queries so we have a place to switch the new API on or off to manage them living side-by-side

Another opportunity is the places where the UI makes multiple transformations of the API data. E.g. the funnelLogic has about 5 versions of the API response. So there's the chance to simplify the UI code and use those transformations to guide the shape of the API.

Is this a big enough breaking change that we'd need to v2.0?

requests-for-comments/2022-11-02-data-exploration.md
macobo: > We'd still always execute only one query.

I was assuming that. My point is that a flat query like `select sum(...) from events where time filter group by foobar` is going to perform better than semantically identical SQL in 3 layers like:

```sql
select sum(...)
from (
  select some_cols from (
     select * from events where time filter
  )
)
group by foobar
```

We have competing pressures here: for code simplicity we want the nested variant. For performance, we need bespoke code that generates a more optimal query..

requests-for-comments/2022-11-02-data-exploration.md
fuziontech: `s/PostHog/ClickHouse/g` ?

requests-for-comments/2022-11-02-data-exploration.md
fuziontech: @macobo hit the nail on the head with regards to the crux of the issue here. CH does not have a cost based query planner and so the structure of your query is _critical_. It determines how the query will be executed (unlike most query engines out there) and will be one of the most challenging parts of implementing this in a way that is optimized for performance.

requests-for-comments/2022-11-02-data-exploration.md
```diff
@@ -0,0 +1,600 @@
...
+interface Node {
+  type: string;
+}
+```
+
+Adding a few filters
+
+```json
+{
+  "type": "events",
```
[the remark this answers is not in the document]
Twixes: As far as I understand, those presentation details would still be at the top level, with options depending on `type`

requests-for-comments/2022-11-02-data-exploration.md
```diff
@@ -0,0 +1,600 @@
...
+The "data exploration" view will basically be one big query editor. The first steps will look similar to the existing [MVP](https://github.com/PostHog/posthog/pull/11981) and expose a big table. 
+
+- Two select boxes and a large table below. 
+- You can select "events", apply "matching filters" and see a list of rows
+- You can switch "events" to "persons" and get a different list.
+- Switching to "recordings" might open a different looking interface (recordings might come later)
+
+The goal is to replace the existing events and persons tables, including the persons table in insight modals.
+
+These person modals will now have an "Expand" button that will open them in the "data explorer" view.
```
Twixes, line 515: I feel like already at this stage this waves way a lot of UI complexity with the "Date Explorer" view. This UX is the part that's unclear to me.

requests-for-comments/2022-11-02-data-exploration.md
timgl: All fair points. I think the principle of the separation of queries into map-reduce-reduce still stands even if we do it within the same (sub)query. We can prototype something but I think this'll be possible.

Twixes: This is useful to know. I do agree we should still be able to reuse the structures more, even with flat queries. I'll be hacking on this next week.

mariusandra: If we think of this as an AST, we can solve the problem the same way compilers optimize code: with clever hacks and shortcuts.

For example, taking the "[line 411](https://github.com/PostHog/meta/pull/66/files#diff-2200663423f552a421711aa0cda6877a88fda13dfc78542dd45bf5dfd0a5b4b4R411-R440)" query Tim referred to:

```json
{
  "type": "persons",
  "breakdown": ["person.account_type"],
  "from": {
    "type": "funnel-dropoff",
    "step": "2",
    "from": {
      "type": "funnels",
      "steps": [
        {
          "type": "events"
        },
        {
          "type": "events",
          "filters": {
            "event": "$pageview",
            "properties": [
              {
                "key": "$browser",
                "equals": "Chrome"
              }
            ]
          }
        }
      ]
    }
  }
}
```

All these types, `persons`, `funnel-dropdown`, `funnels`, `events` could evaluate to their own subqueries that can literally be nested within SQL.

However we can run an optimizer as a first pass.

It could replace a tree like `persons -> funnelDropoff -> funnels -> events` with a more optimized `personsFunnelDropOff -> funnelFromEvents` tree if one is available. In this case converting 4 nested sql queries into two. I'm really not sure what's the right level of verbosity here, and I guess that'll only be determined when actually implementing this.

requests-for-comments/2022-11-02-data-exploration.md
mariusandra: Indeed! I was originally considering just `HogQL -> AST -> Clickhouse SQL`, but it can go in any direction. Why not `AST -> BigQuery` or `AST -> InfluxDB`.... or even `AST -> Random PostgresSQL`, finally solving this issue: https://github.com/PostHog/posthog/issues/1244

A lot of the rest of PostHog is still tied to ClickHouse, and I don't think we could offer "bigquery as a backend" any time soon, but we soon start at least considering the option of putting data from other backends onto a dashboard.

requests-for-comments/2022-11-02-data-exploration.md
mariusandra: Will expand on this.

requests-for-comments/2022-11-02-data-exploration.md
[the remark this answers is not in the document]
mariusandra: I've considered two options.

**Option 1**

```json
{ 
  "type": "barTimeGraph", 
  "source": { 
    "type": "events", 
    "filter": { "date_from": "-7d" },
    "breakdown": "event.properties.$browser",
    "interval": "day"
  } 
}
```

In this option, there are 1) data AST nodes and 2) visualization AST nodes. Everything eventually reduces into a visualization node. The "data explorer" frontend will just get `#q={anything}` in the URL, and will render it as asked. If the query reduces into a data node, it'll be wrapped with some default visualization, e.g. `{ type: "dataTable" }`. 

All breakdowns, aggregations, and other filters would still be applied on the data node. There will not be that many data nodes to wrap (just "events" and "persons" probably, and not both by all visualizations), so all rendered visualization nodes should be smart enough to modify parameters inside their data nodes.

The drawback of this approach is that multiple visualizations that share the same data (think trends graph and the table below) will need special handling. Either caching, clever AST optimizations, or other kind of frontend magic to not make two queries. Seems like a solvable problem.

**Option 2**

The other option I could consider is:

```json
{ 
  "type": "events", 
  "filter": {}, 
  "breakdown": "...",
  "display": [
    "table", 
    ["lineChart", {"opt": "value"}], 
    "summaryTable"
  ]
}
```

In this case there's just one type of nodes, but they get a custom `display` property, that's frontend only. It can be modified without triggering a new API request. The frontend would still be smart enough to make an API call that's sure to include the right filters and breakdowns, such as a time interval.

....

I'm leaning towards option 1 at the moment. It feels more composable, and leans nicely into apps and custom components, if we choose to go down that route.

mariusandra:
> I think you could make a strong argument for doing (part of) this without caring about data exploration. The TS `FilterType` is more than 80 lines long. Only splitting this so there is a separate interface per interaction would clarify the API

If we don't go for this RFC, we should refactor this filter anyway.

> Folk are integrated against the existing API so rolling this out (incrementally) will require some redirection/re-documenting/deprecation management. Perhaps the first step is a branch-by-abstraction/strangler pattern approach that puts something in between the existing API and the queries so we have a place to switch the new API on or off to manage them living side-by-side

There could be `{ type: 'legacyInsight', filters: { fill the blank } }`, which the optimizer nicely transforms into a modern set of AST nodes when relevant.

> Another opportunity is the places where the UI makes multiple transformations of the API data. E.g. the funnelLogic has about 5 versions of the API response. So there's the chance to simplify the UI code and use those transformations to guide the shape of the API.

For sure!

> Is this a big enough breaking change that we'd need to v2.0?

Hmm... For the backend, probably not. If all the users get is "explore persons modal further", then probably not. If, however, we reframe this as the new way to do data analysis, then yes. This is a decision we can take after we build the first flagged MVP.

requests-for-comments/2022-11-02-data-exploration.md
EDsCODE: Food for thought:

If we suspend our understanding of the current system and focus on answering the question: how do we enable any sort of behavioral analysis question, I would draw a line between trends and [funnels, retention, lifecycle, paths, stickiness]. Trends, apart from `unique users`, counts events and every other insight counts users [persons, groups]. You're left with a bunch of variations of "funnels" or event sequences. 

Funnels = counting users who have done X events in order
Retention = counting users who have done 2 events in order over recurring period
Stickiness = counting users who have done 2 events in order where the second event is unbounded and you count
Paths = counting users who have coinciding sequence of events
Lifecycle = counting users by how they performed an event at specific times over the entire course of being tracked

Cohort expressions:
Screen Shot 2022-11-03 at 7 53 22 PM

Screen Shot 2022-11-03 at 7 53 42 PM

So, what's been done so far is trying to focus on making the highest value event sequence queries accessible and filterable. If you ask how to make this limitlessly expressible, you would need an input that gives you the ability to specify any number of events in sequence, the period which the events occur, the relation of time between any of the events in sequence, the count of each event in the sequence, and probably some more parameters here.

**The interface here becomes another challenge.

I'm skeptical how reasonable it is to chase a system like this atm. Mainly adding here for context on the far reaches of expressiveness in behavioral queries and it adds more depth to how questions can be asked. The initial idea proposed suggests that "types" are trends, funnels, retention, paths, etc but categorically it's events aggregated in specific ways and person sequences aggregated in specific ways. For example, how would we extend the types to include "users who did an event A stopped for X amount of time and restarted doing by event B"? We'd have to add a different type for every configuration of event sequences that is being asked for

I'll +1 what @neilkakkar mentioned above that the first step is really `I think this MVP should focus more on validating / figuring out how diverse backend filters can interact together, & whether all of them should interact together.` How does a query `property breakdown in a [path step, retention, lifecycle]` even look, does it work, and will it be performant?

macobo:
I'm torn on this. I think this is solving a lot of real problems, since querying data from an frontend or consumer perspective is hard:
1. due to a proliferation of many endpoints
2. out filters system has grown organically with little connective tissue between different areas
3. everything is events-centric, with special cases for special endpoints
4. (implicit) what format is the data coming out in given a query going in?

I think these need to be solved for data exploration to be successful: We kind of need a single endpoint for querying, a structured system that we can transform easily and ways to query data other than events. 

In my opinion, this proposal tackles 1-3 beautifully and doesn't really touch 4.

However where this is shaky is IMO stepping into query construction side of things which I don't think is actually a problem right now beyond the problems mentioned above. 

We already have an expert system in our backend that given inputs can construct (reasonably) performant and correct queries as an output. 

My belief is that restructuring all of it won't yield 2x better outcomes, and it's too early to be building an ad-hoc query optimizer in-house. The existing backend is hard to work with mostly because it's solving a hard problem and requires a lot of domain knowledge - which I think is OK and #team-analytics will gain a grasp on relatively quickly.

I think we can reasonably evolve the existing query building system to help deal with problems 1-4 mentioned above by gradually refactoring to use "new" filter objects rather than gutting everything and starting from scratch.

Aside: I think figuring out what our deprecation strategy for old endpoints/filters should be is also key at this point in time - having to manage two disparate systems will not lead to great outcomes. 

Just my 2c :microphone:

pauldambra:
> Aside: I think figuring out what our deprecation strategy for old endpoints/filters should be is also key at this point in time - having to manage two disparate systems will not lead to great outcomes.

+100

pauldambra:
I think we're missing (maybe we need a parent RFC for the more product-y context)

* how is exploration different from creation
    * is it "people who come to PostHog and don't know what question they want to answer" or ... ?
* how do we find out if "exploration is a page" vs "exploration is an activity across pages"
* can we see people not able to do this in PostHog today or is exploration a new capability (and regardless what's the dashboard we make to measure this)

mariusandra:
Great points everyone. I'll expand the RFC to cover answers to some of the new feedback, add details on the frontend, and include a strategy for deprecation of the old system. 

> I think we can reasonably evolve the existing query building system to help deal with problems 1-4 mentioned above by gradually refactoring to use "new" filter objects rather than gutting everything and starting from scratch.

100%. This absolutely should not be a "let's rewrite everything in javascript" type project, but something where we build a new skeleton, fit the existing queries/insights/etc into it, and branch out as needed.

A rough plan would be something like:

- Refactor `FilterType` into `InsightTrendsFilter`, `InsightFunnelsFilter`, etc to clean up the mess with the huge filter object. It'd be exactly the same filter, but just with the relevant fields (e.g. remove all `funnel*` fields for the trends filter)
- Create an AST system with flat query nodes `insightTrendsQuery`, `insightPathsQuery`, etc. Document them.
- Create a "backend router", that's just one endpoint that takes a query object in the new format, and returns results in the right way. (Q: traceability?)
- Create a "frontend router", that for any AST renders the right components. (Future: lazy loading and apps)
- Create a new query node `events`, and refactor the events table to use this as input. Show the events table when someone requests this at the frontend.
- Refactor the `insightTrends` and `insightFunnels` queries to use these new `events` queries directly as input, instead of the old way to describe steps.
- Refactor the persons modal to use a 2-level query `{ type: 'people', from: { query: { type: 'insightTrends', ... }, day: '2022-02-02', breakdown: 'bla' }`, and use it as the input to person modals nodes.
- Create a `persons` query, with various filters, and use that to power all various persons pages.
- Find a way to make the persons modal use the same system.

At the end of this, we will have done a lot of work, and nothing will have changed for the users (yay). However we could now 1) [embed the live events list](https://github.com/PostHog/posthog/pull/12632) and other data on dashboards, 2) expand the persons modal and see the same data in a list of filters.

We won't have a beautiful AST that 1:1 compiles down to Clickhouse SQL, but we'll have a nested structure that can describe any data, a few new views to explore that data, and a lot of flexibility in how and where we position elements on the frontend. Oh, and our data exploration capabilities will have greatly expanded.

...

I've always said that the beauty about having an abstraction is that then you can always change it. Currently it's a wild west with regards to all the queries, and their inputs and outputs. This is an attempt to clean this up. If we end up with a slightly incorrect abstraction, we can easily change it, and just increase the `version` number in all new queries. Right now we have no idea what we'll break if we change anything, and it's in many ways crippling us. Let's clean this up.

mariusandra:
> I think we're missing (maybe we need a parent RFC for the more product-y context)
> 
> * how is exploration different from creation
>   
>   * is it "people who come to PostHog and don't know what question they want to answer" or ... ?
> * how do we find out if "exploration is a page" vs "exploration is an activity across pages"
> * can we see people not able to do this in PostHog today or is exploration a new capability (and regardless what's the dashboard we make to measure this)

Exploration is a white piece of paper. Some people freeze in paralysis when they are presented with that. Some people start drawing, exploring, erasing, and find what they're looking for. 

So the answer is that it's everything and nothing at the same time. Exploration will be PostHog, so it's hard to pin it to a feature of PostHog.

requests-for-comments/2022-11-02-data-exploration.md
```diff
@@ -0,0 +1,787 @@
...
+
+### Versioning
+
+TODO. We will possibly need an `apiVersion` field in each AST node.
+
+## Next steps for the backend
+
+This is a gradual rollout. There will be no big rewrite. The steps below will also bring no visible changes on the frontend.
+
+1. Refactor the huge `FilterType` into `InsightTrendsFilter`, `InsightFunnelsFilter`, etc. It's still the same backend, just with cleaner queries (e.g. remove all `funnel*` fields for the trends filter).
```
timgl, line 664: Non-blocking, but could we use the serializers in insight_serializers.py for this? That way, we can validate input correctly _and_ we can use the openapi spec to automatically generate the types for the frontend so they're always in sync.

timgl, line 664: Non-blocking, but I think it's worth having some strategy for the migration between these different data models, especially as we save them as filters in dashboarditem which'll make the caching messy.

mariusandra:
Gonna merge this in. Thanks everyone and watch out for PRs solving chunks of this, such as: 

- https://github.com/PostHog/posthog/pull/12745
- https://github.com/PostHog/posthog/pull/12903

PR #68, opened 2022-11-08
RFC: Inserter Service Requirements

fuziontech:
Draft RFC for requirements, reasoning, and phases of development for an Inserter Service that would be used to power, among other things, the new SQL Interface.

### DIFF
--- a/requests-for-comments/2022-11-07-inserter-service.md
+++ b/requests-for-comments/2022-11-07-inserter-service.md
@@ -1,6 +1,6 @@
 # Inserter Service Requirements
 
-We’ve used the `kafka` table engine on ClickHouse for some time now. It was the perfect table engine to support our ingestion pipeline at the time. We had very few resources and needed a way to safely get events from our `[events.py](http://events.py)` ingestion service into ClickHouse with very little fuss.
+We’ve used the `kafka` table engine on ClickHouse for some time now. It was the perfect table engine to support our ingestion pipeline at the time. We had very few resources and needed a way to safely get events from our `[event.py](https://github.com/PostHog/posthog/blob/master/posthog/api/event.py)` ingestion service into ClickHouse with very little fuss.
 
 It’s time now for us to move on from our friend the `kafka` table engine and move this functionality outside of ClickHouse.
 
--- a/requests-for-comments/2022-11-07-inserter-service.md
+++ b/requests-for-comments/2022-11-07-inserter-service.md
@@ -25,6 +25,11 @@ It’s time now for us to move on from our friend the `kafka` table engine and m
 - [ ] Routes rows to the correct ClickHouse instance based on `team_id`
 - [ ] Dynamically load the ClickHouse instance details (hostname, port, credentials) for `team_id`
 - [ ] We improve the guarantees of messages on the data bus for typing and serialization and deserialization efficiency (Protobuf and co)
+- [ ] Needs to be able to be paused via API to support Async Migrations
+- [ ] Extremely granular control over how we route messages depending on the error
+- [ ] Sentry is notified of errors with all of the context
+- [ ] Exponential backoff for `too many parts` errors as well as other categories of errors
+- [ ] Failsafe send messages to DLQ
 
 ## Non Goals
 
--- a/requests-for-comments/2022-11-07-inserter-service.md
+++ b/requests-for-comments/2022-11-07-inserter-service.md
@@ -1,0 +1,76 @@
+# Inserter Service Requirements
+
+We’ve used the `kafka` table engine on ClickHouse for some time now. It was the perfect table engine to support our ingestion pipeline at the time. We had very few resources and needed a way to safely get events from our `[events.py](http://events.py)` ingestion service into ClickHouse with very little fuss.
+
+It’s time now for us to move on from our friend the `kafka` table engine and move this functionality outside of ClickHouse.
+
+## Why?
+
+- [clickhouse.cloud](http://clickhouse.cloud) does not support `kafka` table engines
+- It’s generally well known that the `kafka` table engine is buggy. We’ve seen this a few times with ClickHouse trying to aggressively to retry and pull down Kafka. Altinity and ClickHouse Inc have warned us against using it.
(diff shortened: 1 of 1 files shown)


requests-for-comments/2022-11-07-inserter-service.md
```diff
@@ -0,0 +1,76 @@
...
+  - [ ] Built for ClickHouse, but is abstracted just enough to work with others (BigQuery, Snowflake)
+- [ ] Ensures only rows destined for ClickHouse are inserted ‘only once’
+- [ ] Routes rows to the correct ClickHouse instance based on `team_id`
+- [ ] Dynamically load the ClickHouse instance details (hostname, port, credentials) for `team_id`
+- [ ] We improve the guarantees of messages on the data bus for typing and serialization and deserialization efficiency (Protobuf and co)
+
+## Non Goals
+
+- [ ] We will not materialize columns. We will use new `[JSON` datatype](https://clickhouse.com/docs/en/sql-reference/data-types/json/)
+- [ ] We will not use `Distributed` or `sharded` tables
```
yakkomajuri, line 32: these two things are anyway not part of the ingestion pipeline. Also, ingestion into tables with a sharded setup is the same as ingestion in a non-sharded way. I suspect you're more laying out the requirements for the ClickHouse Cloud setup here rather than the inserter service?

requests-for-comments/2022-11-07-inserter-service.md
```diff
@@ -0,0 +1,76 @@
+# Inserter Service Requirements
+
+We’ve used the `kafka` table engine on ClickHouse for some time now. It was the perfect table engine to support our ingestion pipeline at the time. We had very few resources and needed a way to safely get events from our `[event.py](https://github.com/PostHog/posthog/blob/master/posthog/api/event.py)` ingestion service into ClickHouse with very little fuss.
+
+It’s time now for us to move on from our friend the `kafka` table engine and move this functionality outside of ClickHouse.
+
+## Why?
```
yakkomajuri, line 7: agree with every why here, thanks for the section

yakkomajuri (commented):
Not much to say here - I think this makes sense. The one point from a practical perspective is that we've been thinking about this over at Team Pipeline as well. We currently have no bandwidth to support another service, but likely will in a couple of months with the two new joiners. 

Happy for you to move fast here and just hook onto the topics and build a Cloud-specific service in a way that unblocks the Experimentation team, but this is something we'll probably take over in the medium-term I suspect? Since it pertains to ingestion / pipeline. We might also bring you over to the dark side instead ;)

requests-for-comments/2022-11-07-inserter-service.md
fuziontech: 😎

fuziontech:
>Happy for you to move fast here and just hook onto the topics and build a Cloud-specific service in a way that unblocks the Experimentation team, but this is something we'll probably take over in the medium-term I suspect? Since it pertains to ingestion / pipeline. We might also bring you over to the dark side instead ;)

I appreciate your ambition to expand the scope of your team by increasing the number of services that it is responsible for. I don't believe a team should simply annex a project because the name of the team warrants it. I would point to [Conway's law](https://en.wikipedia.org/wiki/Conway%27s_law) here and suggest that having this not be owned by the `ingestion / pipeline` team might be a good idea.

**Any organization that designs a system (defined broadly) will produce a design whose structure is a copy of the organization's communication structure.**

HBS and MIT have also found evidence to support this:
**product developed by the loosely-coupled organization is significantly more modular than the product from the tightly-coupled organization**

So in the end I would suggest we should aim to reduce the scope of Ingestion / Pipelines into smaller pieces (events, plugins, schedule/async tasks, etc). This would help us develop much more clear interfaces between each in the inevitable event we want to untangle the bowl of noodles and replace components in place.

timgl:
> I don't believe a team should simply annex a project because the name of the team warrants it. I would point to [Conway's law](https://en.wikipedia.org/wiki/Conway%27s_law) here and suggest that having this not be owned by the ingestion / pipeline team might be a good idea.

Which other team would own this?

yakkomajuri:
Ambition to expand scope? lol 

Please take all my responsibilities away - I'm just wondering who will realistically own this ingestion service if not the ingestion team

requests-for-comments/2022-11-07-inserter-service.md
```diff
@@ -0,0 +1,76 @@
...
+- It’s generally well known that the `kafka` table engine is buggy. We’ve seen this a few times with ClickHouse trying to aggressively to retry and pull down Kafka. Altinity and ClickHouse Inc have warned us against using it.
+- We’d like to decouple our pipeline from ClickHouse.
+- We’d like to be able to insert rows into multiple ClickHouse warehouses (by team) without an explosion of topics.
+- We’d like to support dynamic and parametrizable warehouse connection parameters
+  - Team 2 → Warehouse with connection details x.y.z pulled from API
+- Use whatever serialization we want on ClickHouse without having to ship the schema as a file to ClickHouse and reboot it (Proto and Avro problems)
+- Much more configurable error handling.
+- We can instrument much better metrics on the health of this step in our pipeline.
+
+# Goals
```
macobo, line 19: This consumer should be pausable within (async) migrations. Otherwise a whole category of options we have for migrations becomes unavailable.

requests-for-comments/2022-11-07-inserter-service.md
```diff
@@ -0,0 +1,76 @@
...
+
```
macobo, line 20: Nothing here about error handling which I consider key for this to be successful.

Suggestions:
- it routes all errors to sentry
- exponential backoff for too many parts errors
- unsure what to do about other errors - dlq?
- does not commit any offsets unless everything in a batch passes

macobo (commented):
I think there's a real cost to this project:
- makes self-hosting more costly
- for months we'll have more errors and work due to this service rather than less
- little immediate payoff for pipeline robustness - we could be handling typing errors better in plugin-server

fuziontech:
We will probably need to build this some day. All we know is that day is not today.

PR #71, opened 2022-11-10
RFC: Software Engineering Language Policy at Posthog

fuziontech:
We've talked around this one for a while. I know some people here are polyglots and really love working with and being exposed to a few different languages, others are extremely dedicated to their languages.

The important thing to underscore here is that it is about 10% proposal to use Golang and about 90% honest desire to get people's opinions.

This is a rare chance for us to do Issues over PRs...otherwise I would have just put up a huge PR entirely in Golang and said 🚢 it.

### DIFF
--- a/requests-for-comments/2022-11-08-supported-languages.md
+++ b/requests-for-comments/2022-11-08-supported-languages.md
@@ -1,0 +1,156 @@
+# Request for comments: Supported languages at PostHog
+
+Since day 1 here at Posthog we've supported two and half core languages: Python and Javascript/Typescript.
+
+This set of tools has carried us a long way and we should both the cpython runtime and node a huge hats off for bringing us this far.
+
+## Problem statement
+
+We have the opportunity coming up to rebuild or greenfield build out services that are critical parts of our data pipeline. It will be important for these services to be correct, fast, and efficient. Considering this now is a good time to ask: Are we using the correct tools for the job?
+
+So why even bring this up?
+
+Frankly:
+
+- Python is slow.
+- Node is a memory hog.
+- Node tooling is slow.
+- Dependencies can be huge.
+- No guarantees that code is correct.
+
+Can we be successful continuing to use these languages? Of course.
+
+- They are well understood.
+- Our tooling is setup to support these languages
+- We have internal expertise.
+- Generally they are _good enough_.
+
+Why should we be open to other languages?
+
+- We have different people with experience working with different languages.
+- 100% guaranteed static typing is quite beneficial
+- Compiled languages provide more confidence shipped code is correct
+- There are compiled languages that provide significant efficiency wins on CPU, Memory, and performance
+- Using the right tool for the job is typically the right thing to do (if you can ship it)
+
+Candidate Languages:
+
+- [Golang](https://go.dev/) (big surprise)
+- [Rust](https://www.rust-lang.org/) 🦀
+- [OCaml](https://ocaml.org/)
+- [Elixir](https://elixir-lang.org/) (dynamically typed)
+- [Scala](https://www.scala-lang.org/)
+- [Java](<https://en.wikipedia.org/wiki/Java_(programming_language)>)
+
+## Meet the eligible languages
+
+### Golang
+
+The good:
+
+- Extremely easy to learn. Most engineers can learn and be productive in about a day.
+- Built to remove most contentious parts of development
+- Designed to make engineering easier
+- Built for network services and concurrency
+- Great standard lib
+- Super fast compile times
+- Light on memory
+- Used by plenty of organizations big and small
+- We have plenty of Gophers here.
+
+The bad:
+
+- Most people would not say the language sparks joy (I disagree)
+- Not terribly expressive
+- Verbose, but easy to read
+
+### Rust
+
+The good:
+
+- Very hot right now
+- Very good for when you need to hyper optimize something but would like to avoid C or C++
+- Extraordinarily expressive and fun to program
+- Extremely performant and light on memory
+- Used by plenty of organizations big and small
+- We have a few Crustaceans here!
+
+The bad:
+
+- Slow compile times
+- Harder to read because of how expressive
+- Slower ramp up time to become proficient in as an Engineer
+
+### OCaml
+
+> OCaml is a general-purpose, industrial-strength programming language with an emphasis on expressiveness and safety.
+
+You may not be familiar with this one but it is _very_ safe and performant. Companies that have huge amounts of money on the line who are risk adverse to defects and slowdowns, like the HFT firm Jane Street use OCaml because it is unforgiving in how type safe it is.
+
+> OCaml’s powerful type system means more bugs are caught at compile time, and large, complex codebases are easier to maintain. This makes it a good language for running critical code. At the same time, sophisticated inference makes the type system unobtrusive, creating a smooth developer experience.
+
+The good:
+
+- Garbage collected
+- Algebraic data types
+- Pattern matching
+- Type inference
+- Immutable
+- Static Type-checking
+- First-class functions
+- Parametric polymorphism
+- Used by serious programming shops
+
+The bad:
+
+- Not terribly popular relative to the others
+- Can be considered relatively academic
+- No one experienced here
+
+### Elixir
+
+Oh, Erlang <3
+
+The good:
+
+- Built for streaming data
+- The cockroach of runtimes.
+- Automatic function level clustering
+- Hot reloading of functions
+- Compiled
+- Relatively functional
+
+The bad:
+
+- Dynamically typed
+
+### Scala / Java
+
+Grouping these together because they really are converging. Really this is any language on the JVM.
+
+The JVM is truly a wonderful runtime. It's very fast. The hotspot detection + Just In Time Compiling is _super_ impressive.
+
+The good:
+
+- Surprisingly performant
+- JVM Library interrop means huge availability of libraries out there
+- Tons of expertise and companies big and small are building software in this
+- Kafka and Zookeeper are built on this
+
+The bad:
+
+- Considered somewhat crusty
+- Slow boot times
+- JVM tuning is no fun
+
+## Discussion point
+
+So with this overview laid out let's talk about
+
+## Success criteria
+
+_How do we know if this is successful (i.e. metrics, customer feedback), what's out of scope, whats makes this ambitious?_
+
+The goal here is to spark debate about languages that we should use here at Posthog. What are we open to? Why should we not adopt new technologies.
+
+Success here would be to make a decision and effectively enact a policy on this and have engineers aligned and not worry about this again (at least for some time)
--- a/requests-for-comments/2022-11-08-supported-languages.md
+++ b/requests-for-comments/2022-11-08-supported-languages.md
@@ -57,6 +57,7 @@ The good:
 - Light on memory
 - Used by plenty of organizations big and small
 - We have plenty of Gophers here.
+- A simple binary as a deliverable.
 
 The bad:
 
@@ -74,6 +75,7 @@ The good:
 - Extremely performant and light on memory
 - Used by plenty of organizations big and small
 - We have a few Crustaceans here!
+- A simple binary as a deliverable.
 
 The bad:
 
@@ -123,6 +125,7 @@ The good:
 The bad:
 
 - Dynamically typed
+- Runs in a VM
 
 ### Scala / Java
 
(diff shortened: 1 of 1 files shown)


requests-for-comments/2022-11-08-supported-languages.md
```diff
@@ -0,0 +1,169 @@
...
+
+So why even bring this up?
+
+Frankly:
+
+- Python is slow.
+- Node is a memory hog.
+- Node tooling is slow.
+- Dependencies can be huge.
+- No guarantees that code is correct.
```
mariusandra, line 19: > - Python is slow.

True generally, but relevant only for specific usecases (like here).

> - Node is a memory hog.

Has this been a problem? There are alternative implementations of JS (e.g. [bun](https://bun.sh/), [deno](https://deno.land/)) that do this better, but are too fresh to rely on in production.

> - Node tooling is slow.

Depends on the tooling. The frontend builds quite fast (sub 3-10sec), the plugin server can be improved 10x if needed.

> - Dependencies can be huge.

We're [switching to `pnpm`](https://github.com/PostHog/posthog/pull/12635) that'll fix this

> - No guarantees that code is correct.

Can you ellaborate?

mariusandra (changes requested):
I'm not in any of the teams that'll be maintaining this going forward. Their buy in is more important than mine. That said, I think this is a distraction. First, the RFC reads like it's written to find a problem whose answer is "go". Second, the "frankly" part of the problem statement was frankly weak and unconvincing. 🤷

Let's not. 🤣

requests-for-comments/2022-11-08-supported-languages.md
```diff
@@ -0,0 +1,169 @@
+# Request for comments: Supported languages at PostHog
+
+Since day 1 here at Posthog we've supported two and half core languages: Python and Javascript/Typescript.
+
+This set of tools has carried us a long way and we should both the cpython runtime and node a huge hats off for bringing us this far.
+
+## Problem statement
+
+We have the opportunity coming up to rebuild or greenfield build out services that are critical parts of our data pipeline. It will be important for these services to be correct, fast, and efficient. Considering this now is a good time to ask: Are we using the correct tools for the job?
```
pauldambra, line 9: I'd suggest some of the benefits of switching language comes from re-writing/extracting the service and only part from re-writing/extracting the service _in a different language

At a previous job we spent a while rewriting from Python to C# because "C# is faster than Python". Then realised that we were CPU bound and moved a bunch of work onto the GPU. Python and C# were then faster than we needed them to be.

So: "are we using the correct tools for the job" is a great question. But being really clear on why we're asking that and how we'll know when we're done is super important. E.g. moving ingestion onto the GPU is probably more complicated and not actually faster.

So, there are two interesting questions here:

1) what needs to be done to make ingestion safer/faster/misc
2) how many programming languages can we support

These are really different questions. 

I'm not working on ingestion so I can have opinions on 1 but "so what?"

On "how many programming languages"... The problem with a new (to you) programming language is almost never the language.

Aside from https://neverworkintheory.org/2014/01/29/stefik-siebert-syntax.html Python and Ruby are consistently measured as easiest to learn. C-style languages are no easier to learn than a language made of random keywords.

The problem with a new (to you) language is generally the tooling. Most noticeably, in my experience: dependency management, and building and releasing things.

(go update our android library from java 8 to java 19 if you want proof of this :))

So, we need a comparison of building, releasing, and running in k8s for the languages we consider. I think we can consider a smaller list (although C# marketing should be "Java but good", and I've a friend we could hire if we started using clojure which seems to inspire massive love from folk that use it)

The other thing is adoption within the org. Who needs to learn the language, and how and when do they do that? How do we know when to write it? Are we migrating to it or adding it alongside? Which services mustn't it be used for?

And finally: hiring. "Come work here because you're excited about language X. Incidentally the first few months you'll be working on these bugs in Python". So, we can pull from a wider pool with a wider pool of languages but are we in a place to hire someone who _only_ wants to work on language X

----

Bonus "after finally" point... who *owns* tooling for each language? Does the platform team commit to providing support for building, deploying, and running all the languages? Is that in our definition of platform? Or do we need some champions

requests-for-comments/2022-11-08-supported-languages.md
```diff
@@ -0,0 +1,169 @@
...
+The bad:
+
+- Dynamically typed
+- Runs in a VM
+
+### Scala / Java
+
+Grouping these together because they really are converging. Really this is any language on the JVM.
+
+The JVM is truly a wonderful runtime. It's very fast. The hotspot detection + Just In Time Compiling is _super_ impressive.
```
pauldambra, line 134: I saw a talk by the folk at a German car sales company who were shipping an ML model in Java. They had to write tooling to hit all of the code paths they cared about during boot because JIT was a problem for them.

That said there are a bunch of alternative VMs that are focussed on speed because of serverless-style loads...

requests-for-comments/2022-11-08-supported-languages.md
```diff
@@ -0,0 +1,169 @@
...
+
+It's a simple service and could be written any anything really. It does give us an opportunity to branch off our well traveled path of Python and Typescript. In my opinion I think Golang or Rust would be a great fit here, as would the other languages listed. I'm a Gopher in particular so I would really like to see more written here.
+
+## Success criteria
+
+_How do we know if this is successful (i.e. metrics, customer feedback), what's out of scope, whats makes this ambitious?_
+
+The goal here is to spark debate about languages that we should use here at Posthog. What are we open to? Why should we not adopt new technologies.
+
+Success here would be to make a decision and effectively enact a policy on this and have engineers aligned and not worry about this again (at least for some time)
```
pauldambra, line 169: As above... one decision only?

1) what needs to be done to make ingestion safer/faster/misc
2) how many programming languages can we support
3) how do we support any one programming language

charlescook-ph:
Obviously hiring difficulty shouldn't be a deciding factor here (and I am 100% _not_ suggesting that), but if we don't have a strong feeling overall for Python vs. Go or Rust here, it's worth noting that there are [~5x Python devs](https://www.statista.com/statistics/793628/worldwide-developer-survey-most-used-languages/) out there vs. either of those languages. 

(Again, any decision should be what's best for the company and product, not what is easier to hire for - I'm just calling this out if there doesn't end up being strong consensus one way or the other.)

requests-for-comments/2022-11-08-supported-languages.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Problem statement
+
+We have the opportunity coming up to rebuild or greenfield build out services that are critical parts of our data pipeline. It will be important for these services to be correct, fast, and efficient. Considering this now is a good time to ask: Are we using the correct tools for the job?
+
+So why even bring this up?
+
+Frankly:
+
+- Python is slow.
```
guidoiaquinti, line 15: Can you please expand on this and mention at least 2 use cases from our codebase where Python runtime was the bottleneck?

requests-for-comments/2022-11-08-supported-languages.md
```diff
@@ -0,0 +1,169 @@
...
+- Node is a memory hog.
```
guidoiaquinti, line 16: Can you please expand on this and mention at least 2 use cases from our codebase where Node memory management was the bottleneck? (note: **Node memory management**, not how we write Node).

requests-for-comments/2022-11-08-supported-languages.md
guidoiaquinti, line 19: What does this mean? The code is not deterministic because of the language?

requests-for-comments/2022-11-08-supported-languages.md
```diff
@@ -0,0 +1,169 @@
...
+- [Scala](https://www.scala-lang.org/)
+- [Java](<https://en.wikipedia.org/wiki/Java_(programming_language)>)
+
+## Meet the eligible languages
+
+### Golang
+
+The good:
+
+- Extremely easy to learn. Most engineers can learn and be productive in about a day.
```
guidoiaquinti, line 51: I'm not part of the _most engineers_ group so

guidoiaquinti:
I don’t want to sound too direct, but I don’t think this RFC solves any of the problems we currently have in developing software. Looks like it was written with a solution in mind rather than analyzing and addressing problems we are facing day to day.

Here are the issues this proposal claims to fix:

* Python is slow: are we even measuring this? Is the Python runtime slow or is our code? In any case, did we make any effort in the last 6 months to make it faster?  

* Node is a memory hog: same as above.

* Node tooling is slow: not an expert here, so I can’t judge, but changing a language because some tooling is slow doesn’t seem to be a smart business decision.

* Dependencies can be huge: same as above (plus: is this a problem? Are we running on embedded hardware and we need to deal with super strict limits)? 

* No guarantees that code is correct: as commented earlier, I don’t get what this means.

General Feedback: as part of an analytics company, I invite our entire team to start making bullet-proof proposals and decisions using irrefutable metrics and data.

_A_ is a problem: 
  - Why? As can be seen in metric _XYZ_ 
  - Impact? We are wasting _X_ resources every _Y_ days as can be seen in metric _XYZ_

Proposal A:
  - PRO
  - CONS 
  
Proposal B:
....

Twixes:
I can see how some of these languages might be good choices for a greenfield project (especially Rust and Go in my view), but I'm curious what such greenfield project are we planning? It doesn't seem like it would benefit us to _rewrite_ anything.

requests-for-comments/2022-11-08-supported-languages.md
```diff
@@ -0,0 +1,169 @@
...
+- [Golang](https://go.dev/) (big surprise)
+- [Rust](https://www.rust-lang.org/) 🦀
+- [OCaml](https://ocaml.org/)
+- [Elixir](https://elixir-lang.org/) (dynamically typed)
+- [Scala](https://www.scala-lang.org/)
+- [Java](<https://en.wikipedia.org/wiki/Java_(programming_language)>)
+
+## Meet the eligible languages
+
+### Golang
```
ellie, line 47: Tbh I would not be against us introducing Go for a new service, so long as we're sure that Python does not and cannot suit our use case

requests-for-comments/2022-11-08-supported-languages.md
```diff
@@ -0,0 +1,169 @@
...
+- We have plenty of Gophers here.
+- A simple binary as a deliverable.
+
+The bad:
+
+- Most people would not say the language sparks joy (I disagree)
+- Not terribly expressive
+- Verbose, but easy to read
+
+### Rust
```
ellie, line 68: I _think_ I would not be wrong in saying I have the most Rust experience here. I LOVE Rust. and I'd love to write it at work. I also have several friends who are fantastic engineers + would consider applying here purely because of Rust. 

But tbh I'd be **super** against us adopting it. The learning curve is huge (prepare to not be comparatively productive for around a quarter). Unless it turns out there's actually more people writing Rust for fun here than I though 😄 

While maturing, the ecosystem is nothing like Python. We'd be spending a lot more time in the weeds. There are fewer "obvious" choices than other languages. EG, which async runtime should we use (if any...)? Why? Do we know enough about how our application does/should work in production to even start making such a choice?

Generally organizations choose Rust because

1. They have explicitly decided that they will be a "Rust company", from the beginning (eg see TrueLayer)
2. They have eked out every last drop of performance from their current stack, and need to go yet further (eg see Discord)

ellie (commented):
I have a bunch of thoughts here, and I think the outcome of this depends on what we're trying to achieve

1. are we suggesting a paved road or framework for all future services, which requires buy-in from everyone and continued maintenance into the future? Ie, are we wanting to eventually replace python and node, and consider them "old" technologies in our stack?

2. do we want to use something that isn't Python for the [inserter service](https://github.com/PostHog/meta/pull/68)?

I think if this is the case, instead of arguing that technology X is better than technology Y in general, we should instead argue why introducing another technology is worth it for that specific use case.

From my understanding of the inserter service, it would be well implemented in Go. But we should definitely ensure that we explore Python as a possibility and conclude as to why it _isn't_ a good idea, given Go would add a little friction elsewhere.

-----

Otherwise. Going forwards I'd love us to be more data-driven with our decisions like this. It's almost ironic that we aren't, given our product enables companies to make data-driven decisions about their own products. Before we make sweeping decisions about a language/tech, we definitely need to measure things in much more detail and know why what we have right now does not work well enough for us.

While I love playing with new tech, it's also worth giving this essay a read: https://mcfunley.com/choose-boring-technology

fuziontech:
lol kill it with fire

PR #74, opened 2022-11-25
RFC: Resilient feature flags

neilkakkar:
~~Draft for now, need to flesh this out further, just jotting down core points~~

### DIFF
--- a/requests-for-comments/resilient-feature-flags
+++ b/requests-for-comments/resilient-feature-flags
@@ -39,7 +39,9 @@ On the SDK side:
 
 To make this possible, we cache flag definitions (and not the decide responses) on our side, so these are available even if the DB is down.
 
+It makes sense to include (3) with (1) for decide because the shape of the response is changing anyways, and it's much easier to handle both at the same time in all client libraries, than doing these one by one.
 
+We don't have to build the rest of it out right now, but having this functionality here means when do build it, it will be extraordinarily faster to do.
 
 
 ## Context
--- a/requests-for-comments/resilient-feature-flags
+++ b/requests-for-comments/resilient-feature-flags
@@ -1,0 +1,52 @@
+# Request for comments: Resilient feature flags
+
+## Problem statement
+*Who are we building for, what are their needs (include example cases), why is this important?*
+
+We've recently realised a few limitations with feature flags when PostHog goes down. This isn't great for permanent feature flags as it causes things to break
+in unexpected ways & for clients to think really hard about defaults & write code defensively.
+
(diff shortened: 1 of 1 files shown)


requests-for-comments/resilient-feature-flags
```diff
@@ -0,0 +1,52 @@
...
+
+On the SDK side:
+
+1. We stop obliterating all flags if decide returns a 500, and based on the parameter in (2), update only flags that were computed again, while keeping old values of other flags.
+2. We add timeouts to flushes & shutdowns
+
+
+To make this possible, we cache flag definitions (and not the decide responses) on our side, so these are available even if the DB is down.
+
+
```
neilkakkar, line 42:
```suggestion
It makes sense to include (3) with (1) for decide because the shape of the response is changing anyways, and it's much easier to handle both at the same time in all client libraries, than doing these one by one.

We don't have to build the rest of it out right now, but having this functionality here means when do build it, it will be extraordinarily faster to do.
```

requests-for-comments/resilient-feature-flags
```diff
@@ -0,0 +1,81 @@
...
+Caching decide responses doesn't make too much sense, since it's called usually only once per distinct ID per session. Every anonymous user has a new ID, and except
+for users coming back to the app again & again on the same day, we don't achieve resiliency as we can't cache for users never seen before, nor can we assume that properties
+haven't changed; while introducing new problems like cache invalidation.
+
+### Possible failure modes
+
+1. The database is down/unreachable: `decide` evaluation fails and returns 500
+2. The servers are down/unreachable: requests from client libraries time out / error out.
+
+(2) seems hard to defend against without a distributed app distribution (and then a resolver to go to the 'correct' app), but (1) is possible.
```
guidoiaquinti, line 39: what is a _distributed app distribution_?

neilkakkar: Talk about terrible wording 😅 . I basically mean multiple server deployments where one going down doesn't affect the other, like an edge server; and a load balancer than can appropriately link to healthy closest server.

guidoiaquinti: I don’t think we should build an SDK making an assumption on the underlying infrastructure. 

To build resilient software, use defensive programming: always hope for the best (uptime 100%) but always prepare for the worst (everything is down). AKA: a multi-geo deployment might help you, but you should not rely on it.

neilkakkar: I'm not assuming that at all, I'm just listing out what are ways things can go wrong here.

Even with defensive programming, the issue still remains that when servers go down, you stop getting flag information.

(a point below addresses this)

PR #76, opened 2022-11-29
RFC Get from web performance v0.1 to v0.2 by adding network logs to session recordings

pauldambra:


lharries:
@pauldambra is this related to the work being done by product analytics to improve the performance of our own product? And any extra info on the motivation for the "Why we should do this?" would be great

macobo:
Gonna critique this _only_ from a technical standpoint.

**TL;DR: Schema and ingestion won't work for this.**

The problem with this approach is that it's taking a short-cut by trying to store web performance data in main `events` table. 

This has advantages:
- Easier to leverage existing ingestion system
- Easier to build queries

But a significant disadvantage:
- This system will make _all_ insights across posthog slower (even for users without web performance turned on)
- It increases the amount of effort we need to put into building and maintaining the ingestion system - probably the area that's hardest to scale.
- The resulting product will perform badly

Why? Similar to why session recordings is in a bad place, it all comes down to the schema. 

By dumping it all into `events` table we may have a schema that's able to answer _all_ queries, but also equally sucks at all of them. It also needs to go through the complex ingestion pipeline - even though it might not be required for this data. Lastly, computers are _much much_ better than humans at generating data - if anyone significant starts using this feature.

This isn't a thing we can "fix in post" - we made a similar trade-off with session recordings in 2020 and have repeatedly failed to get that fixed. The trade-off made sense in 2020 as we were just starting out and it helped get traction - however it might seriously hurt the company in 2022/2023.

How to do this right? Create a _specialized_ ingestion system and schema. We have done this successfully already with app metrics - for every event we track N times the amount of "app metric" events. Yet the resulting queries are lighting fast and we store _megabytes_ instead of _terabytes_ for the same amount of data in events table. It's literally millions of times better!

Also - please, _please_ read the clickhouse manual! All of these topics are covered in there in more detail.

pauldambra:
@macobo agreed events table is the wrong place for this (as much as it makes querying easy)

In the RFC I put

```
Since this will send tens or hundreds of events per pageview we could receive a massive amount of traffic if sending each request individually. 

So, similarly to how we process session recordings, it makes sense to send info from the SDK in batches `{sessionId: string, pageViewId: string, entries: PerformanceEntry[]}` and then split them on ingestion. 

The individual performance entries are relatively small but will be numerous. Storing them separately to events would allow TTL deletion as we do with session recordings.
```

I don't understand the bones of the implementation of that but I don't mean for us to store this in `events`

pauldambra:
> this approach is that it's taking a short-cut by trying to store web performance data in main events table.

The existing alpha does that but I'm not proposing that as the way forwards

macobo:
> Since this will send tens or hundreds of events per pageview we could receive a massive amount of traffic if sending each request individually. 
> So, similarly to how we process session recordings, it makes sense to send info from the SDK in batches `{sessionId: string, pageViewId: string, entries: PerformanceEntry[]}` and then split them on ingestion. 
> The individual performance entries are relatively small but will be numerous. Storing them separately to events would allow TTL deletion as we do with session recordings.

This is getting two different ingestions confused: one is library-level, one is our plugin-server - I think our confusion is coming from that.

> The existing alpha does that but I'm not proposing that as the way forwards

I think scoping out the schema is the most important piece of work we can do here as it forces us to also figure out performance, product, ingestion and storage. Note TTLs themselves are not an answer - similar to app metrics or how a "proper session recordings schema" looks like, we might want to leverage aggregation at a schema level or do other things to make this cheap to store/query.

benjackwhite (commented):
Love that we're tackling this with our brains and words first because as Karl hints - we need to scope this right.

I think it is fair to say that this is a highly requested feature for Session Recordings alone so if all we did was make it work for that (we could do this naively as we do console logs) that would already be super valuable. If however we can come up with a more advanced / thought through approach that would allow us to bring this data into insights. 

I bow to Karl's knowledge here about the data but it feels like we need to map out what axis would we need to query on / aggregate on and understand that as well as how to show the full detailed view (initially for SR). Could be that we purposefully have multiple storage approaches for the different problem areas?

pauldambra:
@annikaschmid 👋

@lharries mentioned you've been gathering context on network logs in session recordings. Does it make sense to add the context here?

annikaschmid:
@pauldambra I don't have in-depth feedback on Network Logs for Session Recordings. The feedback I got around this can be summarised as:

- "If we see a session where a user has a bug, or they contact us with an issue, we do check the console logs in the session recordings to see what the issue is" (I expect them to use network logs in a similar way)
- "It would be *awesome* if we would have network logs as well to get the full picture"
- "This is what would make us fully move from LogRocket to PostHog"

The particular 'pain point' here is that users are used to this feature from another tool, e.g. LogRocket, and would therefore expect this in PostHog as well, especially for debugging. I can see you already did a short competitor review, so if we introduce some of their main functionality, that is a great start and we can gather some additional feedback around what's missing.

pauldambra:
We've started releasing this... PRs in posthog-js and posthog repo all tagged `lisbon-web-performance`

PR #77, opened 2022-12-01
The first "Workflow": MVP for "Creating a new feature"

lharries:
Owner: @EDsCODE (team experimentation)
Decision to be made: Should we build this in Q1 as part of team experimentation?
Decision date: 7th December 2022

@annikaschmid could you work with the team to add some more detail?

### DIFF
--- a/requests-for-comments/2022-12-01-workflow-new-feature.md
+++ b/requests-for-comments/2022-12-01-workflow-new-feature.md
@@ -24,10 +24,13 @@ So far we have given product engineers the individual tools to create trends, cr
 
 We have the opportunity to connect these flexible tools with opinionated best practices to focus on specific jobs-to-be-dones in PostHog. These could be called "Workflows".
 
-Two examples of potential workflows:
+Some examples of potential workflows:
 
-1. Building a new feature - assisting with rollout and measuring success *More details below*
+1. Creating and shipping a new feature - assisting with rollout and measuring success *More details below*
 2. Workbooks - making it easy to come up with questions, save related insights, clip particular bits of recordings and jot down thoughts
+3. Iterating on existing features - making sure the new feature is better than the previous iteration
+4. Conversion experiment - making sure that the funnel metrics improve (including quality of new customers)
+5. Fixing bugs - understanding what's broken and confirming that the fix works
 
 ## Vision for "Creating a new feature" workflow
 


requests-for-comments/2022-12-01-workflow-new-feature.md
```diff
@@ -0,0 +1,57 @@
...
+
+Sarah clicks "Done". PostHog automatically creates a feature flag to control the rollout; sets up a dashboard containing session recordings of people using it; and creates the relevant graphs.
+
+Sarah builds the feature between the feature flag and merges it to main branch.
+
+Sarah pings her team to give it a go on a few different devices. When 5 internal users have used the feature PostHog sends her a slack message to the playlist. She reviews it and the error logs - it's all working fine so she gives the thumbs up to escalate to beta users. There are no errors in beta and so it automatically deploys to prod - sending her a slack message to her team to keep them in the loop. PostHog automatically invites 2 users to a user interview after having used the video feature *Luke is working on an MVP for the user interviews feature - can be out of scope for the MVP*.
```
annikaschmid, line 44:
```suggestion
Sarah pings her team to give it a go on a few different devices. When 5 internal users have used the feature PostHog sends her a slack message to the playlist. She reviews it and the error logs - it's all working fine so she gives the thumbs up to escalate to beta users. There are no errors in beta and so it automatically deploys to prod as a phased release - sending her a slack message to her team to keep them in the loop. PostHog automatically invites 2 users to a user interview after having used the video feature *Luke is working on an MVP for the user interviews feature - can be out of scope for the MVP*.
```

annikaschmid:
Thanks for writing this up, @lharries! I think for the purpose of communicating what the feature is about, the vision is great.
There are a couple of things we need to think through a bit further once we start with the feature, e.g.:
- I think you wouldn’t necessarily have to choose between auto-rollout and a/b test. You could start with an a/b test and roll out the feature if the a/b test was successful
- Retention is a bad metric to pick for a feature test, because it’s a lagging indicator. In this case, I don’t think it would lead to good results
- In this case, Sarah made changes to the feature after it has been rolled out. Would she create a new workflow for this? Probably not, but maybe she still wants to roll out changes with a schedule. I suppose the user can decide on a case by case basis whether a change requires a new flag or not (like today)

lharries:
@annikaschmid another extension that we might want to think about (for the longer term rather than this quarter) - you mentioned the idea of "Feature management", what if Feature Management was a 1st class feature in PostHog? What would this look like?

requests-for-comments/2022-12-01-workflow-new-feature.md
```diff
@@ -0,0 +1,57 @@
...
+2. Workbooks - making it easy to come up with questions, save related insights, clip particular bits of recordings and jot down thoughts
+
+## Vision for "Creating a new feature" workflow
+
+Here's the vision of what this could be in 3 - 6 months:
+
+Sarah is a product engineer at a high-growth startup building a B2C website to help companies offset their carbon - called Carbon.io. She's working on a new feature for users to watch videos of their carbon-offsetting projects. She hypothesizes that users who watch these videos will feel more connected to the projects and retain at a higher rate.
+
+The first thing Sarah does is open up PostHog and start the workflow for "Create a new feature". It opens up a page to create the new feature where she names it "Videos of offsetting projects". Optionally, she can enter some of the background information on this feature including the hypothesis. She selects "auto-rollout" rather than "A/B test" - meaning once the feature is made it will be used internally and after manual approval will be rolled out to beta users then low-value users then high-value users. She selects "user retention" as the key metric.
```
EDsCODE, line 38: Do we think this feature will be applicable for existing users or are we hoping to attract a new set of users? If existing, is there evidence that current users would be interested in working this way. For example, are event taxonomies and feature flags decided upon pre feature implementation or post?

If not existing, is there some research we can do to get information on how this plays against the existing popular workflows among product engineers? I would assume there's some level of tooling or assortment of tools that are used to achieve this product planning -> implementation

requests-for-comments/2022-12-01-workflow-new-feature.md
```diff
@@ -0,0 +1,57 @@
...
+Decision to be made: Should we build this in Q1 as part of team experimentation?
+Decision date: 7th December 2022
+
+## Context
+
+This RFC proposes that team experimentation builds the MVP of a workflow for "Creating a new feature" as part of their Q1 OKRs.
+
+It should sit with team experimentation as it would be a key driver of feature flag usage.
+
+This could have multiple benefits across the product:
```
EDsCODE, line 13: The promise for an opinionated workflow makes a lot of sense. If we can properly automate the pre and post implementation for engineers this would be a significant value add.

**I will asterisk that adding opinionated features like this could get confusing from a marketing standpoint. Most of the product is fairly unopinionated "here's analytics tooling you can use to understand your users" while this feature moves in the direction "here's a product _process_"

requests-for-comments/2022-12-01-workflow-new-feature.md
```diff
@@ -0,0 +1,57 @@
...
+
+Sarah pings her team to give it a go on a few different devices. When 5 internal users have used the feature PostHog sends her a slack message to the playlist. She reviews it and the error logs - it's all working fine so she gives the thumbs up to escalate to beta users. There are no errors in beta and so it automatically deploys to prod as a phased release - sending her a slack message to her team to keep them in the loop. PostHog automatically invites 2 users to a user interview after having used the video feature *Luke is working on an MVP for the user interviews feature - can be out of scope for the MVP*.
+
+From the dashboard, Sarah learns most of the users are only watching the first 30 seconds of the 5-minute videos. During the interviews, she asks them what they thought of the videos and while they enjoyed them they said they were too long and dry. Sarah organizes a new set of videos to be made and rolls out an update.
+
+3 months later Sarah gets a slack message from PostHog to view the retention graph. They can see there's a significant correlation between retention between users that watched a video and users who retained. Success!
```
EDsCODE, line 48: Simple reminders like this would be great

requests-for-comments/2022-12-01-workflow-new-feature.md
lharries: Could we give them the tools to customize the create a new feature workflow? E.g. they can edit the schema which decides which graphs, session recordings, feature flag setups are made. We could use the new JSON based stuff that Marius is making with some template variables

This way we can give them some starter workflows but they can also edit it or create their own to suit their processes

neilkakkar:
This is definitely an interesting idea!

I don't think we should build this right now though as I don't think we're ready for opinionated workflows on top of the core functionality. Here's how I'm thinking about this:

1. The value-add of any core feature (which enables something new that wasn't possible before / was frustrating / too hard to do) is more than any feature that explains _how_ to do those things (i.e. workflows).

    For example, I'd be a lot more happier (as a PostHog user) if we built JSON flag support this quarter, rather than a workflow on top of existing features.

2. Well, (1) isn't true in the absolute, and I think there will come a point when workflows can help guide users in the right direction when it's easy to get lost in the product. This makes sense when there's a significantly big chunk of users who want to do things the same way but aren't able to for whatever reason. The cost of these workflows is the same as any other opinionated flow: we make things hard for users not using that workflow, who then end up not using the workflow at all.

   I'm not quite sure how people use flags & what workflow would makes sense (I'm not convinced the above-mentioned workflow is the way to go), so I'd propose gathering this data by building the blocks of core functionality that the workflow demands but isn't possible right now, without actually building the workflow. So, it's something like breaking down what we build into:
   1. Reminders to check statuses of flags X time from now.
   2.  Ability to generate a customisable recordings playlist for flag & notifications/slack when these are > X. 
   .. and then talking to users to figure out which parts of flag features they use the most (& also raw data), to come up with a workflow.

PR #87, opened 2023-01-12
Rename our plans

raquelmsmith:
> Plan names are something seemingly minor that actually get used quite a lot, both internally and by our customers (oftentimes just inside their own heads when evaluating). However, our current plan names leave room for confusion:
> 
> * The names don't indicate right at the get-go which plan you are probably interested in based on your company size/scale/needs. 
> * The names can include information about deployment but lack any real plan name (eg. PostHog Cloud is a paid tier but it's also a deployment)
> * We use the term self-serve in the docs but no one really knows what this means

See the file for more details :)

### DIFF
--- a/requests-for-comments/2023-01-12-plan-naming.md
+++ b/requests-for-comments/2023-01-12-plan-naming.md
@@ -6,7 +6,7 @@ _Who are we building for, what are their needs (include example cases), why is t
 Plan names are something seemingly minor that actually get used quite a lot, both internally and by our customers (oftentimes just inside their own heads when evaluating). However, our current plan names leave room for confusion:
 
 * The names don't indicate right at the get-go which plan you are probably interested in based on your company size/scale/needs. 
-* The names can include information about deployment but lack any real plan name (eg. PostHog Cloud as a paid tier but it's also a deployment)
+* The names can include information about deployment but lack any real plan name (eg. PostHog Cloud is a paid tier but it's also a deployment)
 * We use the term self-serve in the docs but no one really knows what this means
 
 ## Proposal
--- a/requests-for-comments/2023-01-12-plan-naming.md
+++ b/requests-for-comments/2023-01-12-plan-naming.md
@@ -21,7 +21,7 @@ So, I'd propose that our Cloud plans are the _default_ plans and are named:
   - I think this speaks to our goal to have post-PMF companies who are wanting to grow quickly be on paid plans
 - **Enterprise**
 
-Since we are sunsetting the k8s deployment, I don't think we will be selling licenses for self-hosted versions any more, correct? If that is the case then the only other option is the self-hosted docker deployment. We could simply call this the **Open-source** plan.
+Since we are sunsetting the k8s deployment, the only other option is the self-hosted docker deployment. We could simply call this the **Open-source** plan.
 
 ## Success criteria
 _How do we know if this is successful (i.e. metrics, customer feedback), what's out of scope, whats makes this ambitious?_
--- a/requests-for-comments/2023-01-12-plan-naming.md
+++ b/requests-for-comments/2023-01-12-plan-naming.md
@@ -10,7 +10,7 @@ Plan names are something seemingly minor that actually get used quite a lot, bot
 * We use the term self-serve in the docs but no one really knows what this means
 
 ## Proposal
-The Cloud environment was new a couple years ago, so it made sense to call the plans on Cloud the "Cloud" plans. However, the Cloud environment is by far the default way to use PostHog now, and calling it Cloud is confusing (I can imagine thinking, "should I be here? or is there a different non-cloud thing I should use?").
+The Cloud environment was new a couple years ago, so it made sense to call the plans on Cloud the "Cloud" plans. However, the Cloud environment is by far the default way to use PostHog now, and calling the plans "Cloud" is confusing (I can imagine thinking, "should I be here? or is there a different non-cloud thing I should use?").
 
 So, I'd propose that our Cloud plans are the _default_ plans and are named:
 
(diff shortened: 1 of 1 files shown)


requests-for-comments/2023-01-12-plan-naming.md
```diff
@@ -0,0 +1,47 @@
...
+* We use the term self-serve in the docs but no one really knows what this means
+
+## Proposal
+The Cloud environment was new a couple years ago, so it made sense to call the plans on Cloud the "Cloud" plans. However, the Cloud environment is by far the default way to use PostHog now, and calling the plans "Cloud" is confusing (I can imagine thinking, "should I be here? or is there a different non-cloud thing I should use?").
+
+So, I'd propose that our Cloud plans are the _default_ plans and are named:
+
+- **Starter**
+  - I prefer Starter over "Free" because "free" can get confused with the FOSS version
+  - I prefer Starter over "Hobby" because someone might be getting a new startup up and running, and they are very serious about it and working full time on it and are definitely not considering themselves a hobbyist, but this is still the correct plan for them
```
benjackwhite, line 19: I like this reasoning 👍

benjackwhite (approved):

requests-for-comments/2023-01-12-plan-naming.md
```diff
@@ -0,0 +1,47 @@
...
+* The names don't indicate right at the get-go which plan you are probably interested in based on your company size/scale/needs. 
+* The names can include information about deployment but lack any real plan name (eg. PostHog Cloud is a paid tier but it's also a deployment)
+* We use the term self-serve in the docs but no one really knows what this means
+
+## Proposal
+The Cloud environment was new a couple years ago, so it made sense to call the plans on Cloud the "Cloud" plans. However, the Cloud environment is by far the default way to use PostHog now, and calling the plans "Cloud" is confusing (I can imagine thinking, "should I be here? or is there a different non-cloud thing I should use?").
+
+So, I'd propose that our Cloud plans are the _default_ plans and are named:
+
+- **Starter**
```
kappa90, line 17: If the user doesn't have a card in, are they on a plan? The hardest philosophical question ever 😅

requests-for-comments/2023-01-12-plan-naming.md
```diff
@@ -0,0 +1,47 @@
...
+  - I prefer Starter over "Free" because "free" can get confused with the FOSS version
+  - I prefer Starter over "Hobby" because someone might be getting a new startup up and running, and they are very serious about it and working full time on it and are definitely not considering themselves a hobbyist, but this is still the correct plan for them
+- **Scale**
+  - I think this speaks to our goal to have post-PMF companies who are wanting to grow quickly be on paid plans
+- **Enterprise**
+
+Since we are sunsetting the k8s deployment, I don't think we will be selling licenses for self-hosted versions any more, correct? If that is the case then the only other option is the self-hosted docker deployment. We could simply call this the **Open-source** plan.
```
kappa90, line 24: We are, but at the 5k$ enterprise tier.

lharries (approved):
Agreed with this!

requests-for-comments/2023-01-12-plan-naming.md
raquelmsmith: Are you sure? https://github.com/PostHog/posthog.com/pull/5073

ellie (approved):
This makes a lot of sense 🙌

corywatilo:
Note that we phased out the "Scale" terminology on the website a long time ago. We have references to these plan types in docs/guides:

image

image

raquelmsmith:
@corywatilo yes, updating all the docs will be part of executing this RFC. We're actually planning on hooking the docs up to the billing service so it's all automatically populated.

corywatilo:
I suggest we go with very basic, straightforward plan names (to coincide with [this design](https://github.com/PostHog/posthog.com/issues/5151)):

- Free
- Pay-per-use or Not free or Unlimited

The more straightforward, the cheekier aligns with our brand and our audience tends to appreciate.

jamesefhawkins:
I think "free" is genius for the main plan that most people will use

(I get the arguments around 'starter' being better than 'hobbyist', but I think 'free' is much more of an improvement especially with our focus here)

as for the others, i'd vote for simply 'paid' (literal opposite of free / super clear) and 'enterprise' (custom contracts)

raquelmsmith:
**Re: Free plan.** 
I love the name "Free" for the free plan, but my concern was that it would get confused with the FOSS version. Do you think there is any concern with that?

**Re: Paid plans.** 
My hesitation with just "Paid" is that it leaves room for confusion when discussing. "Are they on a paid plan?" "Yes." "Ok is that the regular paid plan or the paid enterprise plan?"

My hesitation with "Pay-per-use" is similar - on Enterprise you also pay per use. But it makes me wonder if enterprise _isn't_ pay-per-use, maybe it's some other pricing model?

"Not free" - this is kinda fun but also doesn't give someone any reason to upgrade. Immediately makes me think "well that sucks, it's not free." Ha.

"Unlimited" - I'm _okay_ with this, but doesn't address one of the [issues](https://github.com/PostHog/meta/pull/87/files#diff-cd8b7eb9a6cfc082c475b0208e3ac009dcd287398fa6a877690a670e547cef81R8) that brought this RFC to light in the first place:
> The names don't indicate right at the get-go which plan you are probably interested in based on your company size/scale/needs.

--

Regardless, any of these names would be an improvement over what we currently have :) I guess my question is what is the priority here - internal clarity or being on-brand (eg. with a cheeky name like "Not Free")?

charlescook-ph:
Cheekier still (feel free to ignore this flyby comment):
- Free
- Cheap
- Expensive

(Or 'Totally Free', 'Great Value', 'Reassuringly Expensive')

raquelmsmith:
I'm gonna stick with Free, Scale, and Enterprise for now. I think there's too much confusion to be had when someone sees something like "Paid" in the docs and isn't sure what it means. The nice thing is that once everything is hooked up it will be very easy to change plan names and have it propagate everywhere!

PR #96, opened 2023-02-15
RFC: proposal for improving our client libraries

lharries:
Decision make/Owner: Luke
Date planning to make the decision: Wednesday next week

What's the decision: Should we try 2 months of working with contractors on improving the developer experience and getting to feature parity for the libraries. If this works, we will continue to work with the contractors. If not, we will re-evaluate.

Feedback requested:
- Input on the open questions

### DIFF
--- a/requests-for-comments/2023-02-15-improving-our-client-libraries.md
+++ b/requests-for-comments/2023-02-15-improving-our-client-libraries.md
@@ -1,8 +1,10 @@
 # RFC: Getting our client libraries up to scratch
 
-## Summary:
+## Summary
 
 Eric raised a number of problems around the developer experience for our libraries (listed below). Additionally, Cameron and Simon have mentioned that often customers are requesting feature parity for the libraries that we have. As there’s a large number of languages and spikey amounts of work this is likely a task better suited to a few contractors. We will do a trial with enlisting contractors to first improve the developer experience and get to feature parity. And then setup a process for the contractors to maintain feature parity for the non-primary libraries.
+
+What's the decision: Should we try 2 months of working with contractors on improving the developer experience and getting to feature parity for the libraries. If this works, we will continue to work with the contractors. If not, we will re-evaluate.
 
 ## Problems
 
@@ -20,7 +22,7 @@ Custom problems:
 - Missing feature parity for a number of libraries:
     - [https://posthog.com/docs/integrate/libraries](https://posthog.com/docs/integrate/libraries)
 
-## Where we want to get to:
+## Where we want to get to
 
 - Test coverage, CI/CD, and quality docs across all the repos
 - PostHog core team implements the features for the “Primary” libraries (the ones we use ourselves): Javascript, Python and Node.js
@@ -29,7 +31,14 @@ Custom problems:
 - TBD who reviews the core dependencies? Should we instead go with individuals that have good knowledge of those libraries
 - Luke to get the system setup for working with the contractors via a shared slack and github project. Each team as they ship features work directly with the contractors for feature implementation. Luke responsible for making sure the ongoing core work and maintenance happens.
 
-## Open questions:
+## What does success look like?
+
+- Stable libraries that are kept up to date with the core features
+- Good developer experience for the libraries
+- Reliable contractors that don't require too much management or excessive PR reviewing
+- Not spending too much money
+
+## Open questions
 
 - Which team should be responsible for reviewing the PRs for core features?
 - Would we want a contractor implementing autocapture for mobile?
@@ -41,7 +50,7 @@ Custom problems:
 - Eric mentioned that the segment libraries have gone through some larger changes. Is there anything that we can learn from this?
 - Are there any libraries that we are missing that we want to add?
 
-## Rough plan:
+## Rough plan
 
 **Owned by Luke**
 
--- a/requests-for-comments/2023-02-15-improving-our-client-libraries.md
+++ b/requests-for-comments/2023-02-15-improving-our-client-libraries.md
@@ -1,0 +1,54 @@
+# RFC: Getting our client libraries up to scratch
+
+## Summary:
+
+Eric raised a number of problems around the developer experience for our libraries (listed below). Additionally, Cameron and Simon have mentioned that often customers are requesting feature parity for the libraries that we have. As there’s a large number of languages and spikey amounts of work this is likely a task better suited to a few contractors. We will do a trial with enlisting contractors to first improve the developer experience and get to feature parity. And then setup a process for the contractors to maintain feature parity for the non-primary libraries.
+
+## Problems
+
+Developer experience problems:
+
+- Incomplete CI/CD
+- Lack of test coverage
+- Random legacy dependencies/development environments that don’t work
+- Patchy documentation
+- “Cross compatibility nuances (kotlin -> java, swift -> objc)”
+
+Custom problems:
+
+- Lack of 1st class support for React and [NextJS](https://github.com/PostHog/posthog-js/issues/492) (very common for high-potential and high-growth customers)
+- Missing feature parity for a number of libraries:
+    - [https://posthog.com/docs/integrate/libraries](https://posthog.com/docs/integrate/libraries)
+
+## Where we want to get to:
+
+- Test coverage, CI/CD, and quality docs across all the repos
+- PostHog core team implements the features for the “Primary” libraries (the ones we use ourselves): Javascript, Python and Node.js
+- Contractors implement new features and the core team reviews the PRs for the “Secondary” libraries (all the other ones). Aiming to keep feature parity within a few weeks
+- Each team is responsible for reviewing their features.
+- TBD who reviews the core dependencies? Should we instead go with individuals that have good knowledge of those libraries
+- Luke to get the system setup for working with the contractors via a shared slack and github project. Each team as they ship features work directly with the contractors for feature implementation. Luke responsible for making sure the ongoing core work and maintenance happens.
+
+## Open questions:
+
+- Which team should be responsible for reviewing the PRs for core features?
+- Would we want a contractor implementing autocapture for mobile?
+- We seem to have posthog-js-lite and posthog-js, what’s the plan and timeline for these? Do we want contractor support? (This seems more core)
+- For test coverage, do we want to aim for a % test coverage? Are we also wanting cypress style tests?
+- What profile do we want? Ideally we’d have some overlap of people e.g. 1 person to do all the mobile libraries, and up to 2 people for the server libraries
+    - I’d assume we’d want a senior engineer
+- How do we want to solve: “Cross compatibility nuances (kotlin -> java, swift -> objc)”? Do we want to build a swift library?
+- Eric mentioned that the segment libraries have gone through some larger changes. Is there anything that we can learn from this?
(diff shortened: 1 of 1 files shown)


lharries:
@simfish85 @camerondeleone any major requests from customers for things missing from our client libraries?

benjackwhite:
I have lots of specific issues with using contractors for this kind of work which I will try and find the time to comment on line-by-line but as a summary:

I don't believe contractors will need us to "10x" better libraries. "2x" would already be surprising. Hiring a bunch of people to come in and throw code at the problem will for sure bandaid some of the things we have in mind (for an incredibly high cost) but I don't believe we would know "did they do a good job" until months later when a customer is using the SDK and either having a good or a bad time.

I feel like there is much more work we can do internally that will actually have a 10x effect like:

1. Defining functionality for a generic "SDK" as a spec with separate parts for client-based and server-based
  - This would answer so many questions like that are currently answered by "well lets see what the js library does and copy that"
3. Scoping our work more appropriately
  - Local evaluation of feature flags as an example. This accounts for roughly half of the code in the server-side python/node libs. We should understand "We have 10 libraries we want to support this, therefore that is 10x the work.
  - I wonder if with that effort discussion in mind we would have approached the solution differently?
4. Documenting (and simplifying) our API
  - The more we can offload to the API, the stronger things can be. For many customers (think Java) they could integrate with is super quickly if we offered a solid, well documented OpenAPI spec.
  - This helps point 1

If we don't do that work, then I think hiring contractors will result in much more lost-time internally, explaining and debugging things with them. 

It feels like we don't have proper ownership over the client-sdks like we do with "pipeline" or "infrastructure" yet I see these as equally as core...

requests-for-comments/2023-02-15-improving-our-client-libraries.md
```diff
@@ -0,0 +1,63 @@
...
+- Reliable contractors that don't require too much management or excessive PR reviewing
+- Not spending too much money
+
+## Open questions
+
+- Which team should be responsible for reviewing the PRs for core features?
+- Would we want a contractor implementing autocapture for mobile?
+- We seem to have posthog-js-lite and posthog-js, what’s the plan and timeline for these? Do we want contractor support? (This seems more core)
+- For test coverage, do we want to aim for a % test coverage? Are we also wanting cypress style tests?
+- What profile of contractor do we want? Ideally we’d have some overlap of people e.g. 1 person to do all the mobile libraries, and up to 2 people for the server libraries
```
benjackwhite, line 47: I think 1 person per-mobile platform as there is rarely strong crossover between iOS and Android native devs. If we are hiring a contractor, to me it only makes sense to do so if they are an expert in that field.

requests-for-comments/2023-02-15-improving-our-client-libraries.md
```diff
@@ -0,0 +1,63 @@
...
+- Good developer experience for the libraries
+- Reliable contractors that don't require too much management or excessive PR reviewing
+- Not spending too much money
+
+## Open questions
+
+- Which team should be responsible for reviewing the PRs for core features?
+- Would we want a contractor implementing autocapture for mobile?
+- We seem to have posthog-js-lite and posthog-js, what’s the plan and timeline for these? Do we want contractor support? (This seems more core)
+- For test coverage, do we want to aim for a % test coverage? Are we also wanting cypress style tests?
```
benjackwhite, line 46: Cypress only applies to web? 

If we are hiring a contractor then I'd want someone who is smart enough to figure this out for themselves. Depending on the platform, testing strategies can be wildly different so I'd rather we say "We want solid test coverage such that we can confidently make non-breaking changes in the future".

requests-for-comments/2023-02-15-improving-our-client-libraries.md
```diff
@@ -0,0 +1,63 @@
...
+## What does success look like?
+
+- Stable libraries that are kept up to date with the core features
+- Good developer experience for the libraries
+- Reliable contractors that don't require too much management or excessive PR reviewing
+- Not spending too much money
+
+## Open questions
+
+- Which team should be responsible for reviewing the PRs for core features?
```
benjackwhite, line 43: This brings me to the core issue I have - who owns this? Individual teams might care about their part but they won't care (or have the context and knowledge) about the overall quality of the SDK

requests-for-comments/2023-02-15-improving-our-client-libraries.md
```diff
@@ -0,0 +1,63 @@
...
+- Would we want a contractor implementing autocapture for mobile?
```
benjackwhite, line 44: This is a perfect example of - this would be really easy for me to do if we had it as a priority. The question is - is it a priority :D 

Autocapture has a lot of complex parts that effect the rest of the core product. I don't know if we could spec this out clear enough for a contractor to build it faster than we could just built it...

requests-for-comments/2023-02-15-improving-our-client-libraries.md
```diff
@@ -0,0 +1,63 @@
...
+- We seem to have posthog-js-lite and posthog-js, what’s the plan and timeline for these? Do we want contractor support? (This seems more core)
```
benjackwhite, line 45: This was always a bit of an experiment to see if 
A) We could unify our Node/React-Native/web core
B) Move over the "core-js" to this. 

I think a contractor would have a high likelyhood of failing here as the challenge is not the code, it's the context of what does and does not need to be supported in the future.

If we sort that out, perhaps?

requests-for-comments/2023-02-15-improving-our-client-libraries.md
```diff
@@ -0,0 +1,63 @@
...
+- For test coverage, do we want to aim for a % test coverage? Are we also wanting cypress style tests?
+- What profile of contractor do we want? Ideally we’d have some overlap of people e.g. 1 person to do all the mobile libraries, and up to 2 people for the server libraries
+    - I’d assume we’d want a senior engineer
+- How do we want to solve: “Cross compatibility nuances (kotlin -> java, swift -> objc)”? Do we want to build a swift library?
```
benjackwhite, line 49: I actually mostly built a swift library whilst working on Mobile session recordings.

Again, the challenge isn't the code here as much as it is coming up with a migration strategy / testing it in the field to get enough certainty that it is a stable rewrite.

Wrt Kotlin for example - this could make total sense as we approach Mobile Recordings and where an expert in the Android field would be great to advise and support on using all the best approaches

lharries:
@benjackwhite thanks for your input!

To check I understand the gist of what you're saying - is your preference that we'd first invest some core team first and then bring in some "expert" contractor support? Or something else? No worries if you're not sure but pointing out some pitfalls with the current proposal

benjackwhite:
> To check I understand the gist of what you're saying - is your preference that we'd first invest some core team first and then bring in some "expert" contractor support? Or something else? No worries if you're not sure but pointing out some pitfalls with the current proposal

Yes and also perhaps a general caution of "I've been down this road before and it went very badly very quickly". 
I guess I don't want us to underestimate the challenge and potential damage of "outsourcing" this work, versus "collaborating" on this work.

lharries:
> I don't believe contractors will need us to "10x" better libraries. "2x" would already be surprising

Yep I renamed the RFC but realized I didn't rename the file. I think the goal shouldn't be 10x but instead more like 2x - get them reliable and (when appropriate) improve feature parity

camerondeleone:
From my perspective, the main concern is clear expectations/documentation over which libraries support which features. This avoids surprises during the sales /onboarding process (when we cs team, as people who understand stuff at a surface level, are prone to make mistakes in interpreting what's available and what's not). This is infinitely easier if we have something like @benjackwhite proposes above:

> Defining functionality for a generic "SDK" as a spec with separate parts for client-based and server-based

I don't hear a ton of complaints from stuff missing, but maybe that's because mobile was historically lacking groups and feature flags, and yet the api supported them and so it was workaround-able. Our early customers seemed to tolerate it, given how technical they were (though probably less true as we grow and people want more polish).

requests-for-comments/2023-02-15-improving-our-client-libraries.md
EDsCODE: Big +1 here as it's the main reason I brought up this topic. The libraries can _work_ with our ad hoc coverage right now but _working_ vs having _best in class_ developer experience as we're building for developers is a big difference

requests-for-comments/2023-02-15-improving-our-client-libraries.md
EDsCODE: Yeah, thinking more on contractors, we may be over assuming how thorough they will be. I have no experience to qualify my hunches but I'd guess that we would need someone on our end "managing" the contractors work quite closely to get the results we're asking for here

EDsCODE (commented):
Thinking more on this, I agree with many of Ben's points. 

1. We probably need owners/teams owning a library completely because even with a contractor, it might take a good bit of management to get what we want. 
2. Just a note: there's a lot of expertise sprinkled throughout Posthog that could probably handle these libraries but given how our teams are shaped, it usually makes sense to care about the library if there's some interest related to something you're trying to do in your main team (session recordings wanting recordings on a library, experiments wanting a feature to be possible, pipeline wanting capture to work a certain way)
3. Before making a decision either way on this, it would make sense to try to gather more concrete evidence that it's a _now_ problem. My original motivation is that it's a _now_ discussion so the team can get on the same page as there's really not much attention here.
4. The first step (which I'm working on for feature flags specifically) could just be to have clear documentation on what's supported and not supported across libraries

requests-for-comments/2023-02-15-improving-our-client-libraries.md
jamesefhawkins: I wonder if when we do a CDP, we should put SDKs into that team's responsibility?

(and I'm also unsure if we'll manage to execute a CDP with our current pipeline team - ie it got deprioritised this quarter - since we always wind up having to prioritise scalability or firefighting)

zachkirsch:
Hey Posthog team!

Hope it's cool to barge into this thread :slightly_smiling_face:. Saw you are looking to improve your client libraries, and thought [Fern](https://www.buildwithfern.com/) might be helpful. We're in the current YC batch and also open source!

**What is Fern?** Fern auto-generates client libraries for REST APIs. You can start with OpenAPI, or for better results, a [Fern Definition](https://docs.buildwithfern.com/definition).

**How can Fern be helpful?** In Posthog's case, my guess is the SDK is split between (1) types, (2) API calls, and (3) custom logic (e.g. local evaluation of feature flags). Fern can be helpful for (1) and (2) - generating all the types + networking logic. Auto-generating from a single source of truth means you can eliminate an entire class of "API details incorrect / out of sync" bugs throughout your client libraries. I expect these rails would be especially helpful if you opt for using contractors.

Would love to chat! [Here's](https://calendly.com/zachkirsch/30min) my calendly if you're up for it.

Some Fern-generated SDKs if you're curious:
  - https://github.com/ravenappdev/raven-node
  - https://github.com/flipt-io/flipt-java

lharries:
Update on this - Tim is exploring a solution with Fern to help with codegen and maintenance. Additionally, for the mobile libraries we'll likely make another hire here at somepoint rather than using contractors.

Therefore pausing this for now

PR #99, opened 2023-03-02
RFC: Temporal at PostHog

fuziontech:
This is a long one. Feel free to reach out directly to me if you have questions or need a tl;dr

### DIFF
--- a/requests-for-comments/2023-03-01-Temporal-RFC.md
+++ b/requests-for-comments/2023-03-01-Temporal-RFC.md
@@ -11,11 +11,13 @@ Hi there, dear reader. I would love for you to scrutinize this RFC and the follo
 - Is this a good idea?
 - Is getting this done feasible?
 
-I’m specifically looking for feedback from teams that are directly impacted by this, namely:
+I’m specifically looking for feedback from teams that are directly impacted by this, namely and in alphabetical order:
 
-- Infrastructure
 - Billing
+- Feature Flags and Experiments
+- Infrastructure
 - Pipeline
+- Product Analytics
 
 # Problem Statement
 
--- a/requests-for-comments/2023-03-01-Temporal-RFC.md
+++ b/requests-for-comments/2023-03-01-Temporal-RFC.md
@@ -1,0 +1,239 @@
+# RFC: Temporal at PostHog
+
+Author: James Greenhill
+
+Date: February 27, 2023 
+
+# For the reader
+
+Hi there, dear reader. I would love for you to scrutinize this RFC and the following RFC with more specifics on how we can build out our workflows. I’m looking for two things in particular from you though:
+
+- Is this a good idea?
+- Is getting this done feasible?
+
+I’m specifically looking for feedback from teams that are directly impacted by this, namely:
+
+- Infrastructure
+- Billing
+- Pipeline
+
+# Problem Statement
+
+We have a number of tasks that we’d like to run and have the confidence that they ran to completion. It would be even nicer to have some confidence that we were able to run it to completion exactly once despite hardware issues, deployments, or James going in and restarting a machine manually. So far we’ve built a few different ways to do this that are all quite different: Celery, Kafka, and Graphile. Each of these is a worker queue or worker queue adjacent that we have to maintain and keep flowing. The goal of this RFC is to propose a new, consolidated solution that we can use to focus more on our business logic (the product that we are actually building) and less on making sure that that logic actually runs. It’s a lot easier to reason about a product if you don’t have to think as much about it’s failure modes, retry logic, and timeouts.
+
+## Requirements
+
+We have a few products that are good candidates for a workflow framework. We’ll cover a few of the requirements here before we jump into any proposals just so that we can have a better understanding of what we are looking for.
+
+### Billing
+
+Billing is one of those things that is really tricky. Ideally you don’t want to bill someone twice (oof). You also don’t want to think you bill someone and then not bill them at all. Ideally you can kick off a flow and at the end of it have confidence that everything worked. The system was able to grab the usage metrics, apply the billing rate, and charge the user without having to double check the usage and the stripe api. Even though I still believe in trust, but verify. Any billing mistakes are generally pretty embarrassing.
+
+Requirements:
+
+- **Idempotent** or **exactly once** - no duplicates
+- **Retry-able** - If the thing fails, keep trying to get the money we are owed
+- **Resumable** - nice to pick up where you left off
+- **Composable** - Query ClickHouse. Calculate bill. Charge stripe.
+- **Parallelize-able** - We should be able to run multiple bills at the same time
+- **Timeouts** - We should be able to set timeouts on steps
+- **Cancelable** - We should be able to cancel any billing gracefully
+- **Terminate-able** - We should be able to rip the carpet out from a billing routine
+- **Nonblocking failures** - A failure of one billing workflow should not halt the execution of others.
+
+### Celery tasks
+
+Our celery tasks are pretty diverse. We use it for collecting metrics, refreshing caches, and if you didn’t know - we used to even use it to process every single event 😱.
+
+What do we use Celery for now?
+
+In general they are all scheduled tasks:
+
+- Running queries and emitting metrics (statsd, prometheus)
+- Warming up caches
+
+Requirements:
+
+- **Retry-able** - If the thing fails, get back on that horse
+- **Resumable** - Some of these are heavy and we don’t want to requery CH
+- **Parallelize-able** - We should be able to run multiple tasks at once
+- **Timeouts** - We should be able to set timeouts on the task
+- **Terminate-able** - We should be able to rip the carpet out from under the Task
+- **Nonblocking failures** - A failure of one task should not halt the execution of others.
+
+### Historical Exports *(AKA Vanilla ETL)*
+
+We do a lot of moving around data. Actually, a good chunk of our business is just getting data form A to B safe and sound, and in on piece. Some of these jobs are short, some are long running. There are a few requirements and nice to haves around ETLs
+
+Requirements:
+
+- **Idempotent** or **exactly once** - no duplicates
+- **Retry-able** - If the thing fails, get back on that horse
+- **Resumable** - Nice to pick up where you left off
+- **DAG-able** - Most ETLs resemble a Directed Acyclical Graph. Move A → B → C
+    - You can think of this as **Composable**
+- **Parallelize-able** - We should be able to run multiple bits of the DAG at once
+- **Timeouts** - We should be able to set timeouts on the DAG and Nodes
+- **Cancelable** - We should be able to cancel the DAG’s execution *gracefully*
+- **Terminate-able** - We should be able to rip the carpet out from under the DAG
+- **Nonblocking failures** - A failure of one DAG should not halt the execution of others.
+
+### Webhooks
+
+Ahh webhooks. Webhooks are a particularly sticky problem.
+
+Requirements:
+
+- **Idempotent** or **exactly once** - no duplicates
+- **Retry-able** - If the thing fails, get back on that horse
+- **Resumable** - Helps with idempotency/exactly once
+- **Parallelize-able** - We should be able to run multiple webhooks at once and out of order (non-blocking)
+- **Timeouts** - We should be able to set timeouts for webhooks
+- **Terminate-able** - We should be able to rip the carpet out from under webhooks
+- **Nonblocking failures** - A failure of one webhook endpoint should not halt the execution of others.
+
+## Requirements Grid
+
+| Task | Idempotent / Exactly Once | Retry-able | Resumable | Parallelize-able | Timeouts | Cancelable | Terminate-able | Nonblocking Failures |
+| --- | --- | --- | --- | --- | --- | --- | --- | --- |
+| Billing | 💯 | 💯 | 💯 | 🙏 | 🙏 | 🙏 | 💯 | 💯 |
+| Celery Task |  | 💯 | 🙏 | 💯 | 💯 |  | 💯 | 💯 |
+| ETLs | 💯 | 💯 | 💯 | 💯 | 💯 | 💯 | 💯 | 💯 |
+| Webhooks | 💯 | 💯 | 🙏 | 💯 | 🙏 |  | 💯 | 💯 |
+
+🙏 - Really nice to have
+
+💯 - Required
+
+# Where are we at now?
+
+Currently our plugin server has been doing all of the heavy lifting and has gotten us quite far in the **************2 years************** that it has existed. I once gave feedback that the plugin server is a diamond encrusted hammer that tempts you into thinking that everything is a nail. Joke was on me because it turned out that it got us way further than I thought.
+
+Since this began we’ve done amazing work in keeping the plugin server going. Chief among the improvements is that we split the plugin server up dependent on the workload into pools of resources that are responsible for different categories of plugin.
+
+### The plugin server, a survey
+
+So what all is the plugin server responsible for? There are really two types of plugins or tasks that we operate on the plugin server: Stateful and Stateless. Stateful plugins make side effects when they are called. This means that an API is called or a mutation is made in a db. Stateless plugins only affect the event as it is in flight. A good example of a stateless plugin is GeoIP where we take the ip that is in the event and based on that information and a static GeoIP db we geolocalize using the ip. 
+
+************************Stateful vs Stateless************************
+
+The practical difference between these two is that a stateless plugin generally executes in a deterministic way. This is mainly because it depends only on its inputs to execute. A stateful plugin on the other hand does not execute so deterministically. This is because of the fact that it depends on something outside of the inputs to the function, like an API. As an example let’s think of a plugin that during execution calls out to an api to either mutate or augment the data that is in flight. This is stateful because depending on the time that this plugin runs the results may vary. If this is a plugin that augments data with the response of an API the data returned by the API may change request to request. The api also may fail or stall. In our case most of these APIs are external to us and may fail or degrade at any point. This makes error handling of stateful plugins very important.
+
+- Plugins
+    - Stateless
+        - Transformations
+    - Stateful
+        - Exports
+        - API Calls
+        - Reverse ETLs
+        - Augmenters
+- Webhooks (Stateful)
+
+So what’s the pain point currently?
+
+Our biggest pain point currently is with the stateful plugins. Regardless of whether they are scheduled or realtime because they depend on external resources that can degrade or fail they require sophisticated timeout and retry logic that we don’t have currently. Another pain point is a relative lack of visibility into the system when things fail. The final pain point is that we don’t have the ability to easily compose these plugins together. What we need is a workflow framework. We are halfway through building one.
+
+You might ask yourself, James, if we are halfway through building a workflow engine, then why don’t we just finish it!? That’s a great question. The real reason we should not complete it is that frameworks already exist to do just this and building another one is not our competitive advantage against our competitors. Building one is a distraction from us building out our actual business.
+
+What is a compounding factor here is that this isn’t just a plugin server issue but an issue that is felt across the product: Async migrations, billing, and PoE are all tasks that fit a workflow engine really well. There will be other tasks that we need to run in the future that will need a solid framework for executing workflows as well. This allows us to focus on these and not something that is outside of our expertise and product space.
+
+# Temporal
+
+Temporal is a workflow engine that provides a way for developers to build and manage distributed applications. It is designed to help developers focus on their business logic by abstracting away the details of failure modes, retry logic, and timeouts. Temporal allows developers to build workflows that describe complex processes as a set of simple, easy-to-understand steps. It provides a programmable interface that enables developers to write the code that defines their workflows and the logic that controls how they execute.
+
+Temporal is used by a number of companies, including Uber, DoorDash, Instacart, HashiCorp, and Coinbase. It has become a popular choice among developers because it provides a way to manage complex distributed systems without having to worry about the underlying infrastructure. Temporal allows developers to build workflows that are resilient, scalable, and fault-tolerant, which is essential for building applications that can handle large volumes of data and traffic.
+
+One of the key features of Temporal is its ability to provide visibility into the status of workflows. It provides a dashboard that allows developers to monitor the progress of their workflows and to see where errors are occurring. This helps developers to quickly identify and fix problems, which is essential for ensuring that workflows are running smoothly.
+
+Overall, Temporal is a powerful tool that enables developers to build complex distributed applications and workflows. It provides a way to manage the complexity of distributed systems and to build applications that are scalable, resilient, and fault-tolerant.
+
+[What is temporal?](https://docs.temporal.io/temporal)
+
+[How does temporal work?](https://temporal.io/how-temporal-works)
+
+### Why is it a good fit?
+
+Temporal is an especially good fit because it provides things that are genuinely hard to build and maintain ourselves:
+
+- Retry logic per Activity and Workflow
+- Timeout logic per activity and workflow
+- Exactly-once semantics (!!)
+- Go, [Python](https://docs.temporal.io/kb/python-sandbox-environment), and Typescript libraries that support deterministic code execution and retries (!!)
+- Composability of Activities and Workflows
+- Dashboard for observing Workflow status
+
+Finally, Temporal is a good fit because we are a small team. We should be focusing on the business of building our product. Nothing about what we are building that would be run on Temporal is novel enough to justify us to invest in a competing runtime.
+
+### Who uses it?
+
+- **Uber** (original developer - originally called [Cadence](https://github.com/uber/cadence))
+- **Coinbase**
+- **Doordash**
+- **Instacart**
+- **Hashicorp**
+- **Netflix**
+- **Snapchat**
+- **Stripe**
+- **Datadog**
+- **Box**
+- **Airbyte**
+- A whole lot more…
+
+### Quick Concept Overview
+
+A workflow in [Temporal.io](http://temporal.io/) is a set of simple, easy-to-understand steps that describe a complex process. Workflows are made up of activities, which are individual steps in the workflow that perform specific tasks. Activities can be executed in parallel, and Temporal provides retry logic, timeout logic, and exactly-once semantics. Workflows and activities can be composed together, and Temporal provides a dashboard for observing workflow status.
+
+# Example Workflow
+
+We currently use it for our manual backfills to S3 for certain clients. You can see an example workflow here:
+
+[https://github.com/posthog/energize](https://github.com/posthog/energize)
+
+Note: that this will be folded into the Posthog/Posthog repo. For now this is separate for example purposes.
+
+The specific workflow for S3 exporting can be found [here](https://github.com/PostHog/energize/blob/main/src/energize/workflows/exports/s3.py).
+
+# Temporal Cloud
+
+Why ☁️?
+
+There is one critical piece of the Temporal topology that causes loss if it goes down: The scheduler. As long as the scheduler is up. When you schedule a job - it will be completed. If the scheduler is down for an extended period of time there is a chance that the requesting service would drop the intent and that workflow would never be run. As long as the scheduler is up and we are with the tunable TTL of the workflow, as soon as a worker comes back online we will process that workflow.
+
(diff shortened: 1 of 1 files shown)


requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+
+# Phases of work
+
+**Phase 1: Manual mode -** We disable historical exports in the app interface. This is the buggiest feature that larger high value customers are complaining about. This transitions into a manual mode where if a customer wants a historical export they request that we run one for them and we do that manually for them using a Temporal workflow. This allows us to be iterative and to start gathering operational knowledge around Temporal.
+
+**Phase 2: Build the Platform -** We stand up Temporal cloud as our scheduler and add worker deployments to K8s. This will build the foundation for migrating tasks to Temporal and unblock those that are interested in migrating their workflows to Temporal. Manual exports will run on the Temporal workers that are deployed on the K8s cluster. Temporal worker code is in the PostHog/PostHog monorepo
+
+**Phase 3: API-ification -** We rewire the historical exports interfaces to schedule Temporal workflows. The api is already there (built for the manual backfills). The only difference now is that we are scheduling workflows from the PostHog application through the user interface
```
timgl, line 231: I'd lvoe a bit more detail around this point. Does this mean each plugin will have code in two places (real time and historical), written in potentially two different languages? Will we port all the plugins?

mariusandra (commented):
Hey, help me understand some points, for which I didn't get answers in the text. I believe going through the Temporal docs would give some answers, but asking directly to make sure everyone's on the same page.

To be perfectly clear, I'm not saying not to do it, even if the tone of my questions might hint that way. I'd just like to understand both the big picture, and the smaller details.

1. Is this going to replace Kafka, and we will send every event through this?
2. Is this only for "batch jobs" or also for the "real time" ingestion?
3. What are the performance characteristics of this? 10x slower than Kafka? 5x more memory than Celery? Just some ballpark figures to know where to mentally position this tech.
4. Where does temporal store in flight tasks/events?
5. What will be remaining of the plugin server once this is implemented as proposed? Will users be able to write custom JS (assuming they have permissions, etc) that will then get executed through these temporal workers, or are we removing support for "custom JS" altogether with this?
6. How long will it take to build this, migrate all things over, and won't we have the [XKCD problem](https://xkcd.com/927/) for a while?
7. Asking the latter point because I'm support this week. Customers are currently complaining about not being able to reliably export data, even from our cloud to our cloud. All of them get a "contact Jams" reply as instructed, but then no followup (see #support-pipeline for the trail). Their needs are simple. They want to export historic + live events, and not loose any nor get duplicates. Why is the solution to this problem to build a whole new world, instead of a function "schedule a historic event until $now when I activate the plugin". With 2% of the effort we could solve 98% of current customer problems. 
8. Can we prioritise responding to those customers please?
9. I'm also very very very suspicious and hesitant about relying on yet another third party cloud service for something that's the core of our business. Why do we want to lock ourselves into other dependencies?

lharries:
Something which isn't mentioned explicitly is the future CDP requirements. Also if temporal (or other solutions) would also enable us to integrate with other MIT CDP connectors that would be ideal (but not a blocker if not)

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+# Where are we at now?
+
+Currently our plugin server has been doing all of the heavy lifting and has gotten us quite far in the **************2 years************** that it has existed. I once gave feedback that the plugin server is a diamond encrusted hammer that tempts you into thinking that everything is a nail. Joke was on me because it turned out that it got us way further than I thought.
+
+Since this began we’ve done amazing work in keeping the plugin server going. Chief among the improvements is that we split the plugin server up dependent on the workload into pools of resources that are responsible for different categories of plugin.
+
+### The plugin server, a survey
+
+So what all is the plugin server responsible for? There are really two types of plugins or tasks that we operate on the plugin server: Stateful and Stateless. Stateful plugins make side effects when they are called. This means that an API is called or a mutation is made in a db. Stateless plugins only affect the event as it is in flight. A good example of a stateless plugin is GeoIP where we take the ip that is in the event and based on that information and a static GeoIP db we geolocalize using the ip.
```
hazzadous, line 116: Ah I'd misunderstood what stateless and stateful plugins were. I thought it was a reference to if it was updating the [Node.js VM context](https://nodejs.org/api/vm.html#vm-executing-javascript) with team specific data such that you can't share a VM.

Will have to go and read the code again to understand properly 🔍

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+
+We do a lot of moving around data. Actually, a good chunk of our business is just getting data form A to B safe and sound, and in on piece. Some of these jobs are short, some are long running. There are a few requirements and nice to haves around ETLs
+
+Requirements:
+
+- **Idempotent** or **exactly once** - no duplicates
+- **Retry-able** - If the thing fails, get back on that horse
+- **Resumable** - Nice to pick up where you left off
+- **DAG-able** - Most ETLs resemble a Directed Acyclical Graph. Move A → B → C
+    - You can think of this as **Composable**
```
tomasfarias, line 74: A corollary of this requirement could be cycle-detection, in case we ever (accidentally) write a workflow that spins forever...

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+[How does temporal work?](https://temporal.io/how-temporal-works)
+
+### Why is it a good fit?
+
+Temporal is an especially good fit because it provides things that are genuinely hard to build and maintain ourselves:
+
+- Retry logic per Activity and Workflow
+- Timeout logic per activity and workflow
+- Exactly-once semantics (!!)
+- Go, [Python](https://docs.temporal.io/kb/python-sandbox-environment), and Typescript libraries that support deterministic code execution and retries (!!)
```
tomasfarias, line 161: Don't forget PHP! https://github.com/temporalio/sdk-php/

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+
+You might ask yourself, James, if we are halfway through building a workflow engine, then why don’t we just finish it!? That’s a great question. The real reason we should not complete it is that frameworks already exist to do just this and building another one is not our competitive advantage against our competitors. Building one is a distraction from us building out our actual business.
+
+What is a compounding factor here is that this isn’t just a plugin server issue but an issue that is felt across the product: Async migrations, billing, and PoE are all tasks that fit a workflow engine really well. There will be other tasks that we need to run in the future that will need a solid framework for executing workflows as well. This allows us to focus on these and not something that is outside of our expertise and product space.
+
+# Temporal
```
tomasfarias, line 140: The other big workflow orchestration tool is Apache Airflow. I feel that a section comparing both would fit nicely.

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+Luckily [Temporal.io](http://Temporal.io) has a hosted version of their scheduler where we can just submit our workflows and connect our workers to and the rest should be relatively smooth sailing. We only have to worry about writing the workflows themselves and keeping the workers happy.
+
+# Ownership
+
+This is why I’ve wanted to get you all here. Here is how ownership here would most likely shake out:
+
+**Infrastructure**
+
+- K8s Deployment of the workers that execute Workflows and Actions
+- CD of these workers
```
tomasfarias, line 211: I would be interested to understand more how does Temporal's worker abstraction map to K8s: is it image with a long-running process that we just spin up? Is it more on-demand (like how Airflow uses `KubernetesOperator`)?

tomasfarias: Also, I feel this responsibility will likely be shared with other teams. At least when it comes to the determination and updating of worker requirements.

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+
+**Temporal cloud**
+
+- The scheduler itself. As long as the workflows are scheduled, they will be executed at some point in the future as long as a worker starts back up and the TTL does not expire the job.
+
+# Phases of work
+
+**Phase 1: Manual mode -** We disable historical exports in the app interface. This is the buggiest feature that larger high value customers are complaining about. This transitions into a manual mode where if a customer wants a historical export they request that we run one for them and we do that manually for them using a Temporal workflow. This allows us to be iterative and to start gathering operational knowledge around Temporal.
+
+**Phase 2: Build the Platform -** We stand up Temporal cloud as our scheduler and add worker deployments to K8s. This will build the foundation for migrating tasks to Temporal and unblock those that are interested in migrating their workflows to Temporal. Manual exports will run on the Temporal workers that are deployed on the K8s cluster. Temporal worker code is in the PostHog/PostHog monorepo
```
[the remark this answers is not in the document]
ellie: They can deploy Temporalite if they're on hobby (basically to temporal what SQLite is to a SQL server), or docker compose deployment

requests-for-comments/2023-03-01-Temporal-RFC.md
ellie: > is it image with a long-running process that we just spin up

Afaik yes, you basically compile your workflows + activities into an executable, which then uses the temporal SDK to pull tasks + run your code

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+        - API Calls
+        - Reverse ETLs
+        - Augmenters
+- Webhooks (Stateful)
+
+So what’s the pain point currently?
+
+Our biggest pain point currently is with the stateful plugins. Regardless of whether they are scheduled or realtime because they depend on external resources that can degrade or fail they require sophisticated timeout and retry logic that we don’t have currently. Another pain point is a relative lack of visibility into the system when things fail. The final pain point is that we don’t have the ability to easily compose these plugins together. What we need is a workflow framework. We are halfway through building one.
+
+You might ask yourself, James, if we are halfway through building a workflow engine, then why don’t we just finish it!? That’s a great question. The real reason we should not complete it is that frameworks already exist to do just this and building another one is not our competitive advantage against our competitors. Building one is a distraction from us building out our actual business.
```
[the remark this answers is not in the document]
ellie: > "building one is us building out our actual business"

I would compare us building our own workflow engine to us building our own OLAP database. We didn't do the latter, and found success because of our choice. Many companies/projects have crashed and burned trying to do the former, when _really_ they just wanted to run business logic/data processing reliably + scalably, not faff around with distributed systems.

So far the majority of our customers have been small. We still do not process truly **big** data, and are struggling with what we already have. If we'd like to progress properly we need to choose rock solid foundations, that have already been proved at way beyond the scale we currently see.

requests-for-comments/2023-03-01-Temporal-RFC.md
ellie: Temporal doesn't do this just yet, you can basically just spawn child workflows (that are the same as the parent) forever. It's an intentional behaviour, though there has been an open issue on their github for ages and ages about being able to limit the stack depth there

requests-for-comments/2023-03-01-Temporal-RFC.md
tomasfarias: Does that mean there is no "hot reload" of workflows? We do need to re-compile with every workflow change?

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+What do we use Celery for now?
+
+In general they are all scheduled tasks:
+
+- Running queries and emitting metrics (statsd, prometheus)
+- Warming up caches
+
+Requirements:
+
+- **Retry-able** - If the thing fails, get back on that horse
```
xvello, line 57: Being pedantic, I'd say these tasks have a different definition of `Retry-able` than billing tasks, because they are scheduled on a tighter loop.
For timeseries generation (and probably dashboard reloads too), we want to deduplicate tasks when retrying. When the failure cause resolves, I don't want to run my task N times, just once.

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+- **Idempotent** or **exactly once** - no duplicates
+- **Retry-able** - If the thing fails, keep trying to get the money we are owed
+- **Resumable** - nice to pick up where you left off
+- **Composable** - Query ClickHouse. Calculate bill. Charge stripe.
+- **Parallelize-able** - We should be able to run multiple bills at the same time
+- **Timeouts** - We should be able to set timeouts on steps
+- **Cancelable** - We should be able to cancel any billing gracefully
+- **Terminate-able** - We should be able to rip the carpet out from a billing routine
+- **Nonblocking failures** - A failure of one billing workflow should not halt the execution of others.
+
```
xvello, line 43: Although it's obvious for both of us, I'd explicitly add an **Observable** requirement: we want to alert on tasks being delayed too long / dropped, and we want to identify the cause quickly and reliably.

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+Requirements:
+
+- **Idempotent** or **exactly once** - no duplicates
+- **Retry-able** - If the thing fails, get back on that horse
+- **Resumable** - Helps with idempotency/exactly once
+- **Parallelize-able** - We should be able to run multiple webhooks at once and out of order (non-blocking)
+- **Timeouts** - We should be able to set timeouts for webhooks
+- **Terminate-able** - We should be able to rip the carpet out from under webhooks
+- **Nonblocking failures** - A failure of one webhook endpoint should not halt the execution of others.
+
```
xvello, line 94: One difficulty with webhooks is the risk of being rate-limited by the destination, either on a `team*destination` basis, or even on a `destination, for all teams` basis. Most sources will nicely return a `429`, some might silently black-hole traffic.
Ideally, we would be able to rate-limit task execution rate on our end (by these dimensions), to buffer spikes and ease-in the volume when recovering from incident. We'd ideally have the choice of delaying (backfilling) or dropping (input spike).

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+🙏 - Really nice to have
+
+💯 - Required
+
+# Where are we at now?
+
+Currently our plugin server has been doing all of the heavy lifting and has gotten us quite far in the **************2 years************** that it has existed. I once gave feedback that the plugin server is a diamond encrusted hammer that tempts you into thinking that everything is a nail. Joke was on me because it turned out that it got us way further than I thought.
+
+Since this began we’ve done amazing work in keeping the plugin server going. Chief among the improvements is that we split the plugin server up dependent on the workload into pools of resources that are responsible for different categories of plugin.
```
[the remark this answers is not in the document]
xvello: We chose Kubernetes as our workload scheduler and, although I have lots of negative feelings towards k8s, I support using its capabilities instead of rebuilding them ourselves.

We are at a growth stage where incidents will be common (both due to unpredictable input volumes, and self-inflicted bugs), and one of the things we need to prioritize is reducing the blast radius of these incidents. Splitting up resources is a foolproof way to achieve this.

requests-for-comments/2023-03-01-Temporal-RFC.md
xvello: Agreed on `Stateful and Stateless` being a bit confusing to me too, what you're describing is closer to the `pure / impure` function concept?

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+Temporal is an especially good fit because it provides things that are genuinely hard to build and maintain ourselves:
+
+- Retry logic per Activity and Workflow
+- Timeout logic per activity and workflow
+- Exactly-once semantics (!!)
+- Go, [Python](https://docs.temporal.io/kb/python-sandbox-environment), and Typescript libraries that support deterministic code execution and retries (!!)
+- Composability of Activities and Workflows
+- Dashboard for observing Workflow status
+
+Finally, Temporal is a good fit because we are a small team. We should be focusing on the business of building our product. Nothing about what we are building that would be run on Temporal is novel enough to justify us to invest in a competing runtime.
```
xvello, line 165: Agreed, these systems can be intellectually satisfying to write, but extremely hard to debug at 3AM.
Given our upcoming scaling challenges, I'm happy to go with boring but reliable.

requests-for-comments/2023-03-01-Temporal-RFC.md
xvello: @tomasfarias my understanding is that any pod can be a worker: they just open a grpc connection to the scheduler, advertise their capabilities and wait for incoming tasks.

While the heaviest tasks will be new dedicated services, we could also add temporal capabilities to the existing services, when that makes sense.

requests-for-comments/2023-03-01-Temporal-RFC.md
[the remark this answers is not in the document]
xvello: We'll add temporal to the compose stack (for both local devenv and hobby deploy), and most probably the chart.
If we strip down the UI, temporal is a single golang container. We can either reuse the same PG as PostHog (on a separate db), or use its sqlite backend with a persistent volume.

requests-for-comments/2023-03-01-Temporal-RFC.md
neilkakkar: What Harry is saying is indeed how we treat it in plugin server, which is probably not what it means in functional programming /rest of the world.

For plugins, having a config => stateful, even though it might not have side effects. So: ex: advanced-geoip is stateful, when all it does is what geoIP does, but with an extra config option which decides which events to add info to.

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+
+The specific workflow for S3 exporting can be found [here](https://github.com/PostHog/energize/blob/main/src/energize/workflows/exports/s3.py).
+
+# Temporal Cloud
+
+Why ☁️?
+
+There is one critical piece of the Temporal topology that causes loss if it goes down: The scheduler. As long as the scheduler is up. When you schedule a job - it will be completed. If the scheduler is down for an extended period of time there is a chance that the requesting service would drop the intent and that workflow would never be run. As long as the scheduler is up and we are with the tunable TTL of the workflow, as soon as a worker comes back online we will process that workflow.
+
+Luckily [Temporal.io](http://Temporal.io) has a hosted version of their scheduler where we can just submit our workflows and connect our workers to and the rest should be relatively smooth sailing. We only have to worry about writing the workflows themselves and keeping the workers happy.
```
xvello, line 202: > should be relatively smooth sailing

Let's assume they'll have outages please :)

I think we can assume they won't have data loss, but we should be resilient to failures when submitting workflow executions.

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+Hi there, dear reader. I would love for you to scrutinize this RFC and the following RFC with more specifics on how we can build out our workflows. I’m looking for two things in particular from you though:
+
+- Is this a good idea?
+- Is getting this done feasible?
+
+I’m specifically looking for feedback from teams that are directly impacted by this, namely:
+
+- Infrastructure
+- Billing
+- Pipeline
```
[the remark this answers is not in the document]
fuziontech: Why? These are the teams that are directly going to have to deal with the implications here.

I'd love your feedback as well, but you shouldn't have to live with the day to day implications here. I'm just making sure those teams take a look and understand how this may effect them.

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+- **Coinbase**
+- **Doordash**
+- **Instacart**
+- **Hashicorp**
+- **Netflix**
+- **Snapchat**
+- **Stripe**
+- **Datadog**
+- **Box**
+- **Airbyte**
```
tiina303, line 179: what do they use it for?
There's a big difference in using an external scheduling system for something real time vs something that just needs to run daily/weekly/... reliably. Asking because if temporal has an outage we are not going to be able to do much about it if it's very core except schedule the tasks manually directly.

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,239 @@
...
+
+- Workflows and Actions written to be executed on Temporal workers
+
+**Pipeline**
+
+- Same, Workflows and actions written and scheduled to be executed on Temporal workers
+
+**Temporal cloud**
+
+- The scheduler itself. As long as the workflows are scheduled, they will be executed at some point in the future as long as a worker starts back up and the TTL does not expire the job.
```
tiina303, line 223: who owns Temporal cloud as a dependency management - their outages etc?

requests-for-comments/2023-03-01-Temporal-RFC.md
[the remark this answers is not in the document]
fuziontech: Companies don't starve, they drown.

requests-for-comments/2023-03-01-Temporal-RFC.md
fuziontech: Sorry I should have been more clear here because these are terms that we use in the Plugin server. I am referring to the Functional context of stateful and stateless. My original draft actually used side-effecty and side effect less, but I swapped that out. 

I just found it a bit easier to categorize things if I though about our plugins as either being pure or side-effecty functions. The pure functions are much easier to reason about from a performance perspective, the side-effecty ones are not so easy and we need to consider many other ways that they can fail considering they have external dependencies.

hazzadous: Makes sense 👍

requests-for-comments/2023-03-01-Temporal-RFC.md
xvello: Because temporal supports mixing tasks written in different languages within a workflow, I hope we'll be able to shim existing apps into temporal tasks (both for imports and exports)
While there's value in making a few most-used integrations more "temporal-native", I'd like to address the long tail with such a shim.

requests-for-comments/2023-03-01-Temporal-RFC.md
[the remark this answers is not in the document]
fuziontech: Ahh this is a really great point that I should have included in the RFC. I'll make a change to include what on prem and dev will look like.

requests-for-comments/2023-03-01-Temporal-RFC.md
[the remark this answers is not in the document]
mariusandra: Well, scheduling long running tasks is something we're doing all the time in product analytics. Be it simple things like insight or dashboard refreshes, long running export queries, calculations to get usage counts for events or properties, you name it. I'm not even touching cohorts, which are technically not part of team product analytics.

I feel like there's a big divide simmering between app and platform, and the worlds should come closer together, not drift further apart. Hence, I wrote that I feel explicitly excluded.

requests-for-comments/2023-03-01-Temporal-RFC.md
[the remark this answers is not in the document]
mariusandra: I hope we can just agree to disagree here then. I won't stand in anyone's way (like I wrote 2x in my original comment), just wanted to understand the tech better and got many answers.

However, I _do_ think we _absolutely should_ write our own k8s, otherwise we're open for disruption. I'd prefer we're the ones doing the disruption.

Don't take that statement 100% literally - no, we shouldn't build exactly what the Kubernetes Platform is, or even close. And no, we don't have the resources to undertake these things right now.

But we should stop fearing hard engineering challenges! Whether it's a query engine, a mobile session recording toolkit, an orchestration layer or anything in between. Let's instead be the company that has solved these problems and is offering them to our users as part of an integrated solution. 🤷

requests-for-comments/2023-03-01-Temporal-RFC.md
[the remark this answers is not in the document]
mariusandra: I'm not here to tell anyone that you _can_ do something when you're all saying you can't. :)

requests-for-comments/2023-03-01-Temporal-RFC.md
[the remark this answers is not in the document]
fuziontech: Ah, that's constructive feedback. Thank you.

Some direct feedback: 'I feel explicitly excluded' is not constructive feedback.
I would also argue that the comment does suggests 'bad intent' which is not a great place to start. I can tell you with authority that I did not explicitly exclude you.

I agree and hadn't considered that. You all should be added 💯 . These are great fits and your input is valuable as such!
- Cache refreshes (for insights and dashboards)
- Tasks for collecting statistics for Taxonomy

It's also a great point with cohorts. This RFC should include experimentation as well as they would be impacted by this.

I'm a bit confused as to why 'long running export queries' is something Product Analytics maintain? Could you explain that a bit? I agree it should fit into this just would like to understand who's owning it to makes sure they are also included.

>I feel like there's a big divide simmering between app and platform, and the worlds should come closer together, not drift further apart. Hence, I wrote that I feel explicitly excluded.

A huge reason why I have spent so much time on this RFC is to make sure that we work together closer. I'm trying to extend an olive branch out to every impacted team that this proposal effects. What would push us even further apart is for pipeline to ship this entirely in isolation, without the input from the effected teams, effectively alienating and isolating them from being able to provide feedback until it is too late.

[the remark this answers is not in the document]
mariusandra: > I'm a bit confused as to why 'long running export queries' is something Product Analytics maintain?

Well, there's this button on the events table that runs through a script:

image

And we're building Direct SQL access. I don't know what's your definition of "long running" in this case, but for me it's anything with timeouts > 60sec that get delivered async.

> What would push us even further apart is for pipeline to ship this entirely in isolation, without the input from the effected teams, effectively alienating and isolating them from being able to provide feedback until it is too late.

I've been feeling the same. I see all these problems, and a clear path on how to fix them. So I keep pushing out my own RFCs, giving brown bags, pushing ideas in meetings, etc... and then only getting negative and discouraging feedback in response. The feedback has always been ultimately useful and guided the direction in some way, but always phrased as "don't do this" instead of "tweak it a bit so".  Then I have nothing else to do but to build things in isolation, and watch the 360 feedback write itself :).

requests-for-comments/2023-03-01-Temporal-RFC.md
```diff
@@ -0,0 +1,241 @@
...
+This is why I’ve wanted to get you all here. Here is how ownership here would most likely shake out:
+
+**Infrastructure**
+
+- K8s Deployment of the workers that execute Workflows and Actions
+- CD of these workers
+
+**Billing**
+
+- Workflows and Actions written to be executed on Temporal workers
```
kappa90, line 217: Talked with Tim, and we might indeed use Temporal for billing.

RIght now our flow is:
* Calculate an aggregate of a bunch of metric, make a report out of it
* Send the report to posthog
* Send the report to the billing service
- Billing service aggregates all reports in the period and sends them to Stripe
- Stripe handles the charges

The first three steps need to be reliable (work every day, no matter what, for each organization), and sending events to PostHog and to the billing service *needs* to happen. If that fails (and it has failed) we undercharge users.

fuziontech: This makes sense and seems like a pretty solid fit honestly.

requests-for-comments/2023-03-01-Temporal-RFC.md
fuziontech: That's a great question - I'd propose joint ownership between Infra and Pipelines.

requests-for-comments/2023-03-01-Temporal-RFC.md
fuziontech: Great question. At Uber it was used for very high frequency workflows. You've probably even been a customer who's order has gone through Cadence if you have ever ordered something on UberEats.

https://www.uber.com/blog/open-source-orchestration-tool-cadence-overview/
https://www.youtube.com/watch?v=-LRghQzfF8k

Though, keep in mind I am not advocating here that we process each event through this. The highest frequency would be webhook jobs.

PR #143, opened 2023-10-25
Per product NPS survey

annikaschmid:


requests-for-comments/2023-10-25-per-product-nps.md
```diff
@@ -0,0 +1,47 @@
...
+- Link to draft survey for feature flags: https://app.posthog.com/surveys/018b665b-7396-0000-25ad-59cd7c843b4e
+- Question 1:
+    
+  <img width="346" alt="Screenshot 2023-10-25 at 14 28 21" src="https://github.com/PostHog/meta/assets/14750837/5c58d7f9-706f-4407-adcd-0d4571b3f460">
+    
+- Question 2:
+
+  <img width="343" alt="Screenshot 2023-10-25 at 14 29 05" src="https://github.com/PostHog/meta/assets/14750837/e2388e9b-0eb5-4749-ad18-00ef3ee2c3f2">
+  
+- Thank you: Get a free PostHog sticker (optional):
```
annikaschmid, line 30: @corywatilo I know we do this for applicants. Would this be easy to set this up for surveys, and send users to a URL where they can get a sticker, that only works once? If not, we can also de-prioritise this

annikaschmid:
@mariusandra @benjackwhite @neilkakkar FYI, since your teams are responsible for these products

requests-for-comments/2023-10-25-per-product-nps.md
```diff
@@ -0,0 +1,47 @@
...
+  
+- Thank you: Get a free PostHog sticker (optional):      
+
+    <img width="350" alt="Screenshot 2023-10-25 at 14 29 55" src="https://github.com/PostHog/meta/assets/14750837/5661f99a-e0d3-470e-92ba-d0f3d83d53a8">
+
+  Note: There is a bug with the button, which is why there isn’t a more engaging CTA yet
+
+## Release conditions
+
+- Visitors on any feature flags page (app & eu). I am using the following regex: `https://(app|eu)\.posthog\.com/feature_flags`
```
raquelmsmith, line 38: can we make it so that they only get it if they signed up 30 days ago or something? we really don't want to ask for nps score if someone just signed up...

requests-for-comments/2023-10-25-per-product-nps.md
```diff
@@ -0,0 +1,47 @@
...
+    - Note: Once we have behavioural surveys, we can update the targeting to only show to users who have created a flag / watched 10 recordings / created a survey / created an insight / created an experiment
```
corywatilo, line 39: Yep, seems like there are some good places to drop these in - when we know a customer has completed a significant action:

- Product analytics: Saved an insight, added an insight to a dashboard, etc
- Session replay: Shared a replay? etc
- Flag: Enabled a flag
- A/B test: Ended a test
- Survey: Launched a survey

Is the plan to launch this once we have behavioral surveys?

requests-for-comments/2023-10-25-per-product-nps.md
corywatilo: Yep, or we can just use a shared code (like we do for applicants). When we're ready, hit me up and I can generate.

requests-for-comments/2023-10-25-per-product-nps.md
annikaschmid: Good point, I think we can do this. I'll update the targeting

requests-for-comments/2023-10-25-per-product-nps.md
annikaschmid: I was planning to launch this before we have behavioural surveys, because I don't want to wait.

That means, the targeting would be they are on the e.g. feature flags page AND have signed up at least 30 days ago. It will be less accurate than what we can do once we have behavioural surveys, but I think it's worth collecting feedback already, even if we know it will not be perfect.

requests-for-comments/2023-10-25-per-product-nps.md
annikaschmid, line 38:
```suggestion
- Visitors on any feature flags page (app & eu). I am using the following regex: `https://(app|eu)\.posthog\.com/feature_flags`
- Visitors who have signed up at least 30 days ago: `joined_at` before `date`
```

requests-for-comments/2023-10-25-per-product-nps.md
annikaschmid: @corywatilo I am just waiting for a quick bug fix from the feature success team, then this is ready to go. If I could have this link by the end of the week / early next week, this would be great.

requests-for-comments/2023-10-25-per-product-nps.md
annikaschmid: @raquelmsmith Since I've updated the targeting to only target activated users, I have removed the 'signed up before' condition

requests-for-comments/2023-10-25-per-product-nps.md
annikaschmid: I've set this up with cohorts. It's quite hacky, but it works.

raquelmsmith:
>We should add a question which aligns with the role_at_organization property, asking users if they are engineering, founder, product, sales, marketing, or other.

shouldn't we know this already based on the info they provide at signup? (I know not everyone answers this question, but we can maybe draw conclusions from those who do)

> remove the 'last 14 days' targeting, as we can't easily target this correctly

I really think we don't want to ask people for NPS scores if they've just joined. Is there some way we can still do this delay?

joethreepwood:
> shouldn't we know this already based on the info they provide at signup? (I know not everyone answers this question, but we can maybe draw conclusions from those who do)

If someone knows a way to correlate this, sure. My main interest is: What roles are using which products, and how do we perform for them? I don't mind how we get that info. 

> I really think we don't want to ask people for NPS scores if they've just joined. Is there some way we can still do this delay?

I agree 100%, I just couldn't see a way to implement this delay. At the moment the survey is only showing to users who signed up before 2023-11-01, which also isn't ideal because we're not getting current info.

annikaschmid:
> I agree 100%, I just couldn't see a way to implement this delay. At the moment the survey is only showing to users who signed up before 2023-11-01, which also isn't ideal because we're not getting current info.

Yes, because we can't do < 14 days properly, I need to regularly update this date manually every couple of weeks. Which is not ideal, but I don't think there is a better solution right now. This is European format, so we are only slightly out of date. I will set myself a calendar reminder so that I won't forget.

> If someone knows a way to correlate this, sure. My main interest is: What roles are using which products, and how do we perform for them? I don't mind how we get that info.

I just had a look how we could get this. If you want to see this on the survey results page, here is how: https://posthog.slack.com/archives/C0351B1DMUY/p1702399749511089 
We have data for ~47% of users, which I think is ok as a start.

annikaschmid:
> lower the limit to users who have seen another survey to exclude users who've seen a survey in the last 14 days
> increase the rollout from the current 20% to 50% of users

Done. I've rolled this out according to your proposal, and also updated the signup date. I will have a look to see if this makes a dent in the response rate.

> Remove the free sticker

Done. I replaced it with a auto-disappearing "Thank you for your feedback!" message.

> Rollout to product analytics

Going to do this this week. I will also set this up for session replay.

PR #145, opened 2023-10-31
RFC: Feature Flag environments

jurajmajerik:
We want to add the ability to copy a Feature Flag to another project. Please see this RFC for more details.

image

### DIFF
--- a/requests-for-comments/2023-10-31-feature-flag-environments.md
+++ b/requests-for-comments/2023-10-31-feature-flag-environments.md
@@ -1,0 +1,70 @@
+# Request for comments: Feature Flag environments
+
+## Problem statement
+
+_Who are we building for, what are their needs (include example cases), why is this important?_
+
+Users find it difficult to work with Feature Flags across environments. To use a flag in a different environments, they must copy it and all its configuration manually.
+
+## Context
+
+_What are our competitors doing, what are the technical constraints, what are customers asking for, what does the data tell us, are there external motivations (e.g. launch week, enterprise contract)?_
+
+See the earlier [RFC Feature flag environments](https://github.com/PostHog/product-internal/blob/main/requests-for-comments/2022-12-08-feature-flag-environments.md#221115---rfc-feature-flag-environments) from @annikaschmid.
+
+## Design
+
+_What are the key user experience and technical design decisions / trade-offs?_
+
+### Option a) Add the ability to _copy or move_ a feature flag to another project.
+
+#### Pros
+
+-   Our users already use projects as environments and copy their flags manually. This solution would free them from manual work (prone to errors).
+-   Easiest to implement
+
+#### Cons
+
+-   Projects are semantically different from environments, which may not align with the needs of larger clients with multiple apps (projects).
+
+### Option b) Introduce environments to Feature Flags only
+
+#### Pros
+
+-   Cleaner separation between projects and environments.
+-   Ability to roll out environment to other parts of the app incrementally and as needed.
+
+#### Cons
+
+-   May introduce confusion by having environments for Feature Flags but not other features like Surveys.
+
+### Option C) Introduce environments app-wide (i.e. dashboards, insights, etc.)
+
+#### Pros
+
+-   Cleanest separation of concerns, respects the distinction between projects and environments.
+
+#### Cons
+
+-   Most complex to implement
+-   Introduces another layer of complexity where it might not be needed (e.g. dashboards).
+
+Based on Annika's RFC and some preliminary discussion, I recommend implementing **Option a) Add the ability to _copy_ or _move_ a feature flag to another project.** This is the quickest win which addresses the users' requirements from the [RFC user feedback](https://github.com/PostHog/product-internal/blob/main/requests-for-comments/2022-12-08-feature-flag-environments.md#user-feedback).
+
+## Implementation
+
+In the Feature Flag view page, users will have the option to "Sync" the feature flag to another projects. To do so, they'll click the Sync button, after which a modal pops up. They can choose the project to sync the flag to via a dropdown.
+
+If they select a project where the flag with the identical ID already exists, a warning will show telling them they are about to overwrite the other project's flag configuration.
+
+Depending on whether the flag exists in the other project, we'll either `PATCH` that existing flag, or `POST` a new feature flag. Alternatively, we can can implement a `PUT` endpoint to handle both scenarios.
+
+<img width="1068" alt="Screenshot 2023-10-31 at 09 53 08" src="https://github.com/PostHog/posthog/assets/22996112/432abf9a-b8cc-4d46-a6e6-a5e5ab586b7e">
+
+### Nice to have
+
+It would be useful for users to have an overview of projects that use the current flag - otherwise, they would have to switch projects to check for this. We can add this overview to the vacant space below the "Insights that use this feature flag" list.
+
+<img width="1166" alt="Screenshot 2023-10-31 at 09 53 21" src="https://github.com/PostHog/posthog/assets/22996112/86df3710-60c5-48e0-b1f7-7ab27b1fd80e">
+
+Alternatively, we can add this overview to the new tab "Projects" within the flag view page. If we decide to do so, it would make sense to also move the Sync button there. Having a separate Projects tab might make sense if we want to add more "project flags" operations in the future. This way, the projects list and the related operations would be nicely grouped under a tab.
--- a/requests-for-comments/2023-10-31-feature-flag-environments.md
+++ b/requests-for-comments/2023-10-31-feature-flag-environments.md
@@ -16,7 +16,7 @@ See the earlier [RFC Feature flag environments](https://github.com/PostHog/produ
 
 _What are the key user experience and technical design decisions / trade-offs?_
 
-### Option a) Add the ability to _copy or move_ a feature flag to another project.
+### Option a) Add the ability to copy a feature flag to another project.
 
 #### Pros
 
@@ -49,7 +49,7 @@ _What are the key user experience and technical design decisions / trade-offs?_
 -   Most complex to implement
 -   Introduces another layer of complexity where it might not be needed (e.g. dashboards).
 
-Based on Annika's RFC and some preliminary discussion, I recommend implementing **Option a) Add the ability to _copy_ or _move_ a feature flag to another project.** This is the quickest win which addresses the users' requirements from the [RFC user feedback](https://github.com/PostHog/product-internal/blob/main/requests-for-comments/2022-12-08-feature-flag-environments.md#user-feedback).
+Based on Annika's RFC and some preliminary discussion, I recommend implementing **Option a) Add the ability to copy a feature flag to another project.** This is the quickest win which addresses the users' requirements from the [RFC user feedback](https://github.com/PostHog/product-internal/blob/main/requests-for-comments/2022-12-08-feature-flag-environments.md#user-feedback).
 
 ## Implementation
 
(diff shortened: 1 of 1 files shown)


requests-for-comments/2023-10-31-feature-flag-environments.md
```diff
@@ -0,0 +1,72 @@
...
+
+If they select a project where the flag with the identical ID already exists, a warning will show telling them they are about to overwrite the other project's flag configuration.
+
+Depending on whether the flag exists in the other project, we'll either `PATCH` that existing flag, or `POST` a new feature flag. Alternatively, we can can implement a `PUT` endpoint to handle both scenarios.
+
+<img width="1068" alt="Screenshot 2023-10-31 at 09 53 08" src="https://github.com/PostHog/posthog/assets/22996112/432abf9a-b8cc-4d46-a6e6-a5e5ab586b7e">
+
+_Note that the original [RFC](https://github.com/PostHog/product-internal/blob/main/requests-for-comments/2022-12-08-feature-flag-environments.md#user-feedback) also mentions the ability to `move` a flag to another project. I believe we don't need to implement this explicitely. To achieve this, users can simply copy the flag to another environment, and immediately delete the flag in the current environment._
+
+### Nice to have
```
neilkakkar, line 66: another nice to have: sync all feature flags from one project to another.

requests-for-comments/2023-10-31-feature-flag-environments.md
```diff
@@ -0,0 +1,72 @@
...
+
+Based on Annika's RFC and some preliminary discussion, I recommend implementing **Option a) Add the ability to copy a feature flag to another project.** This is the quickest win which addresses the users' requirements from the [RFC user feedback](https://github.com/PostHog/product-internal/blob/main/requests-for-comments/2022-12-08-feature-flag-environments.md#user-feedback).
+
+## Implementation
+
+In the Feature Flag view page, users will have the option to "Sync" the feature flag to other projects. To do so, they'll click the Sync button, after which a modal pops up. They can choose the project to sync the flag to via a dropdown.
+
+If they select a project where the flag with the identical ID already exists, a warning will show telling them they are about to overwrite the other project's flag configuration.
+
+Depending on whether the flag exists in the other project, we'll either `PATCH` that existing flag, or `POST` a new feature flag. Alternatively, we can can implement a `PUT` endpoint to handle both scenarios.
```
neilkakkar, line 60: patch & post make sense 👍

requests-for-comments/2023-10-31-feature-flag-environments.md
```diff
@@ -0,0 +1,72 @@
...
+
+<img width="1068" alt="Screenshot 2023-10-31 at 09 53 08" src="https://github.com/PostHog/posthog/assets/22996112/432abf9a-b8cc-4d46-a6e6-a5e5ab586b7e">
+
+_Note that the original [RFC](https://github.com/PostHog/product-internal/blob/main/requests-for-comments/2022-12-08-feature-flag-environments.md#user-feedback) also mentions the ability to `move` a flag to another project. I believe we don't need to implement this explicitely. To achieve this, users can simply copy the flag to another environment, and immediately delete the flag in the current environment._
+
+### Nice to have
+
```
neilkakkar, line 67: another potential future thing: What if I want to sync across orgs / sync to local dev?

I don't think it's worth doing this right now, because currently it feels like only our team would want to do this, but could be nice to export and import a flag json.

Mentioning here because it's worth designing the sync API with this in mind, even if we don't build it right now. (i.e. sync API can take multiple flags, and json with flag info to make this update, there's ideally no custom parameters & special handling here that works only within the project)

requests-for-comments/2023-10-31-feature-flag-environments.md
```diff
@@ -0,0 +1,72 @@
...
+It would be useful for users to have an overview of projects that use the current flag - otherwise, they would have to switch projects to check whether a flag exists in a particular environment. We can add an overview to the vacant space below the "Insights that use this feature flag" list.
+
+<img width="1166" alt="Screenshot 2023-10-31 at 09 53 21" src="https://github.com/PostHog/posthog/assets/22996112/86df3710-60c5-48e0-b1f7-7ab27b1fd80e">
```
neilkakkar, line 70: There's too many buttons on the feature flag now, let's move everything to the left of vertical bar into a 3 dot menu like in surveys & experiments?

neilkakkar (approved):
This makes sense 👍 

I think there will be a bit more back & forth on the actual UI/UX, but that doesn't need to be resolved in the RFC, so go for it! We might uncover new issues when implementing syncing, and the faster we find those, the quicker we'll settle on the final version of the UI.

requests-for-comments/2023-10-31-feature-flag-environments.md
```diff
@@ -0,0 +1,72 @@
...
+-   Most complex to implement
+-   Introduces another layer of complexity where it might not be needed (e.g. dashboards).
+
+Based on Annika's RFC and some preliminary discussion, I recommend implementing **Option a) Add the ability to copy a feature flag to another project.** This is the quickest win which addresses the users' requirements from the [RFC user feedback](https://github.com/PostHog/product-internal/blob/main/requests-for-comments/2022-12-08-feature-flag-environments.md#user-feedback).
+
+## Implementation
+
+In the Feature Flag view page, users will have the option to "Sync" the feature flag to other projects. To do so, they'll click the Sync button, after which a modal pops up. They can choose the project to sync the flag to via a dropdown.
+
+If they select a project where the flag with the identical ID already exists, a warning will show telling them they are about to overwrite the other project's flag configuration.
```
neilkakkar, line 58: Once the basic version is in place, also worth considering how this will play with permissions, what if I have permissions to do ops in this project, but not the other one? How do we communicate this to users?

requests-for-comments/2023-10-31-feature-flag-environments.md
neilkakkar: also, love the mocks, nice work! :) Where did you make them?

requests-for-comments/2023-10-31-feature-flag-environments.md
neilkakkar: ... would be nice to get some more detail on how exactly this would work. (fine to figure it out as you implement)

Will we make a request from the UI itself across projects? (this can run into auth issues in some cases?)

or a sync API that given a projectID handles most of the logic for you. Will this be scoped to the team, or to the org?

requests-for-comments/2023-10-31-feature-flag-environments.md
jurajmajerik: thanks, I use [Excalidraw](https://excalidraw.com/) :)

requests-for-comments/2023-10-31-feature-flag-environments.md
liyiy: Yep reminder that while not really used right now, feature flags has special permissioning states

requests-for-comments/2023-10-31-feature-flag-environments.md
```diff
@@ -0,0 +1,72 @@
...
+#### Cons
+
+-   Most complex to implement
+-   Introduces another layer of complexity where it might not be needed (e.g. dashboards).
+
+Based on Annika's RFC and some preliminary discussion, I recommend implementing **Option a) Add the ability to copy a feature flag to another project.** This is the quickest win which addresses the users' requirements from the [RFC user feedback](https://github.com/PostHog/product-internal/blob/main/requests-for-comments/2022-12-08-feature-flag-environments.md#user-feedback).
+
+## Implementation
+
+In the Feature Flag view page, users will have the option to "Sync" the feature flag to other projects. To do so, they'll click the Sync button, after which a modal pops up. They can choose the project to sync the flag to via a dropdown.
```
liyiy, line 56: Will updating one feature flag automatically sync the other? Will this be a configuration option?

requests-for-comments/2023-10-31-feature-flag-environments.md
```diff
@@ -0,0 +1,72 @@
...
+#### Pros
+
+-   Cleanest separation of concerns, respects the distinction between projects and environments.
+
+#### Cons
+
+-   Most complex to implement
+-   Introduces another layer of complexity where it might not be needed (e.g. dashboards).
+
+Based on Annika's RFC and some preliminary discussion, I recommend implementing **Option a) Add the ability to copy a feature flag to another project.** This is the quickest win which addresses the users' requirements from the [RFC user feedback](https://github.com/PostHog/product-internal/blob/main/requests-for-comments/2022-12-08-feature-flag-environments.md#user-feedback).
```
liyiy, line 52: Are there any other good reasons to implement A besides quickest win? It's likely that users who set up projects like environments will still want environments even after being able to copy feature flags over more easily. I also think as we build out features to support option A, we'll find ourselves just basically building out what a full environments feature needs anyway 😆

requests-for-comments/2023-10-31-feature-flag-environments.md
liyiy: After they're copied over

jurajmajerik: @neilkakkar mentioned that users should be able to have a different configuration (e.g. targeting) for a staging flag than for a production one.

We can add the option to sync automatically, but I'd imagine that might complicate things for users (perhaps hide this behind "Advanced options").

neilkakkar: yeah I think "Sync" is the wrong word to use here, (and don't think we should build automatic sync _yet_ ). It's more like just "Copy"

requests-for-comments/2023-10-31-feature-flag-environments.md
jurajmajerik: The cons of other options are also the pros for option A :)

A few questions:
- In your view, what does building full environments encompass?
- What benefits does it provide compared to option A (syncing from one project to another)?
- Where else in the product might environments be needed?

requests-for-comments/2023-10-31-feature-flag-environments.md
jurajmajerik: We could show all projects in the sync dropdown but grey out the ones where the user isn't allowed to do operations.

How can check what permissions a particular user has for a given project?

requests-for-comments/2023-10-31-feature-flag-environments.md
jurajmajerik: good point, updated the RFC and the mocks to avoid confusion

annikaschmid (commented):
LGTM!

The only comment I have is that the word 'copy' out of context is not super clear to me, because it implies there would be a 'paste' button somewhere too. It's clearer in the popup though, so maybe that's fine. (Unless anyone can come up with a better name).

In terms of design, our top bar now becomes quite cluttered. Maybe it's time for an 'advanced settings' tab soon, or the ... we have on a dashboard for example.

jurajmajerik:
@annikaschmid I iterated on this a bit and moved everything under a new tab "Projects". This hopefully gives more context to the Copy button.

image

annikaschmid:
@jurajmajerik 👍 

I always try to avoid the yellow warnings, that always looks like something went wrong. Instead I would go with a simple grey information box, that says that if a flag with the same key exists in the target project already, those values will be overwritten.

Is there a way we can automatically detect if a flag exists already when the user selects the project and then change the "copy" button to "update"? That would be even better I think.

neilkakkar:
yes, and agree, great idea, we wouldn't need the warning when the button says copy/update depending on existence

jurajmajerik:
Great idea indeed! Just implemented it in https://github.com/PostHog/posthog/pull/18477

PR #149, opened 2023-11-02
Onboarding and G2 Review Request tweaks

joethreepwood:
Tweaking the emails to reflect two main changes. 

## Onboarding
We had feedback from a user that they were getting emails too regularly. I looked into this and saw the emails were indeed arriving sooner that we would like them to. So, I've tweaked the flow for the Cloud onboarding in two ways:

1. I've changed all 1 day waits to 2 day waits
2. I've moved all of these waits to occur _before_ we check user subscription status, instead of after.

Step 2. is important because there's a risk that users could subscribe to a product in this two day window, and then get the wrong email. 

The overall result is that it now takes 6 days, rather than 3, to move through the upsell cycle. 

If we felt strongly that emails are too frequent, we could additionally look at spacing out the newsletter invites or moving them to the end of the flow. 

## G2 Targeting
A suggestion from @andyvan-ph was that we tweak the G2 review requester to focus on engineers, in order to up the quality of our reviews. This roles that in. 

We had previously tried similar targeting aimed at users who were engineers AND high ICP users and it did not create good results, as it decimated our response rate. Before that we had tried targeting just engineers with a different incentive and that too killed our response rate -- but trying targeting engineers with the current incentive seems worth a shot.

### DIFF
--- a/requests-for-comments/2023-10-31-onboarding-emails.md
+++ b/requests-for-comments/2023-10-31-onboarding-emails.md
@@ -48,21 +48,21 @@ This campaign triggers when a user completes the `user signed up` event for the 
     6. If not `founder` `engineering` `product` `marketing` or `sales` then wait 24 hours
 4. Wait until a weekday between 1PM and 5PM in the users' time zone (UTC fallback)
 5. Send AARRR intro article email
-6. Wait 2 days.
+6. Wait 1 week.
 7. Wait until a weekday between 1PM and 5PM in the users' time zone (UTC fallback)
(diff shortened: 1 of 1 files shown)


charlescook-ph:
I think maybe we don't need consistent gaps between emails - for example upselling to a new product should occur after a week maybe? A lot of people will still be very much in tinkering around phase within a week I suspect. 

(But I'm curious what the data says - do you see people clicking through and then activating on those new products?)

Agree on the G2 recommendation 👍

charlescook-ph:
Sounds sensible!

PR #151, opened 2023-11-22
RFC: Scheduled Feature Flag changes

jurajmajerik:
We want to add the ability to schedule changes to feature flags. Please see this RFC for more details.

Related issue: https://github.com/PostHog/posthog/issues/13842

image

### DIFF
--- a/requests-for-comments/2023-11-22-scheduled-flag-changes.md
+++ b/requests-for-comments/2023-11-22-scheduled-flag-changes.md
@@ -33,13 +33,13 @@ POST /api/projects/:project_id/schedule
 }
 ```
 
-A scheduled change can be expressed as a JSON, specifying the operation to be performed and any payload required for the operation. In the context of scheduled changes to feature flags, the operation will be a `db_update`. For other scheduling use cases in the future, the operation may be again a database update or some other side effect.
+A scheduled change can be expressed as a JSON, specifying the operation to be performed and any payload required for the operation. In the context of scheduled changes to feature flags, the operation will be `add_release_condition` or `change_status`. For other scheduling use cases in the future, the operation may be similar updates to the model but could also be side effects.
 
 **Example 1: Enable the flag**
 
 ```
 {
-    "operation": "db_update",
+    "operation": "change_status",
     "payload": {
         "field": "is_active",
         "value": "false"
@@ -51,7 +51,7 @@ A scheduled change can be expressed as a JSON, specifying the operation to be pe
 
 ```
 {
-    "operation": "db_update",
+    "operation": "add_release_condition",
     "payload": {
         "field": “filter”,
         "value": {
@@ -76,9 +76,9 @@ We already use Celery with a Redis queue to execute various background tasks. Ho
 
 The `scheduled_changes` table would have the following schema:
 
-| team_id | record_id | model       | changes (JSON) | scheduled_at (timestamp) | executed_at (timestamp, null) |
-| ------- | --------- | ----------- | -------------- | ------------------------ | ----------------------------- |
-| 1       | 57        | FeatureFlag | [ ... ]        | 2023-11-23 15:30:00      | null                          |
+| team_id (int) | record_id (string) | model (string) | changes (JSON) | scheduled_at (timestamp) | executed_at (timestamp) | failure_reason (string, null) |
+| ------------- | ------------------ | -------------- | -------------- | ------------------------ | ----------------------- | ----------------------------- |
+| 1             | 57                 | FeatureFlag    | [ ... ]        | 2023-11-23 15:30:00      | null                    | null                          |
 
 The table would be indexed on the `scheduled_at` column to efficiently retrieve the columns that have the scheduled\*at value in the past. For every row retrieved, the task would **atomically**:
 
--- a/requests-for-comments/2023-11-22-scheduled-flag-changes.md
+++ b/requests-for-comments/2023-11-22-scheduled-flag-changes.md
@@ -110,6 +110,8 @@ Considering the above, our approach should be as follows:
 
 Every model that needs scheduled updates will implement a `scheduled_changes_dispatcher()` method. This method will receive the payload from the Celery task and update the record. After the successful update, the worker marks the row as completed by setting `executed_at`.
 
+Since this is sync and blocks the entire schedule queue, we can't have long running mutations with this system.
+
 ## Monitoring
 
 We will need to monitor the backlog of tasks and trigger an alert when the delay is causing too much degradation. For guidance, the `ingestion_lag` Celery task is a relevant example.
--- a/requests-for-comments/2023-11-22-scheduled-flag-changes.md
+++ b/requests-for-comments/2023-11-22-scheduled-flag-changes.md
@@ -1,0 +1,97 @@
+# Request for comments: Scheduled Feature Flag changes
+
+## Problem
+
+Currently, users can make updates to feature flags only manually and in real-time. This is not ideal, as it potentially leads to overlooked or forgotten changes, particularly when they fall outside normal working hours. This can lead to poorly timed releases, negatively impacting end users' engagement and feedback.
+
+## Solution
+
+Scheduling feature flag changes allows users to automatically change flag properties at a future point in time. For example, a user can schedule an initial release of the flag for internal testing, then enable it to the beta users group 2 days later, and finally roll it out to 100% of the users in 2 weeks.
+
+## Scope
+
+This feature should allow the following scheduled modification to a feature flag:
+
+-   **Enable/disable the flag** (field: `is_active`)
+-   **Add rollout conditions to the flag** (field: `filters`)
+
+TBD:
+
+-   **Modify existing rollout condition:** I would advise against adding this. When rolling out to more users, it is considered good practice to add to the existing conditions rather than modifying them. This provides an audit trail of changes and allows one to easily roll back the release to the previous state by removing the most recent condition(s).
+-   **Delete existing conditions:** This would allow temporary access to a feature. It might be helpful for some users, but I would advise to wait for more feedback before adding this.
+
+## Implementation
+
+We use Celery to execute tasks on schedule. At the predetermined time, Celery will pick the flag and evaluate scheduled changes. I considered two possible approaches to scheduling:
+
+-   Periodic scheduling - schedule every flag in regular intervals. This simplifies the implementation but creates a lot of unnecessary overhead.
+-   Dynamic scheduling - pick the earliest time from the scheduled changes and schedule the flag to be evaluated at this time. After performing the change, reschedule the flag to the new earliest scheduled time. For this RFC, I am assuming the use of dynamic scheduling.
+
+The scheduled changes can be stored on the flag record in the `scheduled_changes` field with the following structure:
+
+```
+[
+    {
+        "scheduled_at": "2023-11-22T12:30:00Z",
+        "type": "is_active",
+        "value": "false"
+    },
+    {
+        "scheduled_at": "2023-12-01T15:00:00Z",
+        "type": “filter”,
+        "value": {
+            "variant": null,
+            "properties": [{
+                "key": "$geoip_country_name",
+                "type": "person",
+                "value": ["Australia"],
+                "operator": "exact"
+            }],
+            "rollout_percentage": 20
+        }
+    }
+]
+```
+
+At the time specified, the Celery task fetches the flag, and iterates over the `scheduled_changes` items. For any item in this array, the `scheduled_at` field is evaluated, and if it is in the past, the task populates the item on the flag in the following way:
+
+-   When the type `is_active` the task sets the value of the `is_active` field.
+-   When the type is `filter`, the given condition is populated in the `filters` field of the flag.
+
+After populating the change on the feature flag, the change item is removed from the `scheduled_changes` array.
+
+To summarize, the end-to-end flow can go as follows:
+
+1. The user specifies the changes in the UI and saves the feature flag.
+2. The endpoint saves the record and stores a scheduled task in the Celery queue.
+3. Celery executes the scheduled task at the future date. It populates the required fields and reschedules the next task in the future.
+
+## UI
+
+We can add the functionality under the "Schedule" tab within the feature flag view. Users can submit scheduled changes via a form and view currently scheduled changes.
+
+<img width="981" alt="image" src="https://github.com/PostHog/posthog/assets/22996112/ff9f1c4c-c2f6-4dde-a420-bc282c5adcc8">
+
+## Generalizability
+
+The initial implementation idea assumes we schedule the task within the `POST /feature_flags` endpoint. However, we might want to abstract the scheduling functionality to accommodate other models such as `Survey` or `Experiment.` This would involve creating a scheduling API.
+
+```
+POST /projects/schedule
+{
+    type: 'feature_flag',
+    scheduled_changes: [ ... ]
+    scheduled_at: <timestamp>
+}
+```
+
+The API would populate the scheduled changes on the record and store a task in the Celery queue.
+
+I considered the option of only storing the scheduled changes in the task itself. This is, however, problematic since we need the list of changes available on the flag for the table view in the UI. That means this API would still need to both store the task and update the record with the list of scheduled changes.
+
+## Possible failure modes
+
+-   When submitting the scheduled changes to perform in the future, we need to store both the list of changes on the flag and the Celery task. The order matters here - we want to store the _task_ first. This is because:
+    -   If the flag is stored first, but the saving of the task fails, the list of changes is now saved on the flag without the associated task in the Celery queue.
+    -   If the task is stored successfully but the flag update fails, we inform the caller the operation failed. They can simply retry, so no issue here.
+-   When processing a scheduled updated of a flag, we also perform two actions - save the flag and complete the task. It is possible the task will fail after the record has been saved. Assuming Celery auto-reschedules failed tasks, the task would re-run and attempt to perform the changes again. At this point however, the scheduled change in question will have been removed from the `scheduled_changes` array. Thus, re-running the task would not lead to issues such as duplicate conditions being saved.
--- a/requests-for-comments/2023-11-22-scheduled-flag-changes.md
+++ b/requests-for-comments/2023-11-22-scheduled-flag-changes.md
@@ -72,7 +72,7 @@ A scheduled change can be expressed as a JSON, specifying the operation to be pe
 
 _At_ or _after_ the time specified, the change needs to be propagated onto the record.
 
-We already use Celery with a Redis queue to execute various background tasks. However, because the Redis queue is not durable, we cannot schedule tasks at the precise point in the future. A possible solution here is to have Postgres table of scheduled changes. Celery could periodically poll this table, e.g. every minute. During each poll, Celery would process any rows where the _scheduled_at_ time has already passed.
+We already use Celery with a Redis queue to execute various background tasks. However, because the Redis queue is not durable, we cannot schedule tasks at the precise point in the future. A possible solution here is to have Postgres table of scheduled changes. Celery could periodically poll this table, e.g. every minute. During each poll, Celery would process any rows where the _scheduled_at_ time has already passed and the operation hasn't been executed already.
 
 The `scheduled_changes` table would have the following schema:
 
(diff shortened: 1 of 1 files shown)


requests-for-comments/2023-11-22-scheduled-flag-changes.md
```diff
@@ -0,0 +1,97 @@
...
+                "value": ["Australia"],
+                "operator": "exact"
+            }],
+            "rollout_percentage": 20
+        }
+    }
+]
+```
+
+At the time specified, the Celery task fetches the flag, and iterates over the `scheduled_changes` items. For any item in this array, the `scheduled_at` field is evaluated, and if it is in the past, the task populates the item on the flag in the following way:
```
neilkakkar, line 56: This hides a lot of complexity, and I'm not sure it's possible the way we intend.

1. Our current celery backend is redis, which doesn't guarantee tasks will stick around
2. Celery workers will restart on every deploy
3. Celery recommends not scheduling tasks a long time in the future (like more than a few minutes): https://docs.celeryq.dev/en/stable/userguide/calling.html#eta-and-countdown
  and I'm not sure the alternative provided here will work in our case.
  
So, I'm leaning towards periodic checks to see if there's something to be scheduled every minute or so, this will have minimal overhead if we design things right (like a table with index on scheduled_time)

Alternatively, I think it's worth checking with @PostHog/team-pipeline what's the state of temporal, if it makes sense for this use case, & if we're planning to move everything to temporal at all or not. Not 100% clear of the capabilities there but if it's good enough for batch exports running at specific times, might be just what we need? or if it's way too heavy & overkill for what we need

requests-for-comments/2023-11-22-scheduled-flag-changes.md
```diff
@@ -0,0 +1,97 @@
...
+## Solution
+
+Scheduling feature flag changes allows users to automatically change flag properties at a future point in time. For example, a user can schedule an initial release of the flag for internal testing, then enable it to the beta users group 2 days later, and finally roll it out to 100% of the users in 2 weeks.
+
+## Scope
+
+This feature should allow the following scheduled modification to a feature flag:
+
+-   **Enable/disable the flag** (field: `is_active`)
+-   **Add rollout conditions to the flag** (field: `filters`)
```
neilkakkar, line 16: love this, excellent thinking!

requests-for-comments/2023-11-22-scheduled-flag-changes.md
```diff
@@ -0,0 +1,97 @@
...
+}
+```
+
+The API would populate the scheduled changes on the record and store a task in the Celery queue.
+
+I considered the option of only storing the scheduled changes in the task itself. This is, however, problematic since we need the list of changes available on the flag for the table view in the UI. That means this API would still need to both store the task and update the record with the list of scheduled changes.
+
+## Possible failure modes
+
+-   When submitting the scheduled changes to perform in the future, we need to store both the list of changes on the flag and the Celery task. The order matters here - we want to store the _task_ first. This is because:
```
neilkakkar, line 94: Having two sources of truth is indeed problematic, and I don't think we should go down this route.

We should use the scheduling only as a trigger, i.e. it's time to do something, and the 'something' has its own logic for figuring out what to do. More concretely this can look something like:

1. We have a table of all scheduled changes for a model, with the scheduled_at time, and a payload
2. We can use this table to show scheduled changes for a given model-item
3. Celery triggers a job every minute that checks for any model with scheduled_at <= now & if found, runs it & deletes / marks completed.
4. We use the db transaction guarantees to make sure it's executed by only one worker at a time.

I think it's worth generalising _right now_ rather than building something on the flag model itself.

The part I don't 100% like about this is we're building our own message queue, and all the issues that come with it. It's _fine if that's what works best_, but want us to consciously decide so.

neilkakkar (commented):
Nice, some things to sort out, but this looks promising!

requests-for-comments/2023-11-22-scheduled-flag-changes.md
xvello: Agreed, Celery is designed for repeated tasks than can be harmlessly skipped (as they'll be re-scheduled again). Temporal could indeed be a good scheduler for it, but might be a heavier lift.

Could you structure the database schema so that an idempotent "get transitions to execute and update the relevant stuff" transaction could be run? That transaction could then run every N minutes on Celery without reliability concerns. Happy to huddle on that.

neilkakkar: > Could you structure the database schema so that an idempotent "get transitions to execute and update the relevant stuff" transaction could be run? 

Yes, 100%! - I do currently like this the most as well.

xvello: FYI, pipeline is currently designing webhooks v2 that will be based on a custom Postgres task queue, so we'll be able to assist with concurrency when that becomes a scalability concern.

jurajmajerik:
Thanks @xvello, very helpful!
I was going to analyze the Temporal solution too, but because you both seem to suggest it’s an overkill, I focused only on refining the Celery solution. Ready for a review!

requests-for-comments/2023-11-22-scheduled-flag-changes.md
```diff
@@ -0,0 +1,84 @@
...
+}
+```
+
+_At_ or _after_ the time specified, the change needs to be propagated onto the record.
+
+We already use Celery with a Redis queue to execute various background tasks. However, because the Redis queue is not durable, we cannot schedule tasks at the precise point in the future. A possible solution here is to have Postgres table of scheduled changes. Celery could periodically poll this table, e.g. every minute. During each poll, Celery would process any rows where the _scheduled_at_ time has already passed.
+
+The `scheduled_flag_changes` table would have the following schema:
+
+| team_id | id  | changes (JSON) | scheduled_at (timestamp) | status (enum) |
```
xvello, line 69: `id` will be the auto-incremented row ID. From your use case, I'm assuming that you'll want to show these schedules in the feature-flag page, hence another `feature_flag_id` foreign key and index

requests-for-comments/2023-11-22-scheduled-flag-changes.md
```diff
@@ -0,0 +1,84 @@
...
+            "type": "person",
+            "value": ["Australia"],
+            "operator": "exact"
+        }],
+        "rollout_percentage": 20
+    }
+}
+```
+
+_At_ or _after_ the time specified, the change needs to be propagated onto the record.
```
xvello, line 63: Indeed, there can be a backlog of tasks, and you need to be able to measure it and alert on it if the delay is causing too much degradation to the experience.
Easy to do by replicating the `ingestion_lag` celery task.

requests-for-comments/2023-11-22-scheduled-flag-changes.md
xvello, line 69: If you plan on having only two states for `status` (waiting, done), I'd recommend swapping it for an `executed_at` nullable column. This way, you can compute statistics about how many tasks have been executed within proper timing. That would be your [SLI](https://www.atlassian.com/incident-management/kpis/sla-vs-slo-vs-sli) for that feature.

You'll want to remove old rows eventually. What retention would be OK for customers?

requests-for-comments/2023-11-22-scheduled-flag-changes.md
```diff
@@ -0,0 +1,84 @@
...
+| team_id | id  | changes (JSON) | scheduled_at (timestamp) | status (enum) |
+| ------- | --- | -------------- | ------------------------ | ------------- |
+| 1       | 57  | [ ... ]        | 2023-11-23 15:30:00      | scheduled     |
+
+The table would be indexed on the `scheduled_at` column to efficiently retrieve the columns that have the scheduled\*at value in the past. For every row retrieved, the task would **atomically**:
+
+-   Modify the field on the feature flag record
+-   Mark the row as `completed`
+
+We would need to have separate tables for other models such as `scheduled_survey_changes`, due to the schema differences and possibly different scaling requirements.
```
xvello, line 78: Agreed. Both celery tasks can reuse a lot of the queuing logic though

neilkakkar: I'm not clear why we'll need separate tables for each model, why can't / shouldn't they be all in one model?

Scaling imo is not a concern here. Given we have [not that many](http://metabase-prod-us/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7InF1ZXJ5Ijp7InNvdXJjZS10YWJsZSI6MTQ4LCJhZ2dyZWdhdGlvbiI6W1siY291bnQiXV19LCJ0eXBlIjoicXVlcnkiLCJkYXRhYmFzZSI6MzR9LCJkaXNwbGF5IjoidGFibGUiLCJ2aXN1YWxpemF0aW9uX3NldHRpbmdzIjp7fX0=) flags, assuming 1 thing per flag, all scheduled at the same time, that's 20k items to operate on today, and assuming 10x growth, 200k in the next 2 years. [and flags are biggest w.r.t team feature success right now].

Since we're inputting arbitrary json for the operation to do, schema differences also don't mean much to us, the models themselves determine how to consume the json and do the operation, while the scheduler just has the scheduler logic + model + payload (json).

requests-for-comments/2023-11-22-scheduled-flag-changes.md
```diff
@@ -0,0 +1,84 @@
...
+    "scheduled_at": <timestamp>
+}
+```
+
+A change to the record can be expressed as a JSON, specifying the field to be modified and the new value to be set:
+
+**Example 1: Enable the flag**
+
+```
+{
```
neilkakkar, line 39: I think it's worth keeping this more generalisable: let the models determine what to do with the payload, and the scheduler just blindly passes the json.

So, I'd propose a top-level in this json to be two things: "operation" & "payload", and inside "payload" is what you have here. "operation" could be "database_update" for example.

This makes it much easier to extend to other scheduled things that aren't necessarily represented by a model update. And if we don't extend it at all, it's still fine, we're not adding massive generalising complexity here.

requests-for-comments/2023-11-22-scheduled-flag-changes.md
jurajmajerik: ah yes, what I meant here was indeed `feature_flag_id`, or `record_id`

requests-for-comments/2023-11-22-scheduled-flag-changes.md
neilkakkar: I guess the one disadvantage is not having foreign keys / cascade deletes functionality via the db if it's all in one table, but reasonable trade-off, given we can easily handle deleted flags / any inconsistencies.

Although, worth considering if we want a failure status at all? I can't yet think of how we'll end up in such a state, given other constraints, so maybe ok to defer till we find such a case

requests-for-comments/2023-11-22-scheduled-flag-changes.md
jurajmajerik: - Makes sense. Initially, I thought about three states (waiting, claimed, done), but since the operation is idempotent, two states will suffice. No need to claim the row.

- Yes, good point; I'd set it arbitrarily to 90 days and tweak it later as needed.

xvello: > No need to claim the row (and having to figure out what happens when the worker that claimed the row fails).

Indeed, no need to claim it if concurrency = 1, then for higher concurrency I recommend using a `select for update skip locked` approach that does not mutate the row, using transactions instead.

neilkakkar: oh nice, I didn't know there's an easy way to force concurrency with celery, I assumed given 6 pods running celery workers, all talking to the same db, we'll by default have higher concurrency, but if celery can manage this itself, works 👍

jurajmajerik: Okay, is my understanding correct that we want to have concurrency = 1 for the current/foreseeable load?

xvello: > oh nice, I didn't know there's an easy way to force concurrency with celery

I don't have much experience with Celery, so I am probably missing a big caveat here. Can you please check that? It might require using a dedicated task queue (I think it would be beneficial to split our big celery task corpus into separate queue executed by the same workers).

If that's not trivial, we can solve concurrency at the PG level when implementing the task.

neilkakkar: .. looking at celery docs, I'm not sure where this is configured: https://docs.celeryq.dev/en/latest/userguide/workers.html#concurrency, our current settings seem to suggest we are using the defaults, which means # of cores workers, multiplied by 6 for the different pods running the workers.

I don't think we'll change the defaults, as that affects all tasks, but is there a way to enforce concurrency for this specific task?

If not, I think we are already in the `select for update skip locked` approach land

requests-for-comments/2023-11-22-scheduled-flag-changes.md
jurajmajerik: Right, so the method for updating the record will live in the model. The Celery task will pick the appropriate model and pass the JSON to the model's method. Correct?

Curious, what's an example of scheduling where we're not performing a database update?

neilkakkar: >  The Celery task will pick the appropriate model and pass the JSON to the model's method. Correct?

Yep, one slight change: pass it to a function on the model, let's say called `scheduled_changes_dispatcher()`, and every model that needs scheduled tasks implements this function. And then the scheduler doesn't care what happens inside, be it a model update, or something else.

I think most things would do _some_ database update, but it might not be the only thing. For random example: a scheduled cohort computation: it may or may not update the database, but all it needs to do is recalculate the cohort. A bit contrived example, but hopefully gets the point across. It becomes much harder to do side-effects for example, if the only thing the scheduled task can do is perform db updates.

requests-for-comments/2023-11-22-scheduled-flag-changes.md
neilkakkar: no, don't think so. 👍 

---

Although, failure reason does make more sense now, thinking about this, because say for example the db update is invalid (does not pass validation filters for example) - we want to tell the user the scheduled update failed for whatever reason.

While the frontend should make sure users can't input invalid stuff, API level updates can bypass this, so our core validation should be on the backend

requests-for-comments/2023-11-22-scheduled-flag-changes.md
jurajmajerik: If we can use the `skip locked` query here, I don't see any benefit in limiting concurrency to 1. It will be more scalable to have concurrency = N from the start.

xvello: with concurrency you'll need to make sure you don't run concurrent mutations to the same flag, and error handing will get a bit trickier. I think we're talking about implementation details while I don't know your design constraints:

- what delay in mutations is acceptable
- can mutation fails? do they need to be retried?
- do you need to guard against concurrent mutations on the same feature flag (could happen after a long celery incident, this might mess the final state).

I need to focus on another topic right now, but I suggest you think more about these constraints, then we can schedule some sync time to discuss the implementation details. If you want retries and concurrency, temporal might be a simpler way forward than rewriting our own queue implementation.

requests-for-comments/2023-11-22-scheduled-flag-changes.md
jurajmajerik: Makes sense. Another example of a side effect might be calling a 3rd party API?

One more thing - can I expect every scheduling scenario to be related to _some_ model?

neilkakkar: yes that makes sense

requests-for-comments/2023-11-22-scheduled-flag-changes.md
neilkakkar: great points. As a side note, I think we can still enforce concurrency=1 at db instead of celery level and skip the above problems by not adding 'skip locked' to the select, and wrapping it in a transaction. It's another option to consider if solving the above becomes too unwieldy.

requests-for-comments/2023-11-22-scheduled-flag-changes.md
jurajmajerik: One more thing regarding updating the record and the queue item. To reiterate, we want to _atomically_:
- Update the feature flag
- Mark the queue item as done (by setting the `executed_at` timestamp)

However, if the responsibility of updating the record is now on the `scheduled_changes_dispatcher()` method, this method must handle the transaction. Does it make sense that the queue item's update be managed by the model rather than the Celery task?

neilkakkar: > Does it make sense that the queue item's update be managed by the model rather than the Celery task?

No this is no good! All scheduling logic & scheduler table updates should be in the task itself. We lose the general isolation if `scheduled_changes_dispatcher` implementation needs to handle this.

> the scheduled_changes_dispatcher() method, this method must handle the transaction.
This isn't necessary I think? The celery task can start a transaction, and whatever happens inside `scheduled_changes_dispatcher()` is a part of the transaction, without it handling the transaction itself. For an example, see this: https://github.com/PostHog/posthog/blob/master/posthog/models/feature_flag/flag_matching.py#L726 - it's a context manager that starts a transaction, does some updates, without caring what happens inside.

Given we don't really support long running tasks here, I think it's okay to have this all in one transaction. It's a bit iffy wr.t. side effects, but I think an ok trade off to make, given the kind of things we want to schedule here are limited & aren't building a general scheduler for everything.

jurajmajerik: great, makes sense, exactly what I'm looking for!

jurajmajerik:
Thanks for the comments @xvello, @neilkakkar. Reworked the RFC, mainly:
- Made the API schema more generalizable
- Now assuming a single table for all scheduled changes
- Analyzed concurrency/retries requirements and added a possible solution for implementing concurrency=1 with the NOWAIT clause

requests-for-comments/2023-11-22-scheduled-flag-changes.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Queue processing
+
+_At_ or _after_ the time specified, the change needs to be propagated onto the record.
+
+We already use Celery with a Redis queue to execute various background tasks. However, because the Redis queue is not durable, we cannot schedule tasks at the precise point in the future. A possible solution here is to have Postgres table of scheduled changes. Celery could periodically poll this table, e.g. every minute. During each poll, Celery would process any rows where the _scheduled_at_ time has already passed.
+
+The `scheduled_changes` table would have the following schema:
+
+| team_id | record_id | model       | changes (JSON) | scheduled_at (timestamp) | executed_at (timestamp, null) |
```
neilkakkar, line 79: implementation reminder that record_id needs to be string to support both int & string model ids

neilkakkar: below you mention recording failure, that needs to go in the model too!

requests-for-comments/2023-11-22-scheduled-flag-changes.md
```diff
@@ -0,0 +1,121 @@
...
+-   PG is down - need to retry.
+
+Considering the above, our approach should be as follows:
+
+-   Retries: If PG downtime is the only reason for a failed mutation, we can simply wait for the next scheduled run of Celery at which point the schedule change will be retried.
+-   Concurrency: Given that our current/foreseeable count of records is low, we can afford concurrency=1. Even during peak loads, we should be able to handle all scheduled changes within 1 minute. To implement concurrency=1, we can use a transaction with the NOWAIT clause. The first worker that selects the rows "wins", while the queries from subsequent workers will fail with an error.
+
+### Model updates
+
+Every model that needs scheduled updates will implement a `scheduled_changes_dispatcher()` method. This method will receive the payload from the Celery task and update the record. After the successful update, the worker marks the row as completed by setting `executed_at`.
```
neilkakkar, line 111:
```suggestion
Every model that needs scheduled updates will implement a `scheduled_changes_dispatcher()` method. This method will receive the payload from the Celery task and update the record. After the successful update, the worker marks the row as completed by setting `executed_at`.

Since this is sync and blocks the entire schedule queue, we can't have long running mutations with this system.
```

neilkakkar (approved):

PR #167, opened 2023-12-17
RFC: Some thoughts about community in 2024

andyvan-ph:
I've been pondering community for the last week or so and put some thoughts to paper.

**Date for decision:** Jan 2024

### DIFF
--- a/requests-for-comments/commnunity-2024.md
+++ b/requests-for-comments/commnunity-2024.md
@@ -1,0 +1,47 @@
+Up to now, much of our thinking for the community section of the website has centered on creating a Hacker News-like experience where content drives the community. I think this is the wrong direction. This issue outlines why and an alternative vision of the community.
+
+## Why not the current direction?
+
+Content, both our own and good content curated _by us_ can still form an important part of our community, but:
+
+1. There are already numerous places to share content about building startups / engineering topics generally. Why should people post it to ours and not another, more established channel?  
+
+2. The attraction of our content comes from our strong POV and curation. Outsourcing curation to the community undermines a strength we already have.
+
+3. If it were successful, it would create a significant moderation challenge that would suck up a bunch of time. Abuse and self-promotion will happen and we'll have to manage that.
+
+4. There's a significant risk our own content would be drowned out by community contributions and submissions to the point.
+
+5. It doesn't create any kind of stickiness for our product.
+
+## An alternative vision – HogBook
+
+I think the best way to create a thriving community on the website is to focus it around the product. This means:
+
+- Making it easy to request new features and vote on what the teams are working on
+- Making it easy to see what teams are working on and their progress
+- Creating a direct connection between all users and our engineers (not just the teams who pay for a dedicated Slack channel)
+- Creating a feedback loop where users request / vote on a feature then get notified when it's shipped
+- Creating a space where users can share feedback in public, rather than it being hidden in Zendesk / Slack channels.
+
+To this end, I think the community homepage should be more like a Facebook / Twitter-style feed experience comprising:
+
+- Content we've published that's shared by the author (rather than content automatically appearing when it's published, the author would share it like they would on a social network).
+- Feature requests from users and proposals use that people can comment and vote on – much like they currently do via the Roadmap + GitHub.
+- Semi-regular dev diaries / updates from individual teams documenting their progress.
+- Dedicated team pages (effectively like Facebook groups but for our teams)
+
+In this scenario, the sidebar would look something like this:
+
+<img width="684" alt="Screenshot 2023-12-17 at 11 03 14" src="https://github.com/PostHog/posthog.com/assets/92976667/b6f3b6fa-b5b3-468b-ad9f-bd3c260f51d3">
+
+- The team pages would include feature requests, roadmaps, dev diaries from the team, and content relevant to that product.
+- The community homepage would then combine all this into a single feed that combines the content from the team pages and content hubs.
+
+### Random notes:
+
+- Things like customer stories and comparisons etc. wouldn't appear here as they're largely marketing material not intended for the community.
+
+- Would be good if there were more reactions than just upvoting – something closer to Slack-style reactions.
+
+- This would exist somewhat separate to the Questions forum. TL;DR: feature requests and feedback would go in the team pages. Support questions / problems remain in questions.


requests-for-comments/commnunity-2024.md
```diff
@@ -0,0 +1,47 @@
+Up to now, much of our thinking for the community section of the website has centered on creating a Hacker News-like experience where content drives the community. I think this is the wrong direction. This issue outlines why and an alternative vision of the community.
+
+## Why not the current direction?
+
+Content, both our own and good content curated _by us_ can still form an important part of our community, but:
+
+1. There are already numerous places to share content about building startups / engineering topics generally. Why should people post it to ours and not another, more established channel?
```
Lior539, line 7: Not sure I agree with this. 
1. There are many for engineers, but I dont think any have a _product engineer_ focus
2. For building startups, which places exist to share content?

requests-for-comments/commnunity-2024.md
```diff
@@ -0,0 +1,47 @@
+Up to now, much of our thinking for the community section of the website has centered on creating a Hacker News-like experience where content drives the community. I think this is the wrong direction. This issue outlines why and an alternative vision of the community.
+
+## Why not the current direction?
+
+Content, both our own and good content curated _by us_ can still form an important part of our community, but:
+
+1. There are already numerous places to share content about building startups / engineering topics generally. Why should people post it to ours and not another, more established channel?  
+
+2. The attraction of our content comes from our strong POV and curation. Outsourcing curation to the community undermines a strength we already have.
```
Lior539, line 9: This feels like a solvable probably. You could, say, introduce HN style upvoting, so that higher quality ones make it to the top

requests-for-comments/commnunity-2024.md
```diff
@@ -0,0 +1,47 @@
...
+
+3. If it were successful, it would create a significant moderation challenge that would suck up a bunch of time. Abuse and self-promotion will happen and we'll have to manage that.
```
Lior539, line 11: This would be a good problem to have, since it means we _are_ successful 😃

requests-for-comments/commnunity-2024.md
```diff
@@ -0,0 +1,47 @@
...
+
+4. There's a significant risk our own content would be drowned out by community contributions and submissions to the point.
```
Lior539, line 13: Again, solvable problem. We could pin ours to the top etc.

requests-for-comments/commnunity-2024.md
```diff
@@ -0,0 +1,47 @@
...
+
+## An alternative vision – HogBook
+
+I think the best way to create a thriving community on the website is to focus it around the product. This means:
+
+- Making it easy to request new features and vote on what the teams are working on
+- Making it easy to see what teams are working on and their progress
+- Creating a direct connection between all users and our engineers (not just the teams who pay for a dedicated Slack channel)
+- Creating a feedback loop where users request / vote on a feature then get notified when it's shipped
+- Creating a space where users can share feedback in public, rather than it being hidden in Zendesk / Slack channels.
```
Lior539, line 25: We do this already to some extent at a company level, but I do like this idea of taking it a step further to a team level. It will definitely foster creating direct connections with individuals on PostHog, and ultimately be a good user experience for our users.

requests-for-comments/commnunity-2024.md
```diff
@@ -0,0 +1,47 @@
...
+- The team pages would include feature requests, roadmaps, dev diaries from the team, and content relevant to that product.
+- The community homepage would then combine all this into a single feed that combines the content from the team pages and content hubs.
+
+### Random notes:
+
+- Things like customer stories and comparisons etc. wouldn't appear here as they're largely marketing material not intended for the community.
+
+- Would be good if there were more reactions than just upvoting – something closer to Slack-style reactions.
+
+- This would exist somewhat separate to the Questions forum. TL;DR: feature requests and feedback would go in the team pages. Support questions / problems remain in questions.
```
Lior539, line 47: Commenting my overall thoughts on this RFC on this last line (so it's easier to reply to):

Overall:
1. I think HogBook is a great idea, and we should do it.
2. I don't think Hogbook and our current direction are mutually exclusive – we can do both.

I see Hogbook as community building for PostHog users, and our current direction as community building for non-PostHog users. If we want to focus on community building for only one of these, we should be clear which one it is and why.

From a marketing perspective, focusing on PostHog users will help with word-of-mouth growth, whereas focusing on non-PostHog users will also help growth. However, I suspect focusing on non-PostHog users would help us grow more, but this is just based on intuition from the top of my head

requests-for-comments/commnunity-2024.md
benjackwhite, line 25: This feels much more like what I always thought our community focus was - a community around PostHog and working with it.

In terms of biasing for impact, I feel like this is where our current community wants us to be and we sometimes aren't (e.g. split focus with slack community). 

There is sometimes seemingly an appetite for general discussion a la hackernews, but it feels like the edge case rather than the majority of the pie. If we had a community that was _genuinely thriving_ around PostHog, we would see if the other ideas of being a generic product/startup/engineering community naturally form without directly investing in it.

requests-for-comments/commnunity-2024.md
```diff
@@ -0,0 +1,47 @@
...
+- Content we've published that's shared by the author (rather than content automatically appearing when it's published, the author would share it like they would on a social network).
+- Feature requests from users and proposals use that people can comment and vote on – much like they currently do via the Roadmap + GitHub.
+- Semi-regular dev diaries / updates from individual teams documenting their progress.
+- Dedicated team pages (effectively like Facebook groups but for our teams)
+
+In this scenario, the sidebar would look something like this:
+
+<img width="684" alt="Screenshot 2023-12-17 at 11 03 14" src="https://github.com/PostHog/posthog.com/assets/92976667/b6f3b6fa-b5b3-468b-ad9f-bd3c260f51d3">
+
+- The team pages would include feature requests, roadmaps, dev diaries from the team, and content relevant to that product.
```
benjackwhite, line 38: If our roadmap was powered by something like Github's product boards, I feel like the adoption among teams would also be way higher (it goes out of date pretty quick atm)

requests-for-comments/commnunity-2024.md
andyvan-ph: 1. I think that's a narrow distinction we care about more than most people do. 
2. Numerous sub-reddits, Slack groups, Discords, Indiehackers. That said, most tend to discourage sharing "content" because it's just an invitation to spam them.

requests-for-comments/commnunity-2024.md
andyvan-ph: We already have introduced HN-style upvoting that no one is using atm.

requests-for-comments/commnunity-2024.md
andyvan-ph: > If we had a community that was _genuinely thriving_ around PostHog, we would see if the other ideas of being a generic product/startup/engineering community naturally form without directly investing in it.

Very good point. If we did want to return to the content / Hacker News concept, it would have a much greater chance of success if we had people bought into the community already.

joethreepwood (commented):
I have three main concerns. 

**1.** The primary drivers for the existing vision are @jamesefhawkins and @corywatilo. I think we need a community vision RFC laid out by James first before we make any decision.

**2.** Nobody has experience building communities for either vision. We had an SEO consultant in when we started thinking about content, we're chatting to experts in billboards at the moment...we need an equivalent for community.

**3.**  Neither vision asks what our users want and we've seen no evidence they want either. Upvotes aren't used, users never share their own content, we barely get any roadmap comments, and even when we put roadmap subscribers into Slack channels with small teams they never said anything.

Even in the YC-Founder channel where I make a dedicated effort to regularly share relevant content with engaged founders we _still_ only get support requests. Nobody asks about betas, best practices, etc. Just support. 

IMHO we need to examine if either vision gives users what _they_ want. Right now, we may be best off focusing on https://github.com/PostHog/product-internal/pull/546 instead.

requests-for-comments/commnunity-2024.md
corywatilo: Cart before the horse.

requests-for-comments/commnunity-2024.md
corywatilo: Solvable with weighting

requests-for-comments/commnunity-2024.md
```diff
@@ -0,0 +1,47 @@
...
+
+I think the best way to create a thriving community on the website is to focus it around the product. This means:
+
+- Making it easy to request new features and vote on what the teams are working on
+- Making it easy to see what teams are working on and their progress
+- Creating a direct connection between all users and our engineers (not just the teams who pay for a dedicated Slack channel)
+- Creating a feedback loop where users request / vote on a feature then get notified when it's shipped
+- Creating a space where users can share feedback in public, rather than it being hidden in Zendesk / Slack channels.
+
+To this end, I think the community homepage should be more like a Facebook / Twitter-style feed experience comprising:
```
corywatilo, line 27: Funny - I mocked something like this last week:

image

andyvan-ph: Yeah, like that. 😅

ivanagas:
Basically agree with Joe

requests-for-comments/commnunity-2024.md
andyvan-ph: "Liking" feels less loaded than "Upvoting" – definitely a step in the right direction.

charlescook-ph:
Hello gang, there’s been a lot of chat around community recently! We just had a chat as part of the website & docs quarterly planning meeting, and it’s clear that both the vision for community and who is actually responsible for 'doing the things' is drifting between W&D and Marketing, and this is causing some confusion and frustration for people because we're all on slightly different pages. 

We also want to reduce dependencies between small teams, and having W&D build something which Marketing is then responsible for maintaining is not really in the spirit of this. 

Solution: from Q1, getting community at PostHog to product-market fit will be _a W&D responsibility solely_. This means that they will come up with the vision, build the tools and then do the stuff to make it successful (e.g. update the roadmap, make hubs successful, maybe other things we haven’t thought of yet). @jamesefhawkins is going to be our community person as we try and make this work. Depending on what lands and what doesn’t, we will then come up with a longer term plan for how to manage community (e.g. hiring someone).

For the marketing team, this means that building community/making posts successful is _not_ a goal for us in Q1, and we will focus on all the other things on our list. (Obv we can define the newsletter and/or YC stuff as building community if you want to nitpick, but I hope the spirit of what I’m trying to say here comes across…)

PR #168, opened 2023-12-18
RFC - Lior's 2023Q4 content results

Lior539:
In preparation for Tuesday's Q1 planning, here are the results for the content I wrote in Q4


charlescook-ph:
1. This is super! I think this is a great template/example for others to follow. 
2. Do you have links to the insights you're using? (Just to make sure we're all using the same definition of 'user signed up.') Further down the line - probably ~2 months after signup - it may be worth seeing how many of these convert to a paying customer - I use [this action](https://us.posthog.com/data-management/actions/31956) which gives you invoice paid or projected to be paid.

Lior539:
> Do you have links to the insights you're using?

andyvan-ph (approved):

charlescook-ph:
> Here's a link to the dashboard (it's in the RFC too): https://us.posthog.com/dashboard/119652
> I copied the actions @andyvan-ph uses :)

facepalm.jpg

PR #176, opened 2024-01-23
RFC: First-class environments

Twixes:
We've been putting this off for a long time.

Related issues:
- https://github.com/PostHog/posthog/issues/19476

### DIFF
--- a/requests-for-comments/2024-01-23-project-environments.md
+++ b/requests-for-comments/2024-01-23-project-environments.md
@@ -49,6 +49,10 @@ The above must be filtered by environment at all times. In addition, there are c
 
 The "Feature flag environments" RFC brought up one key need users have: often, the test setup needs some deviations from prod. You're likely to play around with instrumentation during development, so feature flags or cohorts might deviate in dev and you probably don't want that setup to be the source of truth. It's the production setup which you want to be a stable source of truth.
 
+### Previous attempts
+
+There's been one attempt at pretty much this, a couple of years ago. It was a test mode switch, meaning essentially two environments. Unfortunately, I don't think there was a spec or RFC for that feature, and we also never rolled it out. Traces of the feature have since been removed from the code.
+
 [TODO: Research other product analytics tools]
 
 ## Design 
--- a/requests-for-comments/2024-01-23-project-environments.md
+++ b/requests-for-comments/2024-01-23-project-environments.md
@@ -133,6 +133,8 @@ To support the most common workflow of drafting dashboards/etc. in dev/staging a
 
 To support large customers with access control needs, we'd need to add environment-specific project membership to the existing project access control system.
 
+To support the common feature flags workflow of rollout conditions being different between dev and production, we'd make flag rollout groups always environment-specific.
+
 To ensure straightforward yet complete separation of data at the ingestion level (incl. persons), *as well as* steady querying performance, each environment would have its own `team_id`. Due to this requirement, environments would have to be predefined by users and not created at ingestion time, thus each environment would also have its own SDK key (aka "project API key" currently). That seems like a safe choice in terms of the UX.
 
 By default, a project would only have one environment implicitly named "Production". A "New environment" button would allow adding a new one, with the environments feature explained in the modal. The current environment would then be selectable from the project selector, roughly looking like this (courtesy of Cory):
@@ -175,5 +177,7 @@ Of course, nothing comes for free: adding evironments has its own downsides. Her
 
   > The transition period ends at this step. We should no longer be using `posthog_team.api_token` anywhere anymore.
 
+1. Backend + frontend, behind a flag: [UX to be specified] We allow feature flag rollout groups to be environment-specific.
 1. Frontend, behind a flag: To prevent the UI from being overwhelming, we turn the organization name into just an icon. If the organization owner has an email address from a non-email-provider domain, we use that domains favicon. Otherwise we show a lettermark (no image uploading for now).
-1. Backend: + frontend, behind a flag: [Underspecified] We add environment-level permissions to the `ExplicitTeamMembership` system.
+  > At this point we can likely release the feature and unflag it.
+1. Backend + frontend: [UX to be specified] We add environment-level permissions to the `ExplicitTeamMembership` system.
--- a/requests-for-comments/2024-01-23-project-environments.md
+++ b/requests-for-comments/2024-01-23-project-environments.md
@@ -165,8 +165,8 @@ To optimize the implementation phase, we can cut opt-in project-wideness out of 
 1. Backend: We add a `Project` model (`posthog_project` table, initially integer column `id` + string column `name`). Meanwhile, `posthog_team` gets one new nullable column: `project_id`.
   From now on `posthog_team` means "environment" internally. For ease of transition, the first environment in a project always has the same ID as the project.  
   At this point in implementation, new-style projects are created automatically for every new environment (i.e. old-style project), but cannot be created manually.
-1. Backend: For every Team which isn't under a Project yet, we create a matching default Project. After this, every single environment has a parent project, thus we make `project_id` non-nullable.
-1. Backend: For all project-scoped viewsets, we change the queryset filtering from `team_id=self.team_id` to `team__project_id=self.team_id`. See list below for models associated with such endpoints:
+2. Backend: For every Team which isn't under a Project yet, we create a matching default Project. After this, every single environment has a parent project, thus we make `project_id` non-nullable.
+3. Backend: For all project-scoped viewsets, we change the queryset filtering from `team_id=self.team_id` to `team__project_id=self.team_id`. See list below for models associated with such endpoints:
     - `Dashboard`
     - `Insight`
     - `Action`
@@ -178,20 +178,21 @@ To optimize the implementation phase, we can cut opt-in project-wideness out of 
     - `SessionRecordingPlaylist`
    
     This way we're able to postpone denormalizing `project_id`.
-1. We move `TeamViewSet` from `/api/projects/` to `/api/environments/`. At `/api/projects/` we instead put a new `ProjectViewSet`, which is based on the `Project` model BUT behaves exactly like `TeamViewSet`. This makes the transition invisible to existing API consumers, and we can do it thanks to the first environment's ID always being the same as the project's ID. As for settings, they stay at the `posthog_team` level, but we can work around this by still returning and updating settings at `/api/projects/`, just always using the first environment. This ensures backward compatibility, even if it's sneaky.  
+4. We move `TeamViewSet` from `/api/projects/` to `/api/environments/`. At `/api/projects/` we instead put a new `ProjectViewSet`, which is based on the `Project` model BUT behaves exactly like `TeamViewSet`. This makes the transition invisible to existing API consumers, and we can do it thanks to the first environment's ID always being the same as the project's ID. As for settings, they stay at the `posthog_team` level, but we can work around this by still returning and updating settings at `/api/projects/`, just always using the first environment. This ensures backward compatibility, even if it's sneaky.  
   For environment creation specifically, we add `/api/projects/:project_id/environments/`.
-1. Frontend, behind a flag: We turn the project selector into the environment selector. This means sectioning the selector by project, renaming "New project" to "New environment" and scoping that flow by project, and finally adding a "New project" button to allow *real Project* creation
+5. Frontend, behind a flag: We turn the project selector into the environment selector. This means sectioning the selector by project, renaming "New project" to "New environment" and scoping that flow by project, and finally adding a "New project" button to allow *real Project* creation
     > During project creation, we must make it crystal clear for customers splitting by region that cross-environment analytics isn't possible – environments are designed specifically to solve data cross-contamination.
-1. Backend + frontend, behind a flag: We add nullable field `project_id` to the `GroupTypeMapping` (along with a constraint to ensure a group type index is not reused within a `posthog_project`), backfill the field in the same migration, and then make it non-nullable. We reuse `ClickhouseGroupsTypesView` for `/api/v2/projects/:project_id/group_types/` and – in the flagged UI – switch to this endpoint from the environment-level version. (Zero real impact currently, as no user has multiple environments within a project *yet*.)
-1. Plugin server: We add the `posthog_team.project_id` field to `TeamManager` and subsequently update `GroupTypeManager` to map group types per `project_id` rather than `team_id`.
-1. Backend + frontend, behind a flag: We add nullable field `project_id` to the `EventDefinition` & `PropertyDefinition`, backfill the field in the same migration, and then make it non-nullable. We reuse `EventDefinitionViewSet` & `PropertyDefinitionViewSet` for `/api/v2/projects/:project_id/{event,property}_definitions/` and – in the flagged UI – switch to these endpoints from the environment-level versions.
-1. Plugin server: We update `PropertyDefinitionsManager` to map group types per `project_id` rather than `team_id`.environments.
-1. Frontend, behind a flag: We move the current "Project settings" to a new "Environment settings" section, and add the new-style project name to "Project setting".
-1. Backend + frontend, behind a flag: [UX to be specified] We allow feature flag rollout groups to be either project-wide or environment-specific.
-1. Frontend, behind a flag: To prevent the UI from being overwhelming, we turn the organization name into just an icon. If the organization owner has an email address from a non-email-provider domain, we use that domain's favicon. Otherwise, we show a lettermark (no image uploading for now).
+6. Backend + frontend, behind a flag: We add nullable field `project_id` to the `GroupTypeMapping` (along with a constraint to ensure a group type index is not reused within a `posthog_project`), backfill the field in the same migration, and then make it non-nullable. We reuse `ClickhouseGroupsTypesView` for `/api/v2/projects/:project_id/group_types/` and – in the flagged UI – switch to this endpoint from the environment-level version. (Zero real impact currently, as no user has multiple environments within a project *yet*.)
+7. Plugin server: We add the `posthog_team.project_id` field to `TeamManager` and subsequently update `GroupTypeManager` to map group types per `project_id` rather than `team_id`.
+8. Backend + frontend, behind a flag: We add nullable field `project_id` to the `EventDefinition` & `PropertyDefinition`, backfill the field in the same migration, and then make it non-nullable. We reuse `EventDefinitionViewSet` & `PropertyDefinitionViewSet` for `/api/v2/projects/:project_id/{event,property}_definitions/` and – in the flagged UI – switch to these endpoints from the environment-level versions.
+9. Plugin server: We update `PropertyDefinitionsManager` to map group types per `project_id` rather than `team_id`.environments.
+10. Frontend, behind a flag: We move the current "Project settings" to a new "Environment settings" section, and add the new-style project name to "Project setting".
+11. Backend + frontend, behind a flag: [UX to be specified] We allow feature flag rollout groups to be either project-wide or environment-specific.
+12. Frontend, behind a flag: To prevent the UI from being overwhelming, we turn the organization name into just an icon. If the organization owner has an email address from a non-email-provider domain, we use that domain's favicon. Otherwise, we show a lettermark (no image uploading for now).
     > At this point we can likely release the feature and unflag it.
-1. Backend: We mark the `team_id` field as deprecated on `GroupTypeMapping`, `EventDefinition` and `PropertyDefinition` - should not be used anymore with `project_id` serving as the replacement.
-1. Zapier integration: Rename project selection to environment selection.
+13. Backend: We mark the `team_id` field as deprecated on `GroupTypeMapping`, `EventDefinition` and `PropertyDefinition` - should not be used anymore with `project_id` serving as the replacement.
+14. We denormalize `project_id` on models from point 3 by adding the field directly to those modules. Then we make viewsets filtering by `team__project_id` filter by `project_id`.
+15. Zapier integration: Rename project selection to environment selection.
 
 ### Migration path for existing customers
 
--- a/requests-for-comments/2024-01-23-project-environments.md
+++ b/requests-for-comments/2024-01-23-project-environments.md
@@ -143,6 +143,10 @@ By default, a project would only have one environment implicitly named "Producti
 
 Of course, nothing comes for free: adding evironments has its own downsides. Here, it's the amount of effort required – this feature would touch every part of the app, plus ingestion. It would realistically be a few person-sprints of work (as always, it's hard to estimate accurately).
 
+### Pricing
+
+Environments should start out as a paid feature, since they're key for professional users, and more of a nice-to-have for hobbyists. I would suggest it falling under the same restrictions as multiple projects, i.e. multiple environments being available after putting in a card.
+
 ### Rough scope of work
 
 1. Backend: We add an `Environment` model (`posthog_environment` table, fields `id`, `team`, `data_team_id`, `name`, `api_token`) and hook it up to a new project-scoped viewset (`/api/projects/:id/environments/`). Every time a new environment is created, we increment the counter of the `Team.id` field and use the obtained value as the environment's `data_team_id`.
--- a/requests-for-comments/2024-01-23-project-environments.md
+++ b/requests-for-comments/2024-01-23-project-environments.md
@@ -158,19 +158,15 @@ API-wise, subdiving `posthog_team` would mean that resources would be *project-w
 
 Let's do the same as above with one tweak: we turn `posthog_team` into environments, and add `posthog_project` as a parent level of `posthog_team`. This way we can even start the beta phase of environments early on, and simply keep adding support for project-wide handling on a feature-by-feature basis. The core of ingestion meanwhile requires no changes.
 
+To optimize the implementation phase, we can cut opt-in project-wideness out of the scope of work. We'll evaluate needs around this topic more deeply once we have users of the environments feature. The final solution may involve something else, such as permissions or tags.
+
 ### Rough scope of work
 
-1. Backend: We add a `Project` model (`posthog_project` table, columns `id`, `name` initially) and hook it up to a new project-scoped viewset (tentatively: `/api/v2/projects/`). Meanwhile, `posthog_team` gets one new nullable column: `project_id`.
-  From now on `posthog_team` means "environment". Every time a new project is created, it gets a default environment. At this point in implementation, new-style projects are created automatically for every new environment (i.e. old-style project) but cannot be created manually.
+1. Backend: We add a `Project` model (`posthog_project` table, initially integer column `id` + string column `name`). Meanwhile, `posthog_team` gets one new nullable column: `project_id`.
+  From now on `posthog_team` means "environment" internally. For ease of transition, the first environment in a project always has the same ID as the project.  
+  At this point in implementation, new-style projects are created automatically for every new environment (i.e. old-style project), but cannot be created manually.
 1. Backend: For every Team which isn't under a Project yet, we create a matching default Project. After this, every single environment has a parent project, thus we make `project_id` non-nullable.
-1. We add endpoint `/api/v2/environments/` ("v2" part tentative) as an alias of `/api/projects/`, and also `/api/v2/projects/:project_id/environments/` specifically for environment creation.
-1. Frontend, behind a flag: We turn the project selector into the environment selector. This means sectioning the selector by project, renaming "New project" to "New environment" and scoping that flow by project, and finally adding a "New project" button to allow *real Project* creation
-    > During project creation, we must make it crystal clear for customers splitting by region that cross-environment analytics isn't possible – environments are designed specifically to solve data cross-contamination.
-1. Backend + frontend, behind a flag: We add nullable field `project_id` to the `GroupTypeMapping` (along with a constraint to ensure a group type index is not reused within a `posthog_project`), backfill the field in the same migration, and then make it non-nullable. We reuse `ClickhouseGroupsTypesView` for `/api/v2/projects/:project_id/group_types/` and – in the flagged UI – switch to this endpoint from the environment-level version. (Zero real impact currently, as no user has multiple environments within a project *yet*.)
-1. Plugin server: We add the `posthog_team.project_id` field to `TeamManager` and subsequently update `GroupTypeManager` to map group types per `project_id` rather than `team_id`.
-1. Backend + frontend, behind a flag: We add nullable field `project_id` to the `EventDefinition` & `PropertyDefinition`, backfill the field in the same migration, and then make it non-nullable. We reuse `EventDefinitionViewSet` & `PropertyDefinitionViewSet` for `/api/v2/projects/:project_id/{event,property}_definitions/` and – in the flagged UI – switch to these endpoints from the environment-level versions.
-1. Plugin server: We update `PropertyDefinitionsManager` to map group types per `project_id` rather than `team_id`.
-1. Backend: To models of all optionally project-wide entities, we add boolean field `is_project_wide`:
+1. Backend: For all project-scoped viewsets, we change the queryset filtering from `team_id=self.team_id` to `team__project_id=self.team_id`. See list below for models associated with such endpoints:
     - `Dashboard`
     - `Insight`
     - `Action`
@@ -181,11 +177,18 @@ Let's do the same as above with one tweak: we turn `posthog_team` into environme
     - `Survey`
     - `SessionRecordingPlaylist`
    
-    For each of these models, we update its endpoint's `get_queryset()` to include entities from other environments that qualify due to `is_project_wide`.
-    > At this point we can make Environments an early access feature, as we keep adding support for project-wide entities to each of the features listed above. Early adopters start creating environments.
+    This way we're able to postpone denormalizing `project_id`.
+1. We move `TeamViewSet` from `/api/projects/` to `/api/environments/`. At `/api/projects/` we instead put a new `ProjectViewSet`, which is based on the `Project` model BUT behaves exactly like `TeamViewSet`. This makes the transition invisible to existing API consumers, and we can do it thanks to the first environment's ID always being the same as the project's ID. As for settings, they stay at the `posthog_team` level, but we can work around this by still returning and updating settings at `/api/projects/`, just always using the first environment. This ensures backward compatibility, even if it's sneaky.  
+  For environment creation specifically, we add `/api/projects/:project_id/environments/`.
+1. Frontend, behind a flag: We turn the project selector into the environment selector. This means sectioning the selector by project, renaming "New project" to "New environment" and scoping that flow by project, and finally adding a "New project" button to allow *real Project* creation
+    > During project creation, we must make it crystal clear for customers splitting by region that cross-environment analytics isn't possible – environments are designed specifically to solve data cross-contamination.
+1. Backend + frontend, behind a flag: We add nullable field `project_id` to the `GroupTypeMapping` (along with a constraint to ensure a group type index is not reused within a `posthog_project`), backfill the field in the same migration, and then make it non-nullable. We reuse `ClickhouseGroupsTypesView` for `/api/v2/projects/:project_id/group_types/` and – in the flagged UI – switch to this endpoint from the environment-level version. (Zero real impact currently, as no user has multiple environments within a project *yet*.)
+1. Plugin server: We add the `posthog_team.project_id` field to `TeamManager` and subsequently update `GroupTypeManager` to map group types per `project_id` rather than `team_id`.
+1. Backend + frontend, behind a flag: We add nullable field `project_id` to the `EventDefinition` & `PropertyDefinition`, backfill the field in the same migration, and then make it non-nullable. We reuse `EventDefinitionViewSet` & `PropertyDefinitionViewSet` for `/api/v2/projects/:project_id/{event,property}_definitions/` and – in the flagged UI – switch to these endpoints from the environment-level versions.
+1. Plugin server: We update `PropertyDefinitionsManager` to map group types per `project_id` rather than `team_id`.environments.
 1. Frontend, behind a flag: We move the current "Project settings" to a new "Environment settings" section, and add the new-style project name to "Project setting".
-1. Backend + frontend, behind a flag: [UX to be specified] We allow feature flag rollout groups to be either project-wide or environment-specific when the flag as a whole is project-wide.
-1. Frontend, behind a flag: To prevent the UI from being overwhelming, we turn the organization name into just an icon. If the organization owner has an email address from a non-email-provider domain, we use that domain's favicon.Otherwise, we show a lettermark (no image uploading for now).
+1. Backend + frontend, behind a flag: [UX to be specified] We allow feature flag rollout groups to be either project-wide or environment-specific.
+1. Frontend, behind a flag: To prevent the UI from being overwhelming, we turn the organization name into just an icon. If the organization owner has an email address from a non-email-provider domain, we use that domain's favicon. Otherwise, we show a lettermark (no image uploading for now).
     > At this point we can likely release the feature and unflag it.
 1. Backend: We mark the `team_id` field as deprecated on `GroupTypeMapping`, `EventDefinition` and `PropertyDefinition` - should not be used anymore with `project_id` serving as the replacement.
 1. Zapier integration: Rename project selection to environment selection.
--- a/requests-for-comments/2024-01-23-project-environments.md
+++ b/requests-for-comments/2024-01-23-project-environments.md
@@ -153,7 +153,7 @@ Of course, nothing comes for free: adding evironments has its own downsides. Her
 
 ### Pricing
 
-Environments should start out as a paid feature, since they're key for professional users, and more of a nice-to-have for hobbyists. I would suggest it falling under the same restrictions as multiple projects, i.e. multiple environments being available after putting in a card.
+Having multiple environments should start out as a Teams/Enterprise feature.
 
 ### Rough scope of work
 
--- a/requests-for-comments/2024-01-23-project-environments.md
+++ b/requests-for-comments/2024-01-23-project-environments.md
@@ -99,7 +99,7 @@ Because feature flags use cohorts and actions in their definitions, these are al
 
 ### Summarizing requirements
 
-It looks like **users always need to keep the _data_ isolated** between environments – features: *events*, *people*, *groups*, *recordings*. Data sources and destinations should be environment-specific as well: *data warehouse tables*, *apps*, and *batch exports*.
+It looks like **users always need to keep the _data_ isolated** between environments – features: *events*, *people*, *groups*, *recordings*. Data sources and destinations should be environment-specific as well: *data warehouse tables*, *apps*, *batch exports*, *Zapier hooks*, as well as *dashboard/insight subscriptions*.
 
 **Analysis setup should be shareable** – crucially features: *dashboards*, *insights*, *actions*, and *cohorts*. Often also: *event and property definitions*, and *group types*. *Playlists* don't seem to have come up, but should act analoguously to dashboards.
 Shareable means that by default all of these entities would be environment-specific by default, but could easily be made synced between environments.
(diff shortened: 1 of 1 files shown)


requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,64 @@
...
+- experiments
+- surveys
+- early access features
+- group _types_
+- event/property definitions
+- actions
+- cohorts
+- pipeline filters and transformations
+
+The "Feature flag environments" RFC brought up one key need users have: often, the test setup needs some deviations from prod. You're likely to play around with instrumentation during development, so feature flags or cohorts might deviate in dev and you probably don't want that setup to be the source of truth. It's the production setup which you want to be a stable source of truth.
```
raquelmsmith, line 50: Stripe has a "Copy to Production" button in the test mode so you can be selective about what you copy over.

Twixes:
Idea from talking with @mariusandra: repurpose test accounts filter.

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,116 @@
...
+1. By stage, driven by the engineering workflow, e.g. dev/staging/production.
+2. By region, driven by data protection requirements, e.g. country A/country B.
+
+When doing that, customers typically want to share setup between environments, while ensuring *data* from one environment doesn't pollute the others.
+
+The multiple environments use case is one of the reasons why projects are a feature of PostHog, and projects are great for tracking separate products, but not ideal for tracking multiple instances of the same product: whatever you set up for one project, you have to manually replicate in all others.
+
+We currently solve this in Feature flags with a feature for syncing flags between projects, based on the flag key. Feature flags make this uniquely easy thanks to the feature flag key being an effective organization-wide identifier. No other entity – dashboards, insights, etc. – has this advantage though.
+
+Because feature flags use cohorts and actions in their definitions, these are also synced in the current solution –but only based on the cohort/action name, so if the name is changed in one environment, the link is silently broken.
```
neilkakkar, line 16: Hmm, this isn't right. When the name is changed in the target environment, we create a new cohort/action that is the same name as the source environment.

So nothing should ever be silently broken, it's just that _new_ synced flags will connect not to a cohort you've changed in the target environment, but a copy of the cohort as it currently is in the source environment.

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,116 @@
...
+
+The needs are unclear for:
+
+- Data warehouse tables
+
+## Rejected solution: Cross-project copying
+
+This is what some users have asked for verbatim. What it would probably mean is a feature flags-style "Projects" view on the item level, allowing an item to be copied/synced across projects. The reliability of syncing is key though, and the problem mentioned before is that there's no inherent cross-project ID for this. Just basing on the entity name would be extremely fragile – there needs to be an underlying link better than that.
+
+This means we'd need to maintain such underlying links (e.g. a cross-project ID, adjacent to the primary key ID) for all entities we'd like to work cross-project. In one way, this could be nice if you'd like to develop, say, a dashboard in the test environment and only copy it to the production environment when final. Divergence becomes easy. But there would be an act of manual syncing, making the experience cumbersome for users at scale. This would work, but it wouldn't be turnkey.
```
neilkakkar, line 70: Not saying we shouldn't reject this solution, but one important thing, at least for flags is that divergence is a feature not a bug - you would want to freely edit a flag once you've copied it to a dev environment to test out flows / do random things - and you don't want any of these to reflect in the prod environment, for whatever reason.

So the usual flow I've seen so far is:
1. Setup flag as we'd like in prod
2. Sync to dev
3. Test in dev, do debugging in dev (editing flag)

neilkakkar: Although, would be pretty nice if there was a way to reset the dev flag to be back to the prod-like state

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,116 @@
...
+
+[insert metric for how often flags differ within org]
+
+## Proposed solution: Environments
+
+I believe to solve this problem once and for all, we need environments to be a first-class feature, specifically: subdivisions of a project.
```
neilkakkar, line 76: Would love to hear more on this, and what it looks like for the end user. Does it mean a separate copy of Cohort with ID 1234 per environment? And users can edit it per environment? Do they still have to 'sync' between environments? Or is there no copy, and the same cohort always takes the same values in all environments?

requests-for-comments/2024-01-23-project-environments.md
Twixes: Yup, that's what I mean by the _link_ being silently broken. The feature flag itself is not broken, but the syncing mechanism is very implicit, so there's no visibility into how the dependencies of a flag are synced on subsequent syncs.

requests-for-comments/2024-01-23-project-environments.md
Twixes: There is no copy – the cohort is defined once in the project. It's like a universal version of the "Filter out internal and test users" toggle, except you define what's "internal and test" when initializing the client library, and not at query time.

requests-for-comments/2024-01-23-project-environments.md
Twixes: Do you think we could make rollout groups optionally environment-specific, assuming that the API token is uniquely assigned to an environment? This would be trivial with decide as far as I can tell, but I know less about what info local evaluation has

requests-for-comments/2024-01-23-project-environments.md
neilkakkar: Hmm, I see, this is a bit scary for any products that affect the end-users app, like flags and surveys: Making changes in the dev environment _shouldn't_ affect production at all.

At the same time, I can see how this can be very nice for dashboards - can't imagine wanting different dashboards in different environments. Although, even here, I think it will be ~~scary~~ annoying if I'm debugging some event issue in dev and end up saving the dashboard, which then causes the production dashboard to update too.

I'm not sure this works well when environments are different country-based envs for example. In here you'll necessarily have different filters on the same dashboard, depending on the environment?

requests-for-comments/2024-01-23-project-environments.md
neilkakkar: Yep local evaluation does the same, it's passed only one set of release conditions based on the token, and _we could_ do the same as decide here to evaluate the right version of flags.

neilkakkar: although, want to dig deeper before we go down this route, because right now this feels like a fancier way of replicating the copy to project button 😅

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Rejected solution: Cross-project copying
+
+This is what some users have asked for verbatim. What it would probably mean is a feature flags-style "Projects" view on the item level, allowing an item to be copied/synced across projects. The reliability of syncing is key though, and the problem mentioned before is that there's no inherent cross-project ID for this. Just basing on the entity name would be extremely fragile – there needs to be an underlying link better than that.
+
+This means we'd need to maintain such underlying links (e.g. a cross-project ID, adjacent to the primary key ID) for all entities we'd like to work cross-project. In one way, this could be nice if you'd like to develop, say, a dashboard in the test environment and only copy it to the production environment when final. Divergence becomes easy. But there would be an act of manual syncing, making the experience cumbersome for users at scale. This would work, but it wouldn't be turnkey.
+
+[insert metric for how often flags differ within org]
+
+## Proposed solution: Environments
```
annikaschmid, line 74: It is not really clear to me what the difference is between multiple projects vs multiple environments. As in, do we just add another level of hierarchy, or would "environments" support different functionality compared to projects?

You have slightly been referring to it in you response to Neil below, but I want to confirm my assumption is correct.

Each environment contains the same "Dashboards, Insights, Actions, Cohorts, Feature flags, etc." because they are being set once on the project level and then available in any environment. Is that correct?

requests-for-comments/2024-01-23-project-environments.md
Twixes: For feature flags, it very much is a fancy way of replicating the "copy to project" button. Users experience this problem on a scale much broader than just feature flags though, hence the RFC. This is one of these problems that needs a product-wide solution – if we were to solve it separately for feature flags _and_ dashboards/insights _and_ cohorts/actions _and_ taxonomy, we'd end up with grand patchwork.

Thanks for pointing out divergence as a feature! In the rewritten RFC I propose we make rollout groups environment-specific

requests-for-comments/2024-01-23-project-environments.md
Twixes: I reviewed more tickets and chatted with some customers. I included some of this research, but it looks like they mostly want dashboards/insights/cohorts/etc. to be environment-specific, with select ones being promoted to production. I think that'll be a safe and intuitive workflow where we still avoid all the woes of copying – included in the RFC

requests-for-comments/2024-01-23-project-environments.md
Twixes: At first glance copying sounds easy and I would have loved it to be the straightforward solution, but there are unfortunately complications… I describe them in the rewritten version RFC, so will mark this as resolved and take further notes to the new iteration.

annikaschmid:
I just reviewed this second draft. Great work including all the additional customer feedback, makes it much easier to have an opinion on the proposed solution.

Here are a couple of assumptions I would make based on the proposed solution, let me know if there is anything I got wrong:

One thing I initially struggled with in the original proposal was that it wasn't clear to me if by default, all dashboards and insights would be available in all environments. This might be something users wouldn't want, e.g. the company that uses projects to provide a similar setup for multiple companies (they might have some dashboards and not others available for certain companies)

What is now clear in this revised edition is that a entity is by default available in one environment only, but the user can decide to make them available in all others (or selected only?).
For each entity, there might be a couple of parameters that are environment specific as well, such as dashboard filters (?) and rollout conditions. The current flag-copy feature should then be reserved for copying specific rollout conditions, while the existence of flags will be managed by this new environments concept.

We might also not have the need anymore to be able to copy flags across projects, our recommendation here should be to use environments.

UI considerations: We probably have to make it very clear if an entity is available in multiple environments, and making changes to it will affect the setup in all environments.

Therefore, advice to users should be: If you want your setup to be independent, use projects. If you want your setup to be synced, use environments.

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,189 @@
...
+
+<img width="939" alt="image" src="https://github.com/PostHog/meta/assets/4550621/ec966c6b-0a1c-4cb0-a593-0234f4807ab1">
+
+### Downsides
+
+Of course, nothing comes for free: adding evironments has its own downsides. Here, it's the amount of effort required – this feature would touch every part of the app, plus ingestion. It would realistically be a few person-sprints of work (as always, it's hard to estimate accurately).
+
+### Pricing
+
+Environments should start out as a paid feature, since they're key for professional users, and more of a nice-to-have for hobbyists. I would suggest it falling under the same restrictions as multiple projects, i.e. multiple environments being available after putting in a card.
```
simfish85, line 150: I'd actually be in favour of moving this to Teams or Enterprise

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,189 @@
...
+
+To support large customers with access control needs, we'd need to add environment-specific project membership to the existing project access control system.
+
+To support the common feature flags workflow of rollout conditions being different between dev and production, we'd make flag rollout groups always environment-specific.
+
+To ensure straightforward yet complete separation of data at the ingestion level (incl. persons), *as well as* steady querying performance, each environment would have its own `team_id`. Due to this requirement, environments would have to be predefined by users and not created at ingestion time, thus each environment would also have its own SDK key (aka "project API key" currently). That seems like a safe choice in terms of the UX.
```
tiina303, line 138: We want to have multiple api keys per project - sources in CDP.

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,189 @@
...
+
+	3 users commented to express their need, one specifically mentioned dashboards, insights, and actions as currently problematic.
+
+#### User interviews
+
+- [Feature flag & experiment & dashboard copying - R, by Annika](https://posthog.slack.com/archives/C011L071P8U/p1671640760520999)
+
+	They run development and staging, same as customers above they find manual replication tedious and error-prone.
+
+### Summarizing needs
```
tiina303, line 100: Could we write out the requirements (must have vs nice to have) more explicitly here, before jumping into discussing solutions.

Furthermore, could we separate the requirements we have for the backend side vs the UI.

Here's a starting point for the requirements:
1. events, people, groups, recordings, surveys are separate and don't need sharing across projects
3. different features/products can have different sharing configurations (e.g. one might want that all dashboards are shared across all projects, but feature flags are not shared across dev and prod projects).
4. we don't duplicate data in the backend nor run syncs across projects
5. UI doesn't require use to individually set sync for each new item they create (note that in the backend we might set something always based on a higher level configuration)
...

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Pricing
+
+Environments should start out as a paid feature, since they're key for professional users, and more of a nice-to-have for hobbyists. I would suggest it falling under the same restrictions as multiple projects, i.e. multiple environments being available after putting in a card.
+
+### Rough scope of work
+
+1. Backend: We add an `Environment` model (`posthog_environment` table, fields `id`, `team`, `data_team_id`, `name`, `api_token`) and hook it up to a new project-scoped viewset (`/api/projects/:id/environments/`). Every time a new environment is created, we increment the counter of the `Team.id` field and use the obtained value as the environment's `data_team_id`.
+1. Backend: For every existing project (`Team`) we create a matching default `Environment`, reusing the project's `id` (for `data_team_id`) and `api_token` (for `api_token`). Any time a new project is created, it also gets a default environment based on the same logic.
```
tiina303, line 155: For maximal flexibility we could instead do this (note that we should hash out the requirements more fully before to better understand the needs):

for each sharable item (feature_flag, insight, ...) we'll change them to instead of having a `team_id` to have an `organization_id` and `sharing_key` (that's a UUID unique across everything <- though technically unique across organizatisation is good enough)

Then we'll have a new table `shared_across_projects`, that has 
1. organization_id
2. team_id
3. sharing_key

We could then have configurations like this:

Imagine the `shared_across_projects` table looks like this:
| organization | team | sharing_key |
|-|-|-|
| PostHog | Website-EU | 1234 |
| PostHog | Website-US | 2233 |
| PostHog | Website-EU | 5555 |
| PostHog | Website-US | 5555 |

we can have release annotations being separate, because they hit production at different times, but some annotations are shared
| Annotation name | organization | sharing_key | timestamp
|-|-|-|-
| release-380 | PostHog | 1234 | 15:00
| release-380 | PostHog | 2233 | 15:10
| Christmas holidays | PostHog | 5555 | dec 24th

We can have dashboard & insights that are shared together across projects (note I'm re-using the sharing key from above, but it could be a different sharing key)
| dash name | organization | sharing_key | ...
|-|-|-|-
| Overview | PostHog | 5555 | 
| EU only dash | PostHog | 1234 |
| US only dash | PostHog | 2233 |

| insight name | organization | sharing_key | ...
|-|-|-|-
| Overview first insight | PostHog | 5555 | 
| Overview second insight | PostHog | 5555 | 
| EU only dash first insight  | PostHog | 1234 |
| EU only dash second insight  | PostHog | 1234 |
| US only dash first insight  | PostHog | 2233 |

Then when looking up dashboards for the project - we'll query all the dashboards for the organization & all the sharing keys for the organization and join over them in the `shared_across_projects`. If this is too slow we can denormalise some info somewhere as a quicker read path. Optionally we can add a product field (dashboards, feature flags, ...) to the `shared_across_projects` for speeding up lookups.

In the UI the user could say 
1. share everything across everything - in which case we'd just use a single sharing key everywhere (note that we could do splits later, by overwriting the sharing keys in the tables)
2. all my dashboards should be share across all my projects, but everything else should be separate - cool, we just use the single sharing key for all dashboards & unique by project sharing keys for everything else. 
3. The user from above can later pick that I want to create a dashboard that isn't shared (by creating a new sharing key in the backend). They can then later choose that this dashboard should be shared with some other projects (we can add entries to the `shared_across_projects` table to make that happen ... note that we'd want to think through how it works for related objects, i.e. insights on a dashboard).
4. during ingestion we'd look up by team_id, sharing_key for group_type creation - we can also denormalise this (duplicate the data), for it to be faster to access

tiina303 (commented):
👍  nice RFC

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,189 @@
...
+
+  > 	A transition period begins at this step. For some time we'll have to ensure `api_token` stays synced between the `Team` (legacy scheme) and its default `Environment` (future scheme). This only means updating the token reset function to update both.
+
+1. Backend: We add environments to `TeamSerializer`.
+1. Backend: We add `current_environment` to the `User` model.
+1. Plugin server: We switch to `posthog_environment` as the source of truth for tokens and team IDs. The only thing we continue processing at the project level is event/property definitions.
```
xvello, line 161: To confirm, does this mean that persons and groups are scoped by environment?
  - on a split by state, this is desirable (to avoid a bug in dev breaking the prod dataset)
  - on a split by region, we must make it clear that cross-environment analytics won't be possible, and that properties / groups might be a more flexible way to achieve this. I'm just foreseeing the project merge requests coming in after people ignore this caveat.

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,189 @@
...
+
+	They run development and staging, same as customers above they find manual replication tedious and error-prone.
+
+### Summarizing needs
+
+It looks like **users always need to keep the data isolated** between environments – features: *events*, *people*, *groups*, *recordings*. *Toolbar authorized URLs* should also likely be different, due to separate domains.
+
+**Analysis setup should be shared** – crucially features: *dashboards*, *insights*, *actions*, and *cohorts*. Often also: *event and property definitions*, and *group types*. Shared *feature flags* are also a common ask, with *experiments* sometimes mentioned.
+
+I've never heard of the following features mentioned in terms of environments: *playlists*, *surveys*, *early access features*, *annotations*, *data warehouse tables*, *apps*, or *batch exports*. Playlists seem similar enough to dashboards that they should also be shared. As for the rest, it's hard to say.
```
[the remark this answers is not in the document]
xvello: Apps are indeed a tricky situation, I see the value in having them synced for consistency, but I feel like per-environment makes the most sense:
  - as a user, I'd love to test out transforms on the dev dataset before promoting them to the prod dataset
  - exports will probably be by environment, as you don't need to archive dev data, or you'll want it in a different warehousing dataset. We could achieve this by filtering the events by environment, but it feels wasteful

It would be relatively easy to have plugin-server support a mix of per-project and per-environment pluginconfigs, but the UI for this might get confusing?

requests-for-comments/2024-01-23-project-environments.md
tiina303: btw I wrote up a quick requirements doc draft for this: https://docs.google.com/document/d/150Uly1PpjAX5C_uZH9qkqGltrlBKyiEhbI9WVX9eW9s

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Proposed solution: Environments
+
+We can make environments a first-class entity in PostHog – each project would be subdivided into one or more environments. The *data* would be isolated between projects, but the *taxonomy* and *analysis setup* would be shared (see "Summarizing needs" for what should be isolated and what should be shared exactly).
+
+To support the most common workflow of drafting dashboards/etc. in dev/staging and then "promoting" to production, entities such as dashboards would be environment-specific by default, but could be rolled out to the whole project with two clicks. When rolled out, the entity's dependencies would be made project-wide as well (e.g. for a cohort, all actions and other cohorts used in the definition). This would involve no copying though, so there would be no effort needed to sync things.
+
+To support large customers with access control needs, we'd need to add environment-specific project membership to the existing project access control system.
+
+To support the common feature flags workflow of rollout conditions being different between dev and production, we'd make flag rollout groups always environment-specific.
```
neilkakkar, line 136: I think this will apply to anything that depends on flags too then?

For example, if you have a survey in dev with a targeting flag that's promoted to prod, the dev flag should be promoted as well with the same conditions

neilkakkar: Also, I'm assuming this mean sync remains then, but becomes a lot more dumb? Because if someone needs to _copy_ dev rollout conditions to prod they should be able to, since these are different per environment. I'm _hoping_ we can have a similar flow as 'promote' for rest of the things, rather than a completely different one for flags, that basically updates one release condition to be the other.

Since cohorts/actions are shared, having the same IDs, this should be pretty reasonable to do with no real syncing logic required.

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Proposed solution: Environments
+
+We can make environments a first-class entity in PostHog – each project would be subdivided into one or more environments. The *data* would be isolated between projects, but the *taxonomy* and *analysis setup* would be shared (see "Summarizing needs" for what should be isolated and what should be shared exactly).
+
+To support the most common workflow of drafting dashboards/etc. in dev/staging and then "promoting" to production, entities such as dashboards would be environment-specific by default, but could be rolled out to the whole project with two clicks. When rolled out, the entity's dependencies would be made project-wide as well (e.g. for a cohort, all actions and other cohorts used in the definition). This would involve no copying though, so there would be no effort needed to sync things.
```
neilkakkar, line 132: One potential complication that we should figure out a plan for in advance:

Essentially this means ability to query across `data_team_id` s for a query, which can be a bit dicey. Ex: a project-wide cohort used in a environment specific flag -> means you get the flag from the environment specific data_team_id and then the cohort from the 'production' data_team_id, because I assume if it's project-wide it doesn't get a second `environment_id` ? Or are you planning that whenever a new environment is added, to add a link from project-wide Cohorts, say, to also have the environment_id of the new environment? (and thus all cohorts that depend on it as well)

neilkakkar:
Great job, excited to see where this goes! 👍

requests-for-comments/2024-01-23-project-environments.md
neilkakkar: Also, assuming we still never want to share anything across teams (projects) right?

requests-for-comments/2024-01-23-project-environments.md
Twixes: What you're outlining is exactly what proposal B "Dashboards/insights/etc. sharing between projects" would be. However, as the RFC points out, that runs into the problem of taxonomy being different projects, and especially group types.
I'm not sure what you mean by "during ingestion we'd look up by team_id, sharing_key for group_type creation" and if that solves the problem. Can you clarify?

This is definitely not feasible: "I'd start with seeing if this model works for dashboards and insights & then slowly add it to more products" – at the very least the set of shareable entity types has to be: dashboards, insights, cohorts, actions. This also means you couldn't say "all my dashboards should be share across all my projects, but everything else should be separate" – no, if dashboard X is shared across projects, then all its dependencies must now be shared across those same projects. And the important complication is that it can't be "use only in this project" or "share across all projects", because then if you want to run two  _completely separate_ projects in one org, you can no longer do that. But let me know if I misunderstood what you meant on this point.

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,189 @@
...
+
+To support large customers with access control needs, we'd need to add environment-specific project membership to the existing project access control system.
```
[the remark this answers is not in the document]
Twixes: Yes, because it would be one dashboard.

requests-for-comments/2024-01-23-project-environments.md
neilkakkar: .. oh and then environment permissions mean restricting some of these flows :oof-2:

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,189 @@
...
+
+Of course, nothing comes for free: adding evironments has its own downsides. Here, it's the amount of effort required – this feature would touch every part of the app, plus ingestion. It would realistically be a few person-sprints of work (as always, it's hard to estimate accurately).
+
+### Pricing
+
+Environments should start out as a paid feature, since they're key for professional users, and more of a nice-to-have for hobbyists. I would suggest it falling under the same restrictions as multiple projects, i.e. multiple environments being available after putting in a card.
+
+### Rough scope of work
+
+1. Backend: We add an `Environment` model (`posthog_environment` table, fields `id`, `team`, `data_team_id`, `name`, `api_token`) and hook it up to a new project-scoped viewset (`/api/projects/:id/environments/`). Every time a new environment is created, we increment the counter of the `Team.id` field and use the obtained value as the environment's `data_team_id`.
```
tiina303, line 154:
```suggestion
1. Backend: We add an `Environment` model (`posthog_environment` table, fields `id`, `project_id`, `source_id`, `name`, `api_token`) and hook it up to a new project-scoped viewset (`/api/projects/:id/environments/`). Every time a new environment is created, we increment the counter of the `Team.id` field and use the obtained value as the environment's `data_team_id`.
```

requests-for-comments/2024-01-23-project-environments.md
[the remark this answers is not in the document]
raquelmsmith: I think this is probably super important to get right - making sure people know exactly where their changes will have impacts when they change a shared entity. 

I'm not certain that having access to edit something you can't see is the right move, but we can probably discuss RBAC for this at a later time.

requests-for-comments/2024-01-23-project-environments.md
Twixes: Maybe rollout groups should _optionally_ be environment-specific then? Not really more complex backend-wise, just trickier UI-wise – but would support _all_ the use cases

requests-for-comments/2024-01-23-project-environments.md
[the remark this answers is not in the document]
Twixes: Good points, I will make it explicit that apps will be environment-specific

requests-for-comments/2024-01-23-project-environments.md
Twixes: Implemented

requests-for-comments/2024-01-23-project-environments.md
Twixes: There is definitely no querying across `data_id` (fka `data_team_id`_ intended). I'm not sure what you mean with "link from project-wide Cohorts, say, to also have the environment_id of the new environment". If you make a cohort project-wide, we make all its dependencies project-wide too and ensure this stays that way. But it's not aware of multiple environments in any situation.

requests-for-comments/2024-01-23-project-environments.md
Twixes: Added this as part of requirements

requests-for-comments/2024-01-23-project-environments.md
Twixes: Yes, good point, added clarity around this as a point

requests-for-comments/2024-01-23-project-environments.md
Twixes: Fair, some requirements I kind of only "revealed" n later sections (reflecting what the process of writing the RFC looked like 😄). Consolidated in this section. Let me know if it looks like I've missed anything

requests-for-comments/2024-01-23-project-environments.md
Twixes: Makes sense, and we talked about this sync with Tiina. Added a step early on where we split out project API keys from `Team` into an `SDKKey` model

xvello: >> new table with a foreign key to a single environment.
> to a single team?

to a single data_team_id, so an environment. We'll want dev and prod to have separate capture tokens, to route events to the right environment within a project.

requests-for-comments/2024-01-23-project-environments.md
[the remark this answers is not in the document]
Twixes: Hmm, this probably doesn't matter that much with dev/prod, but I can see how it might be an issue with a region-based split where different people work on each region. I propose this: You can't save changes to project-wide items if you don't have access to all its environments (and you also can't make things project-wide then). This should work since the engineers in charge of the central set of analysis should anyway have access to all the environments.

requests-for-comments/2024-01-23-project-environments.md
Twixes: Yup

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Downsides
+
+Of course, nothing comes for free: adding evironments has its own downsides. Here, it's the amount of effort required – this feature would touch every part of the app, plus ingestion. It would realistically be a few person-sprints of work (as always, it's hard to estimate accurately).
+
+### Pricing
+
+Environments should start out as a paid feature, since they're key for professional users, and more of a nice-to-have for hobbyists. I would suggest it falling under the same restrictions as multiple projects, i.e. multiple environments being available after putting in a card.
+
+### Rough scope of work
```
xvello, line 158: While I appreciate the zero-downtime transition you're describing, I'm really concerned about the bugs that will arise from having the `team_id` column's meaning being inconsistent between tables (it says the project on event props, but now means environment in apps). The fact that single-environment projects would not be impacted (because env ID == project ID) will increase the risk of such bugs leaking through QA and only manifesting on multi-env teams (our big customers).

I know that renaming columns in PG and CH is a locking operation and a non-backwards-compat change, but I feel like a one-hour maintenance window (where we still capture to kafka and serve decide from redis) to rename all `team_id` columns into `env_id` an `project_id` would save us a lot of headache and bugs down the line. [pgroll](https://github.com/xataio/pgroll) could enable a zero-downtime rename, but I don't think it would play well with future django migrations.

xvello (commented):
Thanks for all the clarifications! I do have concerns about the bugs caused by a vestigial `team_id` column lying around, creating a separate thread for this specific concern.

requests-for-comments/2024-01-23-project-environments.md
neilkakkar: So, cohort A is project-wide, which means all dependent cohorts become project wide.

But, then flag A depends on cohort A, so cohort A doesn't have a dependency on flag, but the flag does. And the flag can still be environment specific, even if cohort A is project-wide right?

Twixes: Yep

requests-for-comments/2024-01-23-project-environments.md
Twixes: Changed to "Teams/Enterprise only"

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -147,42 +153,39 @@ Of course, nothing comes for free: adding evironments has its own downsides. Her
...
-1. Backend: For every existing project (`Team`) we create a matching default `Environment`, reusing the project's `id` (for `data_team_id`) and `api_token` (for `api_token`). Any time a new project is created, it also gets a default environment based on the same logic.
-
-  > 	A transition period begins at this step. For some time we'll have to ensure `api_token` stays synced between the `Team` (legacy scheme) and its default `Environment` (future scheme). This only means updating the token reset function to update both.
-
+1. Backend: We add an `Environment` model (`posthog_environment` table, columns int `id`, `team_id`, `data_id` `name`, `api_token`) and hook it up to a new project-scoped viewset (`/api/projects/:id/environments/`). Every time a new environment is created, we determine its `id` by incrementing the counter of the `posthog_team.id` column and using that value as the environment's `data_id`.
```
tiina303, line 160: can you update this to remove the `api_token` from here too?

requests-for-comments/2024-01-23-project-environments.md
Twixes: Yeah, there's a bunch of cleanup that we should do to avoid confusion in the future 😅 Starting with `posthog_team`, which should be `posthog_project`. It's an exceptionally tricky problem though, so I propose we do it as cleanup at the end of the project.
pgroll sounds great actually, why do you say it wouldn't play well with Django migrations?

requests-for-comments/2024-01-23-project-environments.md
Twixes: 👍 Good catch

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -147,42 +153,39 @@ Of course, nothing comes for free: adding evironments has its own downsides. Her
...
+1. Backend: For every project existing currently we create a matching default `Environment`, reusing the project's `id` the environment's `data_id`. Any time a new project is created, it also gets a default environment based on the same logic.
+1. Backend: We add an `SDKKey` model (columns UUID `id`, `environment_id`, `value`, `created_by`, `created_at`) and populate it with current `api_token` values from `posthog_team` (for each project using the project ID as `environment_id`).
+    > Until we replace the project API key in project settings with environments, we'll have to sync the project API key being reset with its `posthog_sdkkey` equivalent.
```
tiina303, line 163: You don't need to create this table now, to avoid the syncing problem. We can for now keep using the `api_token` in the `posthog.team` table under the `data_id` team & not change anything in ingestion. It's still good to document the plan here though.

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -147,42 +153,39 @@ Of course, nothing comes for free: adding evironments has its own downsides. Her
...
 1. Backend: We add environments to `TeamSerializer`.
 1. Backend: We add `current_environment` to the `User` model.
-1. Plugin server: We switch to `posthog_environment` as the source of truth for tokens and team IDs. The only thing we continue processing at the project level is event/property definitions.
+1. Plugin server: We switch to `posthog_environment` as the source of truth for tokens and event/person `team_id`. From now on, `team_id` in ClickHouse will mean `posthog_environment.data_id`. The only thing we continue processing at the project level is event/property definitions.
```
tiina303, line 166: Related to my comment above: we don't need any changes to plugin-server until we make multiple tokens possible via the the `SDKKey` (maybe `SourceKeys` as it doesn't need to be an sdk they can send us POST requests from anywhere)

requests-for-comments/2024-01-23-project-environments.md
tiina303: heavy +1 for this. Let's start with any new columns we add shouldn't use 'team' in it's name (currently 
> `posthog_environment` table, columns int `id`, `team_id`,

requests-for-comments/2024-01-23-project-environments.md
Twixes: We don't need to create it now, but we do probably need to create it early. Either we move to `SourceKey.value` or to `Environment.api_token`, but we definitely can't keep using `Team.api_token`.

requests-for-comments/2024-01-23-project-environments.md
xvello: > It's an exceptionally tricky problem though, so I propose we do it as cleanup at the end of the project.

It is my experience that such cleanup will not be prioritized if not in the critical path, so we should either acknowledge the tech debt we're creating and discuss how to mitigate its impact, or commit to not creating it in the first place.

> pgroll sounds great actually, why do you say it wouldn't play well with Django migrations?

We'd need to create a django migration for the renames, that is applied on new instances but neutralized on the cloud regions, and apply it through pgroll on prod. I see many ways for us to shoot ourselves in the foot with this and create an outage.

Ellie campaigned for us to move away from django migrations and apply schema migrations in a out-of-band zero-downtime process and I support this, especially as we're decoupling new domains (new webhook delivery, next person model) from the django ORM and having a common migration tooling would be great. I'm afraid that would not be a short-term project though.

requests-for-comments/2024-01-23-project-environments.md
xvello: I'm not sure I follow where you're going here. My understanding is that events and persons are segmented per environment, so each environment needs its separate capture token / api_key. Even if we don't ship token-per-source yet, we'll need several tokens per project, hence the need for the separate table.

Similarly, plugin-server will need to resolve this per-environment token into an env and team for processing, so there's some changes to be done there.

After that table exists, shipping several tokens per env will be a frontend-only change.

requests-for-comments/2024-01-23-project-environments.md
Twixes: Hmm, I think the critical cleanup part here is renaming ClickHouse `team_id` to `environment_id`. This can result in really bad confusion without cleanup.
The rest is a nice to have because we'll still use "team" for "project".

Twixes: As for naming columns in new models `project_id` instead `team_id`, I'm worried this would be _even_ more confusing until we do the complete cleanup on an undefined timeline. When we do the complete cleanup, it will be marginal effort to rename one more column, so I'll go for `team_id` for continued consistency when meaning projects

xvello: >  The rest is a nice to have because we'll still use "team" for "project".

We agreed that apps, batch export and data warehouse entities stay at the environment level, so at least in these tables `team_id` will mean environment unless we migrate them. Right?

xvello: Instead of `team_id` == project and create a new environment entity, I'd love to evaluate what schema changes would be needed if we say `team_id` == environment, and we create a new project entity on top of it. This means that all existing entities stay linked to the environment and can be promoted to shared across the whole project?

tiina303: Just to make this more clear here's what the current proposal is:
Screenshot 2024-03-01 at 16 11 49

here's an alternative we could consider instead

Screenshot 2024-03-01 at 16 11 41

Twixes: Ah, I agree that this is way better. So actually in this scenario the "sources" table is orthogonal, because all of ingestion will still use `posthog_team.api_token` (which will mean project API keys simply become environment API keys for the time being, with no extra changes). Will rewrite the plan to do things this way

requests-for-comments/2024-01-23-project-environments.md
Twixes: Actually scratch that, per https://github.com/PostHog/meta/pull/176/files#r1509157452. We can avoid ingestion surgery by doing things as you propose

requests-for-comments/2024-01-23-project-environments.md
xvello: Indeed, we keep a single token per environment for now, then we'll tackle the per-source-token later this year.

The one thing that's not really clear yet is the eventdefinition/propertydefinition/eventproperty table group: they store a mix of:
  - user-defined stuff (descriptions, verified) that makes sense at the project level
  - all the names that plugin-server sees, that might be an issue: your staging data might mess up the dropdown for prod events

tiina303: updated plan
Screenshot 2024-03-01 at 17 05 06

requests-for-comments/2024-01-23-project-environments.md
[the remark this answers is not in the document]
raquelmsmith: That seems like a good place to start 👍

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,192 @@
...
+
+Of course, nothing comes for free: adding evironments has its own downsides. Here, it's the amount of effort required – this feature would touch every part of the app, plus ingestion. It would realistically be a few person-sprints of work (as always, it's hard to estimate accurately).
+
+### Pricing
+
+Having multiple environments should start out as a Teams/Enterprise feature.
+
+### Rough scope of work
+
+1. Backend: We add an `Environment` model (`posthog_environment` table, columns int `id`, `team_id`, `data_id` `name`) and hook it up to a new project-scoped viewset (`/api/projects/:id/environments/`). Every time a new environment is created, we determine its `id` by incrementing the counter of the `posthog_team.id` column and using that value as the environment's `data_id`.
```
benjackwhite, line 160: I don't want to add a lot of noise at this stage _but_ did you consider the idea of re-using projects for this?

Core idea:
1. Projects can be created as sub-projects of an existing one
2. The additional property `parent_team_id` is added to each Team
3. When you select an "environment" what you are actually doing is selecting a different project but the UI shows it as a sub project
4. Loading of dashboards/insights etc. use the `parent_team_id` instead with only the data specific endpoints going to the sub-project

The potential benefits would be:
1. API routes remain the same as before - everything is project scoped
2. UI routes work as before which is super nice (we don't need `current_environment` or eventually some way to include it in the URL)
3. No need to re-implement permissions - these would apply as before but just with a UI makeover to make it clear what has changed
4. You could import existing projects easily into a parent project giving us a simple migration path for existing users who want to group their staging / prod projects

This might need a little more fleshing out of course. Just wanted to propose it as it feels like a way faster way of getting to the same result without introducing a bunch of extra models.

requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -0,0 +1,192 @@
...
+1. Frontend, behind a flag: We add the environment as a breadcrumb item with a dropdown + creation modal.
+    > We must make it crystal clear for customers splitting by region that cross-environment analytics isn't possible – environments are designed specifically to solve cross-contamination.
+1. Backend: To models of all optionally project-wide entities, we add boolean fields `environment_id` and `is_project_wide`:
+	- `Dashboard`
+	- `Insight`
+	- `Action`
+	- `Cohort`
+	- `FeatureFlag`
+	- `SessionRecordingPlaylist`
+1. Backend: In all endpoints that query ClickHouse, we add support for filtering by environment ID instead of project ID for querying:
```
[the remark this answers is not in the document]
benjackwhite: I added a comment further up about how we could re-use Teams for this instead. If we do that then this comment can basically be ignored, otherwise we should probably figure out this stuff as well as what do about frontend routing (how to link someone to a specific environment)

requests-for-comments/2024-01-23-project-environments.md
benjackwhite: The main issues I see in this approach are
1. Project specific things like Insights with promotion (promoting a dashboard from staging to production). This would have to be done on the parent project still with some additional property or _in_ the sub project but then you have the copying issue
2. Shared ingestion-y type things such as Event/Property definitions - same could apply as in the proposed solution - if a `parent_team_id` is set then we process in that team's definitions instead (or both)

benjackwhite: Made a PR here to flesh out how I think this differs implementation-wise https://github.com/PostHog/meta/pull/185

requests-for-comments/2024-01-23-project-environments.md
Twixes: Implemented this plan as new proposed solution D. There are three differences from the graph above in the current scope of work I wrote out:
1. Insights/dashboards/actions/etc. only have `is_project_wide`, since `project_id` can be derived with a quick B-tree join on `posthog_team`.
2. We're going to leave all the settings on the environment level initially.
3. We're not going to `AggregatingMergeTree` as part of _this_ project (but the door's open if you want to try that later).

xvello: What do you mean by item 3? Is it the revamp of storage for event & property names? Agreed that this is out of scope for this RFC.

xvello (approved):
Solution D is a go for me 👌. Thanks for the good communication through the iteration process.

requests-for-comments/2024-01-23-project-environments.md
Twixes: Yeah yeah, it's the blue post-it on the chart, just clarifying this is out of scope

PR #182, opened 2024-02-26
add a user agent parsing RFC

pauldambra:
after a quick discussion with @tiina303 we agreed to promote https://github.com/PostHog/posthog/issues/20552 to being an RFC before changing anything

### DIFF
--- a/requests-for-comments/2024-02-26-user-agent-parsing.md
+++ b/requests-for-comments/2024-02-26-user-agent-parsing.md
@@ -1,0 +1,41 @@
+# Request for comments: User Agent Parsing & Paul D'Ambra
+
+## Problem statement
+
+_Who are we building for, what are their needs (include example cases), why is this important?_
+
+-   User agent parsing is ~3% of the posthog-js bundle and needs increasing code golf to keep under control.
+-   It is also duplicated in posthog-js-lite and needs people to remember to update one when they change the other
+-   Because it is in the SDK other SDKs cannot take advantage of it and so we have things like the user agent populator plugin
+-   Any changes are slow to roll out because of the long-tail of SDK versions in the wild
+
+## Success criteria
+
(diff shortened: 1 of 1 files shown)


requests-for-comments/2024-02-26-user-agent-parsing.md
```diff
@@ -0,0 +1,41 @@
...
+_What are the key user experience and technical design decisions / trade-offs?_
+
+### Option 1 - do nothing
+
+It's been fine for 2 years, we copy the changes from https://github.com/PostHog/posthog-js/pull/1038 to posthog-js-lite and move on.
+Fast, but nothing is easier next time we need to change this
+
+### Option 2 - move the code into the plugin server
+
+e.g. in this location https://github.com/PostHog/posthog/pull/20548 (but not that implementation)
```
pauldambra, line 30: would need guidance from team pipeline on where the code should live in the plugin server (or for us to implement it - and then change it - to help them decide)

tiina303 (commented):
I'd go for option (3). Furthermore I'd consider moving the `$initial_*` stuff to also be a default enabled universal plugin, so if users don't want the `$initial_` etc stuff on their account they can get rid of it.
Alternatively (1) seems to be a fine state to be in until we have time to do (3).

requests-for-comments/2024-02-26-user-agent-parsing.md
```diff
@@ -0,0 +1,41 @@
+# Request for comments: User Agent Parsing & Paul D'Ambra
+
+## Problem statement
```
pauldambra, line 3:
```suggestion
## Decision

Option 3 - extend the user agent parser so it matches (or betters) the JS SDKs behaviour, enable it for everyone, and deprecate from the SDKs

(doing nothing is a fine holding position so this isn't urgent)

## Problem statement
```

pauldambra:
see also https://github.com/PostHog/posthog/issues/18025

PR #184, opened 2024-02-29
RFC: Daily onboarding emails for users

joethreepwood:
Going to get started putting this together because an opt-in flow feels uncontroversial, but thoughts appreciated. 

Especially interested in thoughts from @MarconLP about common setup issues or questions we can head off here.

### DIFF
--- a/requests-for-comments/2024-02-29-daily-emails.md
+++ b/requests-for-comments/2024-02-29-daily-emails.md
@@ -1,0 +1,43 @@
+# Request for comments: Daily emails for users
+**Owner:** @joethreepwood
+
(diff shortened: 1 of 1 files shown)


charlescook-ph:
Making it opt in is interesting - you could frame it in a fun way potentially as like a micro-course or something too? (Beyond just 'hey sign up for daily tips'). 

But happy to try this as an experiment and see what happens!

MarconLP:
I like this idea and the content covers the most common issues. I think `setting up the reverse proxy` should be the first email, to avoid "I don't see events in PostHog" tickets for users with an installed AdBlock extension.

joethreepwood:
Launched this. Will report back with metrics in a few weeks.

joethreepwood:
Just for transparency, we have deactivated this campaign today after only seeing 11 users sign up for it.

PR #185, opened 2024-03-04
wip: Alternative proposal for environment implenentation

benjackwhite:
Felt like the most useful way to explain my thinking on this comment https://github.com/PostHog/meta/pull/176/files#r1510674757

### DIFF
--- a/requests-for-comments/2024-01-23-project-environments.md
+++ b/requests-for-comments/2024-01-23-project-environments.md
@@ -131,15 +131,15 @@ The UX would still be a little clunky, with a need to manually set up multiple f
 
 ## Proposed solution: Environments
 
-We can make environments a first-class entity in PostHog – each project would be subdivided into one or more environments. The *data* would be isolated between projects, but the *taxonomy* and *analysis setup* would be shared (see "Summarizing needs" for what should be isolated and what should be shared exactly).
+From a user perspective that is what we will call them as it is clear and understandable. From an implementation perspective we would extend the `Team|Project` concept to allow any Project to have a parent Project indicating that it's meta configuration should happen there (dashboards, insights etc.).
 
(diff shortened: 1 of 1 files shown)


requests-for-comments/2024-01-23-project-environments.md
```diff
@@ -131,15 +131,15 @@ The UX would still be a little clunky, with a need to manually set up multiple f
 
 ## Proposed solution: Environments
 
-We can make environments a first-class entity in PostHog – each project would be subdivided into one or more environments. The *data* would be isolated between projects, but the *taxonomy* and *analysis setup* would be shared (see "Summarizing needs" for what should be isolated and what should be shared exactly).
+From a user perspective that is what we will call them as it is clear and understandable. From an implementation perspective we would extend the `Team|Project` concept to allow any Project to have a parent Project indicating that it's meta configuration should happen there (dashboards, insights etc.).
```
Twixes, line 134: This sounds a lot like [Tiina&Xavier's proposal to turn the meaning of "Team" into "Environment"](https://github.com/PostHog/meta/pull/176/files#r1509205302), and build projects as a _parent_ level to environments (instead of _subdividing_ `Team`s into environments, which is what I proposed initially).
There is just one thing I'm wary about in your proposal @benjackwhite, but it's a bug one: tech debt. Some `posthog_team` acting as parents (projects) to other ones (environments) sounds like a mess to serialize _and_ develop. For example, some fields on `posthog_team` then apply individually in each project, while others are only relevant in "root" ones (e.g. the person display name). And the same quirk starts to apply to shared event/property definitions, as well as any other shared resources.

benjackwhite: Hmmm I guess their proposal is similar but with a dedicated parent object which I'm also not against _necessarily_. 

The main thing is - we will have a real issue with API and UI endpoint design that we should figure out up front (`current_environment` feels like a real hack :/

PR #203, opened 2024-05-07
Add RFC for repeatable surveys

Phanatic:
My first RFC at Posthog, would love any feedback you have on content/style/approach.

### DIFF
--- a/requests-for-comments/2024-05-07-repeatable-surveys.md
+++ b/requests-for-comments/2024-05-07-repeatable-surveys.md
@@ -32,6 +32,10 @@ Our current workarounds of `just repeat the survey manually` might get them to r
 1. [Pendo's custom NPS survey type](https://support.pendo.io/hc/en-us/articles/360033527151-Set-up-an-NPS-survey#h_01HGD0SKYSRB6SNKXBFJ1QG7M7)
 2. [Surveymonkey's multiple responses](https://help.surveymonkey.com/en/surveymonkey/send/allowing-multiple-responses/)
 
+This is how UserPilot shows NPS scores changing over time 
+
+<img width="1115" alt="image" src="https://github.com/PostHog/meta/assets/82819/3fbaa08d-aa79-4781-b0f4-07b56242486f">
+
 
 ## Design
 _What are the key user experience and technical design decisions / trade-offs?_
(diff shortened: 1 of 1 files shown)


requests-for-comments/2024-05-07-repeatable-surveys.md
```diff
@@ -0,0 +1,75 @@
...
+
+1. `repeat this survey 4 times, once every 3 months`.
+2. They can also just leave the `survey_max_iteration_count` as zero to have an infinitely repeating survey `repeat this survey, once every 3 months`
+
+These two properties are nullable and we need to make our work backwards compatible to make existing survey responses function without issues to show NPS scores etc. 
+
+#### UI changes. 
+1. Show the user an interface to pick the repeat interval of the survey, this can be relative (every 4 months) or absolute (pick specific dates). 
+4. Show the user an interface to pick the iteration count, how many times should this survey repeat? 
+5. If a survey has an iteration count of more than 1, change the existing UI to run custom clickhouse queries to account for a survey with multiple response sets, one for each iteration.
```
neilkakkar, line 64: would love to hear more about what this means, and how we want to do this?

For example, what if during the 2nd iteration, the same person never responds? Does it still make sense to breakdown results into iteration sets?

It's not clear to me how we should special case this, if at all.

I _think_ what might be useful here is a 2nd view only for users who've responded multiple times, and aggregating only for these users to show changes over time: so we have an overall view, that treats multiple responses as distinct, and this iteration-aggregation view, that only looks at people who've responded multiple times.


To make the above magic work, we absolutely need to make sure we get rid of this bug of users accidentally responding multiple times, which I think you're considering already.

requests-for-comments/2024-05-07-repeatable-surveys.md
```diff
@@ -0,0 +1,75 @@
...
+
+These two properties are nullable and we need to make our work backwards compatible to make existing survey responses function without issues to show NPS scores etc. 
+
+#### UI changes. 
+1. Show the user an interface to pick the repeat interval of the survey, this can be relative (every 4 months) or absolute (pick specific dates). 
+4. Show the user an interface to pick the iteration count, how many times should this survey repeat? 
+5. If a survey has an iteration count of more than 1, change the existing UI to run custom clickhouse queries to account for a survey with multiple response sets, one for each iteration. 
+
+#### posthog-js changes.
+1. To decide when to show a survey, see if a user has a survey response submitted for the current survey start date, and show the survey if they haven't.
```
neilkakkar, line 67: > see if a user has a survey response submitted for the current survey start date

How are you planning to query this?

Also worth noting that you'll need to put the local storage cache on a ttl / find another way around not showing the same survey repeatedly. There are some complex patterns here to handle:

1. When a user first dismisses/responds to a survey, we add something to local storage to not show this survey again.
2. Now, I'm assuming there's some mechanism like flags(?) controlling when to next show it. This will almost certainly have some delay (since we need to send properties back via ingestion, which can take some time).
3. Because of the async nature of (2), the client side needs some way to remember not to show this survey again, hence (1). But now, if we do want to re-show it, we also want to delete (1), hence some ttl there.

requests-for-comments/2024-05-07-repeatable-surveys.md
```diff
@@ -0,0 +1,75 @@
...
+3. When a survey is submitted, add properties for survey iteration count and survey start date in the current iteration  to the person properties. So this person is accounted as having submitted the survey for this iteration/start date.
```
neilkakkar, line 68: unclear to me how this is used in point (1) above, or how we use this info?

Are you creating a targeting flag behind the scenes with info like:

`$survey_response/${survey.id}/iteration_count` is less than `4`

AND

`$survey_response/${survey.id}/${current_iteration}` is not set ?

neilkakkar: this looks somewhat different than what you're proposing

neilkakkar (commented):
Nice!

Some questions on the implementation, but overall seems reasonable!

optional, would love to see a bit more about exactly what do competitors do with the results display

requests-for-comments/2024-05-07-repeatable-surveys.md
Phanatic: > For example, what if during the 2nd iteration, the same person never responds? 
Great callout, I hadn't considered variable response rate from users over a time period. I'll do some more thinking here and update the doc. 

> I think what might be useful here is a 2nd view only for users who've responded multiple times
Yeah, this makes sense. You can't track NPS if its a new set of people every time who responded in a given time period.

requests-for-comments/2024-05-07-repeatable-surveys.md
Phanatic: For a given survey iteration, e.g. 2 of 4, if a user submits a response for iteration 2 on May 13th 2024, 
we will add the following person properties. 

``` 
$survey_response/${survey.id}/2
$survey_response/${survey.id}/05-13-2024
```

requests-for-comments/2024-05-07-repeatable-surveys.md
Phanatic: This is, as you noted, going to be complicated. 
For a given survey iteration, we can't depend on the local storage having the necessary information to show/hide a survey given how long the iterations last. 
To get over this, we could associate each iteration of the survey with its own feature flag, that has a time based activation.

e.g. `survey-targeting-flag-iter-1` is only true for 90 days, until Aug 11 2024. and 
`survey-targeting-flag-iter-2`  is from Aug 11 2024 to November 9th 2024. And we create these flags on a schedule based on the survey frequency. `posthog-js` just uses this feature flag information to show/hide the dialog. And when it sends the event, it sends the `iteration_id` as a person property, from the survey object properties.

requests-for-comments/2024-05-07-repeatable-surveys.md
Phanatic: I added some more color about how the results will be displayed and additional chart types we can show

PR #207, opened 2024-05-28
RFC: Bringing back the monorepo for rust services and main app

neilkakkar:
Looking for alignment on how to do this, and help on doing some of these changes

### DIFF
--- a/requests-for-comments/2024-05-28-monorepo.md
+++ b/requests-for-comments/2024-05-28-monorepo.md
@@ -11,6 +11,8 @@ Having things split up between hog-rs and posthog app make a couple things prett
 3. It's harder to keep track of changes that need to be made everywhere / keeping code in sync across repos.
 4. We need to re-create similar scaffolding to make things work in the rust repo (example: GeoIP db downloading, webhooks data model setup, etc.)
 
+Now, I'm looking for alignment on the steps below, and whether this makes sense, and if I'm missing anything.
+
 ## Design 
 *What are the key user experience and technical design decisions / trade-offs?*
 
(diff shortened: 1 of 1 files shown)


requests-for-comments/2024-05-28-monorepo.md
benjackwhite, line 1: Only want to say big 👍 there are always downsides to a monorepo but in our current upsides the benefits far out way the cons (and we can always reverse it later).

requests-for-comments/2024-05-28-monorepo.md
```diff
@@ -0,0 +1,46 @@
...
+Now, I'm looking for alignment on the steps below, and whether this makes sense, and if I'm missing anything.
+
+## Design 
+*What are the key user experience and technical design decisions / trade-offs?*
+
+To make this work, here's what we need to do: (non-exhaustive, hoping you can add what I'm missing!)
+
+Effectively, nothing should change for anyone developing on `hog-rs` currently, and people developing on main `posthog` repo should get more options to hook things up.
+
+1. [ ] Introduce a `rust/` top-level folder that has all rust services
```
benjackwhite, line 23: If we go with a rust top level it should make the whole "only run the CI for this code" quite straightforward.

benjackwhite: Also one thought - you could do this in the `hog-rs` repo first to test out that it works as expected.

benjackwhite (approved):

requests-for-comments/2024-05-28-monorepo.md
```diff
@@ -0,0 +1,46 @@
...
+2. [ ] Copy over hog-rs deployment / docker-image build, and make sure this doesn't run on random django changes
+3. [ ] Figure out all main repo CI workflows that we need to now skip for the `rust/` folder, if any
+5. [ ] [Can be done a bit later] - setup e2e tests that run across rust services and main app to make sure things work properly together
+
+
+### Migration plan
+
+I'm not 100% sure yet how we want to migrate things over, or over what timeline.
+
+Given hog-rs deploys happen manually right now, it seems de-risked enough to do it in one fell-swoop to minimise keeping-code-in-two-places-in-sync-pains. So I'd propose:
```
Gilbert09, line 42: All sounds good, except this. TIL we have to deploy hog-rs manually. If this was embedded within `posthog/posthog` then I would have assumed that deployments were automatic with the rest of the codebase. With `hog-rs` being a separate repo, I would have intentionally seeked out how it got deployed

neilkakkar: Correct - eventually this will be automated, but for now it's good enough for hog-rs editors 👍

mariusandra (approved):

requests-for-comments/2024-05-28-monorepo.md
Twixes: For a True Monorepo™ it would also be great to put everything which is not `rust/`/`plugin-server/`/`plugin-transpiler/` under `web/` – since it doesn't make sense that the plugin server and such are _under_ the Django server in the hierarchy. Architecturally, they're all on the same level. (This is scope creep)

neilkakkar:
not all the things have been done, but monorepo is here now, so merging this, we can incrementally improve the rest of the things when we want/need to!

PR #216, opened 2024-07-02
Add alerting models, task architecture RFC

webjunkie:


requests-for-comments/2024-07-01-alerting.md
```diff
@@ -0,0 +1,160 @@
...
+4. **Efficiency and Scalability**: Handle tasks efficiently to reduce load on the system.
+
+#### Architecture
+
+1. **Models**:
+    - **Alert**: Represents the alert configuration.
+    - **AlertCheck**: Logs each evaluation of an alert, including results and notification statuses.
+
+2. **Task Execution**:
+    - Use Celery to handle periodic checks and notifications.
```
pauldambra, line 24: just checking whether it should be temporal rather than celery (i don't even know how we'd choose 🙈)

requests-for-comments/2024-07-01-alerting.md
```diff
@@ -0,0 +1,160 @@
...
+
+@shared_task
+def schedule_alert_checks():
+    alerts = Alert.objects.all()
+    alert_ids = [alert.id for alert in alerts]
+
+    alert_id_groups = [alert_ids[i:i+10] for i in range(0, len(alert_ids), 10)]
+
+    group_of_chains = group(
+        chain(check_alert.chunks(group, 10)) for group in alert_id_groups
```
pauldambra, line 156: when we do implement this it definitely needs a comment - i think i understand what it's doing but i bet i'm wrong 🙈

requests-for-comments/2024-07-01-alerting.md
pauldambra, line 1: we already have a set of scheduled tasks which try and keep the cache warm and we're not monitoring them well (IMHO as the person who left them not well monitored last time they were changed) 

this is effectively the same mechanism for a different reason

we could overlap these mechanisms so that caching overall gets better too
but we should definitely learn from the past and make sure that we know how often these tasks run, when they fail, and have appropriate alerting in place

requests-for-comments/2024-07-01-alerting.md
```diff
@@ -0,0 +1,160 @@
...
+    
+    # Keep in mind idempotency in case of retries - e.g., check if check is already in the database for this interval
+
+    calculated_value = 42  # Example calculated value - hand off calculation to existing insight code
+    threshold_met = True  # Example threshold check
+
+    alert_check = AlertCheck.objects.create(
+        alert=alert,
+        calculated_value=calculated_value,
+        threshold_met=threshold_met,
```
pauldambra, line 108: since we're adding thresholds... i'd love to be able to add thresholds in the UI that don't alert just so I can get some nice visuals on insights without needing alerts

that has the benefit we can ship it as soon as its made and start benefitting and getting user feedback on the UX

pauldambra:
it's so cool we're working on this

requests-for-comments/2024-07-01-alerting.md
```diff
@@ -0,0 +1,160 @@
...
+
+We might consider discarding old `AlertCheck` records after a certain period to manage database size.
+
+#### Task Execution with Celery
+
+##### Why Use Groups, Chains, and Chunks
+
+**Groups**: Allow multiple tasks to be executed in parallel, providing the ability to run concurrent checks for 
+different alerts. By creating groups of tasks, we can limit the number of parallel operations to control system load 
+and ensure efficient resource utilization.
```
nikitaevg, line 69: I wonder maybe [limit_concurrency](https://github.com/PostHog/posthog/blame/4c4a2733987fda5619f648e3aafbc679f4005ea4/posthog/clickhouse/client/limit.py#L47) (that you introduced) would work here? Basically that's all we need - limit the number of tasks in flight, no?

If we add limit_concurrency and simply process all alerts in a `for`, IMO it would have the same effect as if we used chains + groups + chunks, but it would be simpler, or do I miss something?

requests-for-comments/2024-07-01-alerting.md
nikitaevg: That's my concern as well, it looks a bit difficult, and that's why I suggest for + limit_concurrency

requests-for-comments/2024-07-01-alerting.md
nikitaevg: Worth noting that it might be not a good idea to use cache for alerts, users probably want the freshest value to be tested, at most 5-10 minutes of staleness.

I should look into it actually, because alerts that use cached stale values are not very good alerts.

nikitaevg (commented):
Thanks for writing this!

requests-for-comments/2024-07-01-alerting.md
```diff
@@ -0,0 +1,160 @@
...
+notification preferences, and other settings based on user requirements. Future integration with CDP is also tbd.
+
+##### AlertCheck Model
+
+The `AlertCheck` model logs each evaluation of an alert:
+- `alert`: Foreign key to the Alert model.
+- `created`: Time of the check.
+- `calculated_value`: Result of the check.
+- `anomaly_condition`: Copy of the condition for the alert in case original got updated.
+- `threshold_met`: Whether the condition was met.
```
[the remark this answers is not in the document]
pauldambra: maybe `trigger_condition` and `trigger_condition_met` ??

requests-for-comments/2024-07-01-alerting.md
pauldambra: mostly I don't think we should have a cache task processing X per minute and an alert task processing X for per minute, since we're going to overload processing infrastructure

or at least we shouldn't ignore that we have both

the more i think about it i think this might need limits - free customers can have 2 alerts, paid can have 10, (fill in numbers as you see fit) 

we can always allow more, but can't restrict them (AWS has soft and hard limits to protect service, but you can ask them to increase soft limits)

(similarly we should consider if this is going to be in the open source or EE licensed portion of the code)

requests-for-comments/2024-07-01-alerting.md
```diff
@@ -0,0 +1,160 @@
...
+
+    alert_check = AlertCheck.objects.create(
+        alert=alert,
+        calculated_value=calculated_value,
+        threshold_met=threshold_met,
+        ...
+    )
+
+    if threshold_met:
+        send_notification.s(alert_check.id).delay() #  Launch notification task based on check object
```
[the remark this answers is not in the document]
pauldambra: > The only profit I see is that if a notification fails and we want to retry it, we won't calculate the metric value again

importantly 

> we won't calculate the metric value again.

that's a really valuable thing
some queries are inherently very slow, if we can avoid running them too often we should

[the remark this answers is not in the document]
pauldambra: we absolutely should be able to retry a notification N times but only calculate the value once (or at least <N times since we don't want to retry for 6 hours without checking if the alert has recovered 🙈)

requests-for-comments/2024-07-01-alerting.md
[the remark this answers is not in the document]
nikitaevg: > trigger_condition

:shrug: sounds good

> trigger_condition_met

I'd prefer having an enum here, IMO it will need more values than 2. You think check_result is not good?

requests-for-comments/2024-07-01-alerting.md
nikitaevg: > the trick here (however we do it) is to limit the number of running tasks such that if the number of celery pods scale the max number of tasks does not.

IIUC the limit_concurrency implementation, it looks at the number of tasks in flight and drops the new one if it's above the limit, so I'd say it will keep the limit if we increase the number of pods, right?

> maybe teaching everyone to suck eggs 🙈

I haven't worked with Celery before, so not at all! What an odd expression though.

requests-for-comments/2024-07-01-alerting.md
nikitaevg: > the more i think about it i think this might need limits - free customers can have 2 alerts, paid can have 10, (fill in numbers as you see fit)

Yeah, I agree with that, Mixpanel allows 50 alerts per client, probably there's a reason for that.

2 and 10 alerts seems a bit small though:) maybe anything above some number should be paid? 

> (similarly we should consider if this is going to be in the open source or EE licensed portion of the code)

I'd prefer keeping it in the open source, at least while I'm participating - I might require an extra approval from my employer to contribute to EE directories .

requests-for-comments/2024-07-01-alerting.md
pauldambra: > What an odd expression though. 

🤣 

i have a bad habit of speaking in idioms - sorry!

requests-for-comments/2024-07-01-alerting.md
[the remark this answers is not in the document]
webjunkie: Regarding "anomaly": Is there any background or reasoning for this term? I think it evokes other expectations (anomaly detection) and is not quite what we have here with strict set thresholds. I would actually vote to use another term.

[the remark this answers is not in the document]
webjunkie: >On the second thought, maybe it shouldn't be a boolean, but an enum. If there's an error in calculation, I don't know what boolean value can be assigned here. 

We need to decide whether to send out a notification (= threshold met) yes or no. That is boolean.
Errors can be stored separately (see `error_message`).

requests-for-comments/2024-07-01-alerting.md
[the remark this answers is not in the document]
webjunkie: The other reason I put this is that we can easily launch other things based on the trigger, e.g. hand it off to CDP, launch a Slack notification task, and so on. In this way it's prepared to be pluggable.

requests-for-comments/2024-07-01-alerting.md
webjunkie: The limit concurrency implementation I did recently acts more as a safeguard. The reason is that if you put 100 tasks in the queue and they hit the limit, they will start to retry and retry but only do so a couple of times and then fail. So this system isn't good for "slam stuff in the queue and wait for it to be done".

Setting up groups and chains allows us to control that there are 10 groups launched (= 10 running at the same time) and they sequentially check the alerts. That's a simple way I would say to control how much is done at a time.

requests-for-comments/2024-07-01-alerting.md
pauldambra: yep, i want to see the threshold on the graph because that's useful but don't need an alert for every threshold

e.g. measuring web vitals, i want to colour the graphs because its hard to remember the 4 different thresholds (well, 8 because they have good, medium, poor) but I don't need an alert when they change cos they don't always indicate something i need to immediately react to

requests-for-comments/2024-07-01-alerting.md
pauldambra: yep, super important, and generally better to stay within a tools expectations, just differing levels of celery skill and knowledge so we should make it super clear in the implementation (or safe in some other way tests/alerts/etc)

nikitaevg: About chains, a failure in the middle of a chain will make the whole chain fail, probably not what we want? So we either tweak tasks (e.g. catch all exceptions in the task?) or use another mechanism. 

About limit_concurrency, what if I modify it so that the tasks that don't meet the limit are retried using Celery retry? This solves the problem I described above and can be reused for other tasks.

Also, I looked a bit closer at Celery primitives, won't this all work using `chunks` only? Something like
`check_alert.chunks(all_alert_ids, 10)()`? IIUC, this will invoke 10 tasks, each with its own chunk of work, just like we want. There's a `group()` method for the chunks object, but I don't think it's necessary. The problem with the failures causing other failures remains.

And one more note, maybe the decision what to retry should not come from the task, because it's not synchronized. I'd say ideally it's the dependency's responsibility to decide what it can execute and it can't. But that's outside of this project.

So what do you say I use chunks + catch all exceptions and see how it works? Simple and should work.

webjunkie: If we don't chain the chunks, we will have the effect of Celery picking up all chunks all at once, which is exactly what is to be prevented. Regarding exceptions: That's also useful. If there is some error e.g. regarding Clickhouse connection or something, we break the chain and all others don't need to try. Calculation errors (query syntax problems, timeouts) we can catch and store as error in the alert check.

requests-for-comments/2024-07-01-alerting.md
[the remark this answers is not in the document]
nikitaevg: > Is there any background or reasoning for this term?

It just popped in my head. What term would you prefer? I would say a value that is above or below a user-defined threshold _is_ an anomaly.

> We need to decide whether to send out a notification (= threshold met) yes or no. 

1. What if there's a problem with the alert or insight _configuration_, for example the user misconfigured something, we should notify users about that, no? So whether we send out a notification might depend not only on meeting the threshold.
2. If there's an error with the calculation, what value would the "threshold_met" have? I would say it's neither true nor false, and I wouldn't like to have an optional boolean as it doesn't help readability.
3. In the future alerts might get other anomaly detection mechanisms, so it won't be only about thresholds. That's why I think it's better not mention "threshold" here.
4. Enums are cheap, IMO provide better readability, and are extensible, why do you think it's better to use a boolean? 

Sorry, missed this comment last time.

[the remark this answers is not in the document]
pauldambra: i've built a couple of similar matchers in the past and they always evolve to support `&&` and `||` and i always ended up returning something like `{matches: bool, explanation: string }` - so familiarity with something similar is altering my opinion here (doesn't mean i'm right 😊)

i'd definitely have `error` as a separate case, if there's an error the condition was neither met nor not met

in other parts of posthog we ended up having `error_count` and `last_error_date` so that we could decide whether to carry on trying something

----
tbh there's a predicate condition and a predicate result

maybe it's better to have it that generic
----
this might not always be an anomaly to the user - "tell me when an order value is about $1k" is a happy thing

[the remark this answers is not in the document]
pauldambra: but i should leave you @webjunkie to it 😊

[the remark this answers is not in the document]
nikitaevg: > if there's an error the condition was neither met nor not met

That worries me a bit, I usually find difficult to interpret optional booleans. My rule of thumb is that if something is slightly more difficult than yes/no then I default to an enum - it's more explicit and more extensible. I'm not advocating for different error types in the check_result. I'm just suggesting we use an enum to express this absence of the outcome more explicitly, have a required enum with three states.

And still "Enums are cheap, IMO provide better readability, and are extensible". It's better to have a door open to extensions, especially with databases, isn't it?

requests-for-comments/2024-07-01-alerting.md
nikitaevg: > If we don't chain the chunks, we will have the effect of Celery picking up all chunks all at once, which is exactly what is to be prevented.

I'm not sure this is true. I couldn't find anything in the documentation, so I looked in the implementation. `chunks` splits arguments into groups and then applies `xstarmap` to each group (see [this](https://github.com/celery/celery/blob/bd3b3c6afb8b0c44b416f469b3db6a725d71b609/celery/canvas.py#L1441-L1454)). `xstarmap` is essentially the `starmap` method (see [this](https://github.com/celery/celery/blob/bd3b3c6afb8b0c44b416f469b3db6a725d71b609/celery/app/task.py#L878)). And the starmap method runs sequentially as per the [documentation](https://docs.celeryq.dev/en/stable/userguide/canvas.html#map-starmap).

Also the chunk [documentation](https://docs.celeryq.dev/en/stable/userguide/canvas.html#chunks) hints that there will be only n tasks processing the load ("so that if you have one million objects, you can create 10 tasks with a hundred thousand objects each").

Maybe let's try using chunks alone and see how it goes? Say limit the number of concurrent executions to 1, add 10 alerts and see if they are executed sequentially? I was thinking about experimenting it locally, but I don't think it's easy to set up a fair experiment.

> Regarding exceptions: That's also useful. If there is some error e.g. regarding Clickhouse connection or something, we break the chain and all others don't need to try. Calculation errors (query syntax problems, timeouts) we can catch and store as error in the alert check.

I don't know about that, I would default to checking every alert. IMO it's bad to not execute an alert because it got in a wrong batch, it would hurt users. So I suggest here rethrow a couple types of exceptions that we're sure affect all alerts, but by default swallow exceptions, wdyt?

webjunkie: > you can create 10 tasks

Then they will be running in parallel and might lead to high load.

>Say limit the number of concurrent executions to 1

I suggested a pretty straightforward way to start out and limit the load/concurrency for these alert checks. Since it wasn’t in your PR, I’ve already added it because it’s not complex and we can just try it. You seem to disagree and rule out doing it this way without offering an effective alternative to limit the load. Until it’s safe and won’t slam our cluster with dozens of calculations at the same time, I'm not sure how to proceed.

nikitaevg: > Then they will be running in parallel and might lead to high load.

I don't understand this, we want to detect anomalies in parallel as long as we control the number of tasks in flight, that's what I suggest. If you are claiming that "hundred thousand objects" will run in parallel, I don't think that's true, see my explanation about starmap above.

> What mechanism does this refer to?

Using chunks alone. `check_alert.chunks(all_alert_ids, len(all_alert_ids) / num_of_maximum_parallel_tasks)()`. IIUC the documentation and the implementation it will create `num_of_maximum_parallel_tasks` tasks, they will be executed in parallel, but every chunk will be executed sequentially, see what I wrote about starmap.

> I suggested a pretty straightforward

I'm worried about the straightforwardness of the approach. I need a couple minutes, a piece of paper and a pen to understand what it does, and I'm still not sure I understand it well.

My suggestion is simple - let's try chunks, run an experiment with num_of_maximum_parallel_tasks=1 and 10 alerts, and see if they are processed sequentially or in parallel.

webjunkie: > they will be executed in parallel, but every chunk will be executed sequentially

That's right. And I get what you are suggesting now. However if only using chunks, imagine we have 1000 checks and chunk it by 10, that means the task will have a list of 100 insights to calculate. If this numbers grows and the average calculation time per insight is not super low, each task will run too long (thinking of deploys or other interruptions here). We could start out with that, I just imagine that we basically need to still improve this before releasing alerts more widely.

nikitaevg: Yeah, that's a valid point. Let's keep the chunks+chains+groups then.

nikitaevg: Ok, so about the interruptions you mentioned, I think chains have this problem too. Suppose a worker crashes while checking an alert (due to unrelated reasons), the whole chain will be cancelled, which is not good. We could mitigate it by using [acks_late](https://docs.celeryq.dev/en/stable/userguide/tasks.html#Task.acks_late), but then what if  the crash is caused by the alert check? Then retries will crash the worker indefinitely. I'm not sure either option is good.

 What if instead of splitting the load we limit the concurrency of the worker (like [this](https://github.com/PostHog/posthog/blob/2b1e1a5fb190bdd486229cec4a5bc89a3cd0b827/posthog/tasks/utils.py#L14))? IMO it's a much simpler approach, but I'm not sure it'll work since I don't know PostHog set up. Paul mentioned multiple pods, and it might become an issue, but maybe it's not that difficult to control and doesn't have many tradeoffs?

webjunkie: You are right this would also break the chain 🤔 

I also thought the same if we can limit worker concurrency... the problem is, this concurrency setting is for one worker. We might spin up several pods though. And they also might autoscale to even more. Unfortunately we currently cannot control the worker amount and autoscaling granular per worker type.

I can think about it more, but maybe the case that alerts fail from time to time is not that severe for a first version?

nikitaevg: I suppose there's a way to create a dedicated pod with exact number of CPUs for this queue. But yeah, I see it's not that simple. Let's try the chains and see how it goes.

requests-for-comments/2024-07-01-alerting.md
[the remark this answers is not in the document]
webjunkie: One other thing that comes to mind, which might influence the modeling here: We currently would alert each time we check, if the threshold is continued to be breached, right?
If we find the threshold is met, we might check previous `AlertCheck`s to see if this was already the cases there and if a notification has been sent already.

Regarding the boolean discussion: I'm happy to talk about a specific alternative if you may present one.

>this might not always be an anomaly to the user - "tell me when an order value is about $1k" is a happy thing

This was also my thinking.. anomaly is a too strong word, and we can leave it more generic.. But naming is hard 🤷🏻

requests-for-comments/2024-07-01-alerting.md
webjunkie: It is indeed possible, but someone would need to dig into the configs and templates in the other repo 😓

webjunkie: I just noticed that our infra team pushed an unrelated adjustment to auto scaling, which now allows for exact worker/pod configuration of maximum instances. I'll test this with some other tasks and we might end up just using the same mechanism here then after all 🎉

nikitaevg: Oh, cool!

What do you say we keep the mechanism with chains for now (to unblock development of other features and to merge that [old PR](https://github.com/PostHog/posthog/pull/22762/files)) and if the configuration works out I'll adjust it?

PR #234, opened 2024-08-15
RFC: Hire a YouTuber

ivanagas:
Decider: Charles and Andy
What's the decision: Yes, but in Q1 2024
Planned date for decision: end of Q3

### DIFF
--- a/requests-for-comments/2024-08-14-youtuber.md
+++ b/requests-for-comments/2024-08-14-youtuber.md
@@ -1,5 +1,7 @@
 # Request for comments: Hire a YouTuber
 
+> **Decision:** We are planning on hiring a YouTuber in Q1 2024 :)
+
 PostHog is built on content + word of mouth. YouTube has proven to be an effective channel for word of mouth grow, based on the significant number of people mention YouTube (or Theo) on signup, but we don’t have the expertise to do content right there. We currently differentiate by creating great content and YouTube is a massive opportunity to do that.
 
 YouTube rewards consistency and iteration. Our subscriber base and watch time continues to grow even though we don’t upload more videos. YouTube doesn’t work if you do it part time. You need to go all-in on video. 
--- a/requests-for-comments/2024-08-14-youtuber.md
+++ b/requests-for-comments/2024-08-14-youtuber.md
@@ -1,0 +1,30 @@
+# Request for comments: Hire a YouTuber
+
+PostHog is built on content + word of mouth. YouTube has proven to be an effective channel for word of mouth grow, based on the significant number of people mention YouTube (or Theo) on signup, but we don’t have the expertise to do content right there. We differentiate by 
+
+YouTube rewards consistency and iteration. Our subscriber base and watch time continues to grow even though we don’t upload more videos. YouTube doesn’t work if you do it part time. You need to go all-in on video. 
+
+I think we have seen some successes, but have been too scattered to really capitalize on them:
+
+- Theo + YouTube mentions shows its promise
+- Although YouTube **tutorials** about PostHog don’t seem to work super well, topic-based (like blogs) YouTube videos do.
+- Although, anecdotally, my small number of YouTube videos get **more** mention than my blogs or newsletters from people, especially when they are learning PostHog. The complexity of PostHog lends itself to video. It is a great way to teach people how to use it.
+
+Beyond YouTube, what could a YouTuber do:
+
+- Create video social content + changelog like so many people have wanted to.
+- Experiment with short form content.
+- Collaborate with influencers (we have done this with newsletters and it works well). This would also help with my influencer plan.
+- Read out 20,000 names (just kidding… unless…).
+
+This does require the **right person**, a proper YouTuber, probably someone who already is one now. I think this is a really cool role for them and PostHog is the right company to attract them. We could talk to the YouTubers we are working with now. 
+
+**What does great look like?** YouTube grows as a source of signups. YouTuber is publishing multiple videos per week. Videos are consistently hitting 10k+ views, some viral hits. Developing video as expertise at PostHog.
+
+**Why not another (writing) content person?** There is always more newsletters, docs, tutorials, and blogs to do, but I don’t feel it is enough to add another content person (beyond filling in for James). 
+
+### Next steps:
+
+- Charles + Andy + James: Yes/No
+- Create job post
+- Ian reaches out to influencers we already work with
--- a/requests-for-comments/2024-08-14-youtuber.md
+++ b/requests-for-comments/2024-08-14-youtuber.md
@@ -17,7 +17,7 @@ Beyond YouTube, what could a YouTuber do:
 - Collaborate with influencers (we have done this with newsletters and it works well). This would also help with my influencer plan.
 - Read out 20,000 names (just kidding… unless…).
 
-This does require the **right person**, a proper YouTuber, probably someone who already is one now. I think this is a really cool role for them and PostHog is the right company to attract them. We could talk to the YouTubers we are working with now. 
+This does require the **right person**, a proper YouTuber, probably someone who already is one now. Not Theo or someone with hundreds of thousands of subs. I think this is a really cool role for them and PostHog is the right company to attract them. We could talk to the YouTubers we are working with now. 
 
 **What does great look like?** YouTube grows as a source of signups. YouTuber is publishing multiple videos per week. Videos are consistently hitting 10k+ views, some viral hits. Developing video as expertise at PostHog.
 
--- a/requests-for-comments/2024-08-14-youtuber.md
+++ b/requests-for-comments/2024-08-14-youtuber.md
@@ -21,7 +21,30 @@ This does require the **right person**, a proper YouTuber, probably someone who 
 
 **What does great look like?** YouTube grows as a source of signups. YouTuber is publishing multiple videos per week. Videos are consistently hitting 10k+ views, some viral hits. Developing video as expertise at PostHog.
 
-**Why not another (writing) content person?** There is always more newsletters, docs, tutorials, and blogs to do, but I don’t feel it is enough to add another content person (beyond filling in for James). 
+**Why not another (writing) content person?** We need to fill for James. There is always more newsletters, docs, tutorials, and blogs to do, but I don’t feel it is enough to add another content person (beyond filling in for James).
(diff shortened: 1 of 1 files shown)


requests-for-comments/2024-08-14-youtuber.md
```diff
@@ -0,0 +1,30 @@
+# Request for comments: Hire a YouTuber
+
+PostHog is built on content + word of mouth. YouTube has proven to be an effective channel for word of mouth grow, based on the significant number of people mention YouTube (or Theo) on signup, but we don’t have the expertise to do content right there. We differentiate by
```
charlescook-ph, line 3: 'We differentiate by...' BY WHAT IAN, I HAVE TO KNOW

requests-for-comments/2024-08-14-youtuber.md
```diff
@@ -0,0 +1,30 @@
...
+- Although, anecdotally, my small number of YouTube videos get **more** mention than my blogs or newsletters from people, especially when they are learning PostHog. The complexity of PostHog lends itself to video. It is a great way to teach people how to use it.
+
+Beyond YouTube, what could a YouTuber do:
+
+- Create video social content + changelog like so many people have wanted to.
+- Experiment with short form content.
+- Collaborate with influencers (we have done this with newsletters and it works well). This would also help with my influencer plan.
+- Read out 20,000 names (just kidding… unless…).
+
+This does require the **right person**, a proper YouTuber, probably someone who already is one now. I think this is a really cool role for them and PostHog is the right company to attract them. We could talk to the YouTubers we are working with now.
```
charlescook-ph, line 20: Do you see this more as a contractor or a full time properly in house person at PostHog, or do you not have an opinion either way?

andyvan-ph: I'd say in-house full-time, personally. If we're going to do it, we should go all in.

I love this idea. We did something similar to kickstart a YouTube channel back when I worked in publishing and it was one of the best things we ever did.

There is always a danger they'd eventually leave if their own YT does well, but there's way more incentives to work at PostHog than there was to work for a UK publisher with no money!

Let's do this.

andyvan-ph: Additional thought: There's a small chance there could be some overlap here with the Community Manager role, depending on who we hire there. Worth bearing that mind so don't end up doubling up.

joethreepwood (commented):
I'm butting in here, but I'd pretty strongly advise pausing on this until we have a community manager in. We're actively hiring for that role and we already have some really good candidates in the pipeline and approaching superday. I'd 100% expect this person to have strong opinions here in terms of how we make content, who we work with, and in terms of making content themselves. The candidates I'm looking at have express expertise here. 

I agree with the fact that, currently, we've seen _some_ successes but that it's been too scattered and lacks a strategy. The solution then, IMO, is to recruit someone who can bring a strategy - which is the community manager - and then make more content, rather than to just to make more content. We could hire both at once, but that doesn't feel like it'd work well and I don't think the level of success we've seen would justify two hires at once. 

There's also significant overlap here between the goals of what a YouTube could do beyond making videos and what the Community Manager will do. Influencer work, for example, is an explicit and planned quarterly goal for them.

~I also absolutely think we should not try to hire from influencers we already work with, by the way. That feels like it would be very expensive, would ruin the authenticity and supposed impartiality that those influencers bring by being external, and would risk us hiring someone with an existing style that isn't going to match our brand.~

andyvan-ph (commented):
Agree we should wait until we've hired the community manager role.

I don't see an issue with exploring people we're already working with. Many of them aren't massive and may welcome the chance to develop their skills in-house. We're talking about people early on their journeys here, not Theo-level people.

requests-for-comments/2024-08-14-youtuber.md
```diff
@@ -0,0 +1,30 @@
...
+- Create video social content + changelog like so many people have wanted to.
+- Experiment with short form content.
+- Collaborate with influencers (we have done this with newsletters and it works well). This would also help with my influencer plan.
+- Read out 20,000 names (just kidding… unless…).
+
+This does require the **right person**, a proper YouTuber, probably someone who already is one now. I think this is a really cool role for them and PostHog is the right company to attract them. We could talk to the YouTubers we are working with now. 
+
+**What does great look like?** YouTube grows as a source of signups. YouTuber is publishing multiple videos per week. Videos are consistently hitting 10k+ views, some viral hits. Developing video as expertise at PostHog.
+
+**Why not another (writing) content person?** There is always more newsletters, docs, tutorials, and blogs to do, but I don’t feel it is enough to add another content person (beyond filling in for James).
```
andyvan-ph, line 24: Just so I'm clear, you're suggesting we hire this role instead of replacing James, or in addition to?

charlescook-ph: I read this as in addition to replacing James

joethreepwood:
> Agree we should wait until we've hired the community manager role.
> 
> I don't see an issue with exploring people we're already working with. Many of them aren't massive and may welcome the chance to develop their skills in-house. We're talking about people early on their journeys here, not Theo-level people.

Ah gotcha - I heard influencer in this case and assumed we were thinking to go big leagues. Sorry for misunderstanding!

ivanagas:
@joethreepwood @andyvan-ph added more on why we shouldn't wait for a community manager

requests-for-comments/2024-08-14-youtuber.md
```diff
@@ -0,0 +1,53 @@
...
+
+3. I do think the YouTuber will have a better idea of what content will work on YouTube. We already have some sense of this, but don’t have the bandwidth to execute. For example, more content like: [demo](https://www.youtube.com/watch?v=2jQco8hEvTI), [Clickhouse vs Postgres](https://www.youtube.com/watch?v=iLXXoDaFoxs), [event tracking guide](https://www.youtube.com/watch?v=LIJ_TuyMq74), [modern data stack sucks](https://www.youtube.com/watch?v=2N2cvCmv4us). The community manager will also be scattered with other initiatives and not dedicate themselves to YouTube like I am suggesting. 
+
+4. I still see influencer work sitting entirely with the community manager. It makes it EASIER for them to do their job by having a YouTuber to collaborate with. It brings us to the attention of more YouTubers. 
+
+I’m perfectly fine with this being a “no, we shouldn’t hire a YouTuber” but I don’t think “we should let the community manager decide” is the right answer. Other ones that I think could be better reasons:
```
charlescook-ph, line 42: Yeah I agree that the two roles don't need to be coupled - we have enough informed and strong opinions that a relatively new person doesn't need to come in and tell us who to hire (and they will sit on different teams).

andyvan-ph: This all makes sense to me. I'd suggest:

1. We should at least wait until the upcoming SuperDays are completed. If we end up hiring one of them, they can provide some input before we start advertising.

2. If none of them make it, we should get an ad up for this because it could take a while to find someone suitable, so delaying would be bad.

3. Before we go live with an ad, I think we should spend a bit more time forming an opinion on what we want. We can also look at how other companies like us do YouTube, such as Supabase and Arc. I'd go so far as coming up with a long list of potential video ideas, just to get us thinking about this as a channel before we 100% commit.

requests-for-comments/2024-08-14-youtuber.md
```diff
@@ -21,7 +21,30 @@ This does require the **right person**, a proper YouTuber, probably someone who
...
+
+## We shouldn’t wait for the community manager
+
+Review comments:
+
+> We should wait for the community manager
+
+I do see where you are coming from on a community manager front, this will likely overlap with what their work is, but I disagree on it being the reason we should not hire someone here reasons.
+
+1. The job listing for community manager went out June 7th-ish. This shows it takes a long time to hire for a role. It will take them more time to decide to join, onboard, and get used to the PostHog culture. It is a big unknown, both in terms of what they want to do and when it will happen.
```
joethreepwood, line 34: Small nit here just so we're comparing correctly: the job ad was created then, but IIRC it wasn't public on the site for a fair old while as we were considering handpicked candidates only. The ad only actually went open to public applications in August - and we've had a steady stream of strong applications since.

ivanagas: Fair! Didn't realize this.

requests-for-comments/2024-08-14-youtuber.md
```diff
@@ -0,0 +1,53 @@
...
+- We don’t want to spend more on content, would rather focus on paid ads and other growth marketing.
+- We really don’t think YouTube would work even though it is driving a significant amount of signup mentions.
+- We want to hire this role, but we have too much on our plate hiring a content marketer and community manager. We should wait until they are hired and hire someone 1-3 months later. Not because they will decide for us, but because we don’t have time.
+- Something else 🤷
+
+### Next steps:
+
+- Charles + Andy + James: Yes/No
+- Create job post
+- Ian reaches out to influencers we already work with
```
Lior539, line 53: Flagging Eric Roby, who we've worked previously for the tutorials. He's a software engineer too. Here's his own channel - https://www.youtube.com/@codingwithroby

requests-for-comments/2024-08-14-youtuber.md
```diff
@@ -0,0 +1,53 @@
...
+- Theo + YouTube mentions shows its promise
+- Although YouTube **tutorials** about PostHog don’t seem to work super well, topic-based (like blogs) YouTube videos do.
+- Although, anecdotally, my small number of YouTube videos get **more** mention than my blogs or newsletters from people, especially when they are learning PostHog. The complexity of PostHog lends itself to video. It is a great way to teach people how to use it.
+
+Beyond YouTube, what could a YouTuber do:
+
+- Create video social content + changelog like so many people have wanted to.
+- Experiment with short form content.
+- Collaborate with influencers (we have done this with newsletters and it works well). This would also help with my influencer plan.
+- Read out 20,000 names (just kidding… unless…).
```
Lior539, line 18: My only concern with this RFC is that we're not clear on the type of content we want this person to create, so Im concerned that they wont be set up for success. 
   
   I think to get them started, they can begin converting our newsletter issues to youtube videos. After a few months, they'll hopefully get a good feel for what kind of PostHog content works well and be able to drive it on their own

PR #285, opened 2025-01-29
Add RFC: Streamline startups workflow

kevan-g:
Suggestions for improving the way we manage incoming startup applications

### DIFF
--- a/requests-for-comments/2025-01-29-startups-workflow.md
+++ b/requests-for-comments/2025-01-29-startups-workflow.md
@@ -1,0 +1,37 @@
+# Request for comments: Startups workflow streamlining
+
+## Problem statement
+We need to streamline the startup application and approval workflow. It’s a multidisciplinary job and we need some hands who good at integrations (in between Zapier, Stripe, Vitally, Slack and PostHog-the-app and PostHog.com). Doing this will serve a core audience better, increase our speed.
+
+Right now, when startups and YC members apply for our programs and deals, it's a manual process that results in friction, delay, data loss, missed opportunities, bad reporting and general inefficiency.
+
+## What's wrong with current flow:
+- **It's slow for founders:** Manual effort, email-based (which we hate), asking for duplicate info, all of which gets in the way of their JTBD (shipping). 
+- **It's slow for us:** This program matters to PostHog, and this is unnecessary friction. (Like, poor Scott. There's at least 30 of these a day.)
+
+## How the current flow works:
+1. People fill out the forms on https://posthog.com/startups 
+2. An email (via Zap) goes out from Scott Lewis, asking participants to manually submit their info
+3. Participants manually submit, over email, the date they were founded, their project ID, last 4 digits of their CC, amount they've fundraised, the email address associated with their PostHog account
+4. Over email, Scott receives this, manually reviews the info across Stripe and Vitally to see if they are eligible, and to check whether they have the credits or not.
+5. He then presses Approve in Zapier (which applies the credits in Stripe, and triggers the next workflow of emailing the person with their update + merch stuff)
+
+## What can we do about it? 
+- **Steps 2 and 3,** above, should not have to happen. (We already have a record of *most* of the info, because a PostHog account is required to apply.)
+- I think this means the actual **sign-up form should be gated**, accessible from the app itself — requiring a user to be in a logged-in state to submit. 
+- I think the system should **auto-filter applications:** if they are not eligible, do not apply the credits. *(We can write nice rejections: “Did we get that wrong? Contact Scott if you think there’s been a mistake. We like being generous, but sometimes the our auto-bots are fierce defenders of the realm.”)*
+- I think the system should **auto-approve people:** if they pass the above criteria check, the credits should simply be applied. We can be alerted to abuses if needed, starting with daily manual reviews. 
+
+## Proposed flow:
+ * Remove the actual application form fields (just the form fields) from [the page](https://posthog.com/startups/apply)
+* Replace it with a call-to-action to “Sign in or sign up to apply”
+* Use PostHog-the-app to filter by whether a user has clicked that or not, and show a banner that says “🎉 You’re eligible for the startup plan! Get started here.”
+* Build a new page on the app: /project/###/startups
+* Host the completion steps of the form here. 
+* * All we need to ask for is what we don't already know. I think that's just "How much in total funding have you raised (USD)" and "The date that your company was incorporated."
+* I get a little lost in the plumbing after this and am asking for support thinking through the path.
+
+## Three other considerations:
+* **Solving for the YC program too:** It needs similar care, as the form at [yc-onboarding](https://posthog.com/yc-onboarding) generates all the same challenges. It has the same issue of manual back and forth that causes too much friction. The same issues apply, and I think a similar solution can be considered.
+* **Solving for other deals:** Besides the YC deal, PostHog will continue to set up other deals with other accelerators / VC programs. It would be smart to think of this as a workflow we can replicate and edit the minor details for emerging deals.
+* **Solving for internal reporting:** Because so many systems are talking to one another here, there is a lot of manual reporting between Zapier tables, Stripe data, Salesforce, Vitally and PostHog to get a source-of-truth view of how our startups programs and YC adoption is doing. I think if we can find ways to let the systems do the talking, we should hopefully result naturally in cleaner data, but if we can undertake this job with this in mind, we'd be even better off. 


requests-for-comments/2025-01-29-startups-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+- I think the system should **auto-filter applications:** if they are not eligible, do not apply the credits. *(We can write nice rejections: “Did we get that wrong? Contact Scott if you think there’s been a mistake. We like being generous, but sometimes the our auto-bots are fierce defenders of the realm.”)*
+- I think the system should **auto-approve people:** if they pass the above criteria check, the credits should simply be applied. We can be alerted to abuses if needed, starting with daily manual reviews. 
+
+## Proposed flow:
+ * Remove the actual application form fields (just the form fields) from [the page](https://posthog.com/startups/apply)
+* Replace it with a call-to-action to “Sign in or sign up to apply”
+* Use PostHog-the-app to filter by whether a user has clicked that or not, and show a banner that says “🎉 You’re eligible for the startup plan! Get started here.”
+* Build a new page on the app: /project/###/startups
+* Host the completion steps of the form here. 
+* * All we need to ask for is what we don't already know. I think that's just "How much in total funding have you raised (USD)" and "The date that your company was incorporated."
```
timgl, line 31: I guess to make things easier we can automatically apply the credits immediately, but then asynchronously check whether they're actually allowed to get them and reverse if not?

requests-for-comments/2025-01-29-startups-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+## How the current flow works:
+1. People fill out the forms on https://posthog.com/startups 
+2. An email (via Zap) goes out from Scott Lewis, asking participants to manually submit their info
+3. Participants manually submit, over email, the date they were founded, their project ID, last 4 digits of their CC, amount they've fundraised, the email address associated with their PostHog account
+4. Over email, Scott receives this, manually reviews the info across Stripe and Vitally to see if they are eligible, and to check whether they have the credits or not.
+5. He then presses Approve in Zapier (which applies the credits in Stripe, and triggers the next workflow of emailing the person with their update + merch stuff)
+
+## What can we do about it? 
+- **Steps 2 and 3,** above, should not have to happen. (We already have a record of *most* of the info, because a PostHog account is required to apply.)
+- I think this means the actual **sign-up form should be gated**, accessible from the app itself — requiring a user to be in a logged-in state to submit.
```
joethreepwood, line 21: This is actually a brilliant solution if we can find a way to make it work within the app. 

Do you have an idea on where it would sit? Billing page would seem a good choice, to me. We could also use this as an opportunity to implement [the billing badges](https://github.com/PostHog/posthog.com/issues/9498) I proposed a while ago and which I think @raquelmsmith and @patricio-posthog have been hoping to ship this quarter too.

joethreepwood: We'd definitely want to keep the existing /startups page on PostHog.com, right? We can just have the 'Apply now' button direct to the relevant area of the app and cut the form.

requests-for-comments/2025-01-29-startups-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+ * Remove the actual application form fields (just the form fields) from [the page](https://posthog.com/startups/apply)
+* Replace it with a call-to-action to “Sign in or sign up to apply”
+* Use PostHog-the-app to filter by whether a user has clicked that or not, and show a banner that says “🎉 You’re eligible for the startup plan! Get started here.”
+* Build a new page on the app: /project/###/startups
+* Host the completion steps of the form here. 
+* * All we need to ask for is what we don't already know. I think that's just "How much in total funding have you raised (USD)" and "The date that your company was incorporated."
+* I get a little lost in the plumbing after this and am asking for support thinking through the path.
+
+## Three other considerations:
+* **Solving for the YC program too:** It needs similar care, as the form at [yc-onboarding](https://posthog.com/yc-onboarding) generates all the same challenges. It has the same issue of manual back and forth that causes too much friction. The same issues apply, and I think a similar solution can be considered.
```
joethreepwood, line 35: The slight difference here is that we need to check that the person is in the YC program by their Bookface access, and we _really_ want them to hit 'Using this deal' on our Bookface page. However, maybe we can have that be part of a white-glove service handled by the proposed YC Implementation hire.

requests-for-comments/2025-01-29-startups-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+* **Solving for other deals:** Besides the YC deal, PostHog will continue to set up other deals with other accelerators / VC programs. It would be smart to think of this as a workflow we can replicate and edit the minor details for emerging deals.
+* **Solving for internal reporting:** Because so many systems are talking to one another here, there is a lot of manual reporting between Zapier tables, Stripe data, Salesforce, Vitally and PostHog to get a source-of-truth view of how our startups programs and YC adoption is doing. I think if we can find ways to let the systems do the talking, we should hopefully result naturally in cleaner data, but if we can undertake this job with this in mind, we'd be even better off.
```
joethreepwood, line 37: @minekansu  I think you're the expert of the Zap and Vitally side of things, so just tagging you in case you have Thoughts

joethreepwood (approved):
I love this idea!

requests-for-comments/2025-01-29-startups-workflow.md
zlwaterfield: +1 - this credits won't be used for at least 30 days so we have time to roll back from bad actors.

requests-for-comments/2025-01-29-startups-workflow.md
minekansu: If the application info will be in posthog or stripe then we can sync vitally and salesforce directly without using zapier which would make it easier to maintain 💯 instead of zaps we can just use posthog destinations or if this info will be in stripe metadata we can sync from there

a couple other things that came to mind:
- we probably need a step ensuring applicants add a credit card before approval otherwise we cannot add credits, should this be a blocking step?
- right now we check clearbit data for each applicant to verify their funding raised and incorporation date. If their clearbit data does not meet our criteria we send an automatic follow-up even if the information they entered in the form does meet our criteria. Clearbit data isn’t always available but if a company appears in clearbit it’s usually a signal that they might be a larger company, probably helpful to have this data somewhere for the manual review process.

requests-for-comments/2025-01-29-startups-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+- I think this means the actual **sign-up form should be gated**, accessible from the app itself — requiring a user to be in a logged-in state to submit. 
+- I think the system should **auto-filter applications:** if they are not eligible, do not apply the credits. *(We can write nice rejections: “Did we get that wrong? Contact Scott if you think there’s been a mistake. We like being generous, but sometimes the our auto-bots are fierce defenders of the realm.”)*
+- I think the system should **auto-approve people:** if they pass the above criteria check, the credits should simply be applied. We can be alerted to abuses if needed, starting with daily manual reviews. 
+
+## Proposed flow:
+ * Remove the actual application form fields (just the form fields) from [the page](https://posthog.com/startups/apply)
+* Replace it with a call-to-action to “Sign in or sign up to apply”
+* Use PostHog-the-app to filter by whether a user has clicked that or not, and show a banner that says “🎉 You’re eligible for the startup plan! Get started here.”
```
zlwaterfield, line 28: We don't really have this data available in app so maybe it's available to everyone somewhere on the billing page?

requests-for-comments/2025-01-29-startups-workflow.md
```diff
@@ -0,0 +1,37 @@
...
+* Build a new page on the app: /project/###/startups
+* Host the completion steps of the form here. 
+* * All we need to ask for is what we don't already know. I think that's just "How much in total funding have you raised (USD)" and "The date that your company was incorporated."
+* I get a little lost in the plumbing after this and am asking for support thinking through the path.
```
zlwaterfield, line 32: Assuming we submitted for form directly here and they met the requirements, we could do the things that the current "Approve" button in Zapier does
- apply credits in stripe
- apply metadata in stripe
- send an email

Areas that are trickier 
- where do we track this so Scott or Kevan can review it for false positives? 
- this is sorta standardized so we wouldn't know whether or not to at the extra $5k for our partners
- how can we reply on the submitted data? will Scott/Kevan still manually review it all? Or do we enrich it someone and check against that data?

requests-for-comments/2025-01-29-startups-workflow.md
zlwaterfield: I agree I don't know if doing this for YC is the best path because then we wouldn't be able to confirm their batch and if they are in YC. 

We could have a custom url that is linked from the deal which renders a version of the form just for YC? Downside is they may not always go through the YC deal link if it's accessible elsewhere. How can we demonstrate you get more via the deal then just normal startups?

zlwaterfield (commented):
I spent some time with @minekansu today. 

Overall I think this is doable but a few things we need to get answered first.
1. How do we apply custom credits for partners automatically (the extra $5k)?
3. Where do we send the data after approval so it can be verified? Do we enrich it with clearbit?
4. How do we verify YC companies? And make sure they are clicking from the deal?

Side note:
- we need to update the bookface deal, some of the language it out of data (i.e. A/B testing), there are missing products (i.e. Errors and LLM Oberservability)
- as discussed here: https://posthog.slack.com/archives/C075D3C5HST/p1730305630450429, we should just remove the $25k deal for YC startups and do $50 for all of them to simplify things. We need to update the Bookface deal and the YC landing page. 

Screenshot 2025-03-05 at 11 44 47 AM

requests-for-comments/2025-01-29-startups-workflow.md
kevan-g: Zach, loving your thoughts here, and I've booked time for us to chat tomorrow.

**On reviewing applications**
I have recently introduced a new question on the automated email that asks people to write a bit of a blurb on what they are building, and provide their landing page URL. This is because our partners at Digital Ocean were detecting some cases (not many) of security/fraud risk, and we wanted to make sure we had good players. But it also does something valuable for PostHog: encourages applicants to use the human touch (as we do in our job applications) to show they are serious, real, they want this, and there is at least a SLIGHT bar. Not a high one. 

All that to say, **keeping** the manual review might not be a terrible idea. But removing the **steps** it takes for us to _approve_ (zapier > vitally > stripe id > zapier > click button) -- that seems really nice. 

**$5k for our partners**
I actually haven't come across this in my workflows. Who gets the extra $5k?

kevan-g:
> 1. How do we apply custom credits for partners automatically (the extra $5k)?

(I'm less familiar with the extra $5k deal, but would like to learn what that's about)

> 2. Where do we send the data after approval so it can be verified? Do we enrich it with clearbit?

Great question; I'm open. A zapier table, spreadsheet, internal dashboard, email digest...if the steps of _approval_ are simplified, the job of _reviewing_ becomes a little easier. (I totally don't mind some aspect of manual review, especially with the new "application-style" question I mentioned in my other note.)

> 3. How do we verify YC companies? And make sure they are clicking from the deal?

A custom URL could work here.
I wonder about a promo code system.
Presently we ask for a bookface screenshot. 
Supabase asks for an intro/welcome call as part of the onboarding. 

> * we need to update the bookface deal, some of the language it out of data (i.e. A/B testing), there are missing products (i.e. Errors and LLM Oberservability)

Awesome; I have access to edit that directly now. Let's chat about what we want to see.

pawel-cebula:
I've started looking into this today more closely and wanted to check a few things before I suggest a revised plan:

1. How do we currently deal with rejections (if someone doesn't qualify for the program)? Do we just keep those in the Zapier table without approval or is that marked somewhere and/or communicated to the users?
2. Is this list of places where we need the startup/YC program application/approval data to be available exhaustive: `billing`, `posthog`, Stripe, Salesforce, Vitally?
3. How is this data used in Salesforce and Vitally? AFAIK we initially just create contacts. Is further usage based on cases like credits expiring/being almost used up so that we can notify them and/or the sales team can approach those accounts?
4. Is upgrading to paid plan a requirement for YC program too? I assume so but just double-checking since it wasn't clear on the YC page.
 
cc @scottlewis-ph @minekansu @joethreepwood since you might have the most context on this

pawel-cebula:
I've started working on this and wanted your thoughts on the implementation so far.

My suggestion is to first migrate the form from posthog.com to the app and connect it to the existing infra (Zapier table) - and make some minor tweaks to existing automations, e.g. don't ask for data we already have via email. 

After that we can improve the data enrichment and decide if we want to keep Zapier table as the main data store and interface for approval reviews etc.

📹 Loom: https://www.loom.com/share/2652c18e0a75449a801037b1d90d57a3

_(ignore the design and copy for now, I'll later improve it and adjust it to match existing forms more closely)_

- `posthog.com` pages (https://posthog.com/startups and https://posthog.com/yc-onboarding) will link to the new form in the app (e.g. `https://us.posthog.com/startups` or `https://us.posthog.com/startups/yc`)
- On the migrated form
  - We'll check if the user has an active subscription and prompt them to upgrade otherwise
    - Currently links to billing page but we can probably use the in-app checkout that @zlwaterfield is currently testing to upgrade without leaving this page?)
  - We'll use available info for most data we send to webhook (e.g. name, email, domain name, org name/id, Stripe customer_id) and won't show those fields
  - We'll only ask for things that we don't know
    - If they're building with LLMs - required (wondering if we need this since 75%+ already do and AFAIK @joethreepwood is incorporating some stuff around this into emails?)
    - Non-YC only
      - Incorporation date - required (for non-YC), checks if > 2 years ago, shows error and blocks submission if 
      - Amount raised - required (for non-YC), checks if > $5m, shows error and blocks submission
    - YC only    
      - YC batch (YC only) - required
        - After they pick the batch we query https://github.com/yc-oss/api to check (based on name and domain)
      - Bookface screenshot (YC only) - if we couldn't verify cohort membership via API, we ask them to upload a screenshot of their Bookface profile showing they use PostHog
  - We won't ask a few things anymore
    - Company name - were asking for it in addition to organization name but I don't think we need it?
    - Who referred you - we should instead profile links with a query param (we could also use to e.g. show the partner logo)
    - "Anything to add" - doesn't seem to add a ton of value
- These changes will provide use with all the required info upfront
  - Reduce the founder workload, e.g. less things to fill in, not needing to provide same info over email, less requests for more info
  - Reduce our (especially Scott's) workload, e.g. checking missing customer_id, org name or email mismatches, responding to emails
  - Shorten the time from applying to approval and getting credits
    - We could eventually just auto-approve it and grant credits on successful submission in all cases and just review them all once a week, which could be semi-automated if we find a better way to enrich this data (mainly incorporation date and founding) and only flag those where there's a mismatch etc.
- Right now all the logic is on the client side so this is very far from foolproof but it might be sufficient for now since we'll be verifying the submissions and approvals after the fact to detect abuse
- Next steps could include
  - Getting this to work as expected and improving design/copy (in line with existing forms)
  - Reviewing automated emails that ask for more info (avoid asking for things we have) 
  - Connecting the new form to Salesforce webhook and to the Zapier table directly (currently we sync from Salesforce to Zapier table)
  - Figuring out better ways to enrich data to validate incorporation date and funding raised
  - Deciding if we stick to Zapier tables as source of truth and/or interface for reviewing approvals etc)
  - Automating approvals and credit assignment and switching just to retrospective reviews

cc: @minekansu @Scott-Sowers-PostHog @joethreepwood @zlwaterfield

zlwaterfield:
This is amazing.

- We can definitely use the in app payment entry modal for this - LMK if you want a hand integrating it.
- Love the usage of the YC API to auto validate + the customer URL that we can link to from Bookface. 
- Agree we can probably remove the LLM question

> Who referred you - we should instead profile links with a query param (we could also use it to e.g. show the partner logo)
- This is a good idea, we'd need to store the lookup and images but would allow us to customer the page a little without much effort. Curious to hear Joe's thoughts on this. 
- For next steps - how much do you want to do on the first release vs improve over time? I think we should be auto approving and applying the credits right away (that's a big reason we are doing this) but moving away from Zapier and such could be done down the line.

joethreepwood:
> Agree we can probably remove the LLM question

100%. We don't use this for anything. 

> Zapier

The dream for me would be to cut Zapier out entirely and use PostHog events to trigger messages (in Customer.io), update Salesforce + Vitally (via CDP) and so on. One gap there is merch via Brilliant, which I don't think we have an integration with specifically. But if we have to rely on Zapier for now then 🆗 

This looks really great to me though!

scottlewis-ph:
this will be such a game changer! incredibly exciting to see this v1 demo

a few nice-to-haves that would solve a few extra downstream problems:
- a _'i'll need another merch pack for my co-founder_' option for the YC applications as we get lots of additional requests just for this. or we could maybe scrape if they have a cofounder from their bookface company page?
- some YC founders email me to get re-invited to the slack channel if the link has expired. they know my email from zaps/comms i've had with them. i'm happy to continue re-inviting manually, but with this new format i'm conscious they wouldn't know who to contact unless:
  - in a _'you've been accepted onto the deal'_ confirmation email, we include a line like _'message yc@posthog.com for any extra support'_ of _'reply to this email'_ to make it clearer
   - or include an in-app _'add me to the slack channel'_ button
- occasionally, startup & yc founders have merch problems they want to follow up on, or might have not received the merch link in the first place. similar point as above where we should signpost them to the right contact details for help (maybe merch@posthog.com?). otherwise they'll be raising a ton of CS tickets just about merch
- if we automate approvals, it would be good to be able to turn this off/deactivate this easily on demand? at the minute, for example, i've paused approving YC companies for a few days as our backpacks are out of stock. wouldn't want to auto-approve everyone and create a problem for ourselves in situations like these so we need a kill switch for rare things like this

pawel-cebula:
> We can definitely use the in app payment entry modal for this - LMK if you want a hand integrating it.

Will do!

> This is a good idea, we'd need to store the lookup and images but would allow us to customer the page a little without much effort. Curious to hear Joe's thoughts on this.

There are some company logo APIs that we could ideally use to avoid having to deploy anything, will add it to next potential steps.

> For next steps - how much do you want to do on the first release vs improve over time? I think we should be auto approving and applying the credits right away (that's a big reason we are doing this) but moving away from Zapier and such could be done down the line.

I was initially planning to improve enrichment first before we start auto-approving but maybe that's too conservative. We'll want to review them all anyway. If we see there's too many fraudulent cases early on we could turn it off (via feature flag).

> The dream for me would be to cut Zapier out entirely and use PostHog events to trigger messages (in Customer.io), update Salesforce + Vitally (via CDP) and so on. One gap there is merch via Brilliant, which I don't think we have an integration with specifically. But if we have to rely on Zapier for now then 🆗

Very helpful to understand the dream scenario! I'll definitely keep this in mind in the next iterations and might ping you to understand some of these better. But will aim to ship the form migration first ASAP to "stop the bleeding" and then look at this more closely.

> a 'i'll need another merch pack for my co-founder' option for the YC applications as we get lots of additional requests just for this. or we could maybe scrape if they have a cofounder from their bookface company page?

Can you remind me if there was an upper limit for the merch packs? Adding a field for this would be straightforward. It might encourage more people to request for all co-founders proactively though. We could see if it changes the utilization dramatically and remove if needed after. 

> in a 'you've been accepted onto the deal' confirmation email, we include a line like 'message [yc@posthog.com](mailto:yc@posthog.com) for any extra support' of 'reply to this email' to make it clearer (...) or include an in-app 'add me to the slack channel' button
(...)
similar point as above where we should signpost them to the right contact details for help (maybe [merch@posthog.com](mailto:merch@posthog.com))?

We can definitely add these to emails to direct their queries to the right inbox - do these two already exist and are actively managed?

> if we automate approvals, it would be good to be able to turn this off/deactivate this easily on demand? at the minute, for example, i've paused approving YC companies for a few days as our backpacks are out of stock. wouldn't want to auto-approve everyone and create a problem for ourselves in situations like these so we need a kill switch for rare things like this

Makes sense! We could use a feature flag for this so that you could turn it off at any time.

scottlewis-ph:
> Can you remind me if there was an upper limit for the merch packs? Adding a field for this would be straightforward. It might encourage more people to request for all co-founders proactively though. We could see if it changes the utilization dramatically and remove if needed after.

no hard limit on merch packs, but most yc founders have at least one co-founder. the more people getting their hands on the merch in each batch, the better, so i think it's not the end of the world when applicants are trigger happy. let's include the choice for an extra pack, and we can hit undo if people take advantage

> We can definitely add these to emails to direct their queries to the right inbox - do these two already exist and are actively managed?

Yep both are distribution lists so someone can always jump in and respond

minekansu:
> Could you also take a quick look at the Looms, tables and Zaps to see if it generally makes sense since you've worked on the previous iterations?

Can you check if the tables are shared? i'm getting a weird error (I can access old tables fine)
Screenshot 2025-04-15 at 10 03 54 AM


> What would be the easiest way to set up a sync from Salesforce to Zapier tables to add Clearbit and Clay enrichment data?

we create a salesforce contact automatically for every sign up and send to enrichment so most of these should exist as enriched contacts in salesforce. if you add a zap step similar to steps 10 and 11 in [this zap](https://zapier.com/editor/249925520/published/249925531/fields) it should work! if we find that this isnt sufficient we can set up a new clay table for startup of enrichment

> I've not enabled the existing Salesforce contact Zap yet in production code (see [here](https://github.com/PostHog/billing/blob/8f0a2d883b99d690ebd68d96a64a3063f10cc235/billing/api/startups.py#L144-L168)) - could you check if what I'm planning to send there makes sense and if I should anything else that's available?

ah great these are the exact salesforce fields we would want for these applicants:
```
Email
Posthog_Org_ID__c
Record_Source__c : YC
YC_Batch__c
StartupDomain__c
Incorporation_Date__c
PostHog_organization_name__c
Raised__c
```

pawel-cebula:
> Can you check if the tables are shared? i'm getting a weird error (I can access old tables fine)

Oops indeed the tables weren't shared - fixed now. The Zaps should be accessible as they're in the shared folders.

> we create a salesforce contact automatically for every sign up and send to enrichment so most of these should exist as enriched contacts in salesforce. if you add a zap step similar to steps 10 and 11 in [this zap](https://zapier.com/editor/249925520/published/249925531/fields) it should work! if we find that this isnt sufficient we can set up a new clay table for startup of enrichment

Got it, so since they need to sign up before submitting the form, by the time they submit it, they'll already be in Salesforce and IIRC Clearbit enrichment is almost instant so it'll be there at the time. But in case someone signs up and submits a form shortly after, Clay data might not be there yet (since it takes up to 24h)?

If that's the case, the simplest option might be to do this step once right away and then again 24h after?

> ah great these are the exact salesforce fields we would want for these applicants:

I was looking at using the [Website to Salesforce Zap](https://zapier.com/editor/248748195/draft) and there I was looking at the property names used in the various steps (mostly `Create Contact`) - they seem to be pointing to the first step (`Catch Hook`) and had different spelling than the one above. I assume I should align them with those?

minekansu:
yep i can see the tables!

> If that's the case, the simplest option might be to do this step once right away and then again 24h after?

yep that's correct, if they sign up and apply right after it'll only have clearbit data (that's instant) for clay we have to rerun

> they seem to be pointing to the first step (Catch Hook) and had different spelling than the one above. I assume I should align them with those?

yea honestly I'm not sure, the ones i posted are api field names that i took from salesforce, maybe try yours first and switch if it doesnt work?

pawel-cebula:
@minekansu Here are zap runs where I've created two Salesforce contacts:
- [YC](https://zapier.com/editor/248748195/run/0087bd10-324c-a142-70d0-c5a1e167b241/251092965/run-details?sidebar=runs&filterRun=true)
- [Startups](https://zapier.com/editor/248748195/run/0087bd10-677c-af53-9923-87f46d4c0051/248748195/run-details?sidebar=runs&filterRun=true)

I'm now trying to add a step to the [PostHog for Startups - Form to Table](https://zapier.com/editor/291009790/draft) zap that will get the Salesforce account and use the Clearbit/Clay data but I'm running into some issues
- I'm trying to use PostHog Org ID and account name for search but it's not finding it
- Looking at the [Website to Salesforce](https://zapier.com/editor/248748195/run/0087bd10-677c-af53-9923-87f46d4c0051/248748195/run-details?sidebar=runs&filterRun=true) zap, in the startups path it tries to find a record (account) using account name but it doesn't create one if it doesn't exist
- It then creates a contact, populating `PostHog organization name` field but
  - It doesn't set PostHog org ID (it doesn't appear as a possible field on a contact)
  - There are two other `PostHog Org` fields
- I tried re-running step 5 of [PostHog for Startups - Form to Table](https://zapier.com/editor/291009790/draft) later manually, in case there was a race condition, but it's still not finding an account based on the name
- So if I understand correctly, an account doesn't actually get created in this path? Same applies to the YC path.
- I do see that the old [Salesforce contact to table](https://zapier.com/editor/249925520/published/249925524/fields) zap is based on contact's account ID and e.g. in this [run](https://zapier.com/editor/249918248/run/0087bd10-1d94-ac7c-076e-44f5e6411071/249918248/run-details?sidebar=runs&filterRun=true) I do see account ID on a new contact in step 1 but maybe this account already existed with some other name? What's the logic for creating an account and matching a new contact to existing account?
- I though I could fetch a contact instead but it doesn't seem possible in Salesforce zaps/actions, plus it seems that contacts don't have the relevant Clearbit/Clay fields
- Possible fixes I can think of so far
  - A) Modify step 7 in [Website to Salesforce](https://zapier.com/editor/248748195/run/0087bd10-677c-af53-9923-87f46d4c0051/248748195/run-details?sidebar=runs&filterRun=true) to use PostHog Org ID to find a record and create an account if not found - feels cleanest in terms of separation of concerns and potentially fixes an issue with missing out on creating new accounts (unless that's intended or happens in the background already)
  - B) Connect or merge the two zaps ([Website to Salesforce](https://zapier.com/editor/248748195/run/0087bd10-677c-af53-9923-87f46d4c0051/248748195/run-details?sidebar=runs&filterRun=true) and [PostHog for Startups - Form to Table](https://zapier.com/editor/291009790/draft)) so that we can use the account ID returned after creating new contact to then find the record - I'd prefer not to mix them as they both have different purposes
  - C) Add a step to [Website to Salesforce](https://zapier.com/editor/248748195/run/0087bd10-677c-af53-9923-87f46d4c0051/248748195/run-details?sidebar=runs&filterRun=true) zap so that after creating a contact we use that to find the account in salesforce and then find a record in a table and update it (with a delay of e.g. 5min to make sure table record was created) - feels better than B but still mixes things between these two zaps - if A is not feasible, I'd lean towards this one

minekansu:
> * I'm trying to use PostHog Org ID and account name for search but it's not finding it

just org id should work! field name on contact object is Posthog_Org_ID__c. if you create a contact we have a separate flow within salesforce that creates an account attached to the contact (if it doesnt exist already)

> * Looking at the [Website to Salesforce](https://zapier.com/editor/248748195/run/0087bd10-677c-af53-9923-87f46d4c0051/248748195/run-details?sidebar=runs&filterRun=true) zap, in the startups path it tries to find a record (account) using account name but it doesn't create one if it doesn't exist
> * It then creates a contact, populating `PostHog organization name` field but
>   
>   * It doesn't set PostHog org ID (it doesn't appear as a possible field on a contact)
>   * There are two other `PostHog Org` fields
> * I tried re-running step 5 of [PostHog for Startups - Form to Table](https://zapier.com/editor/291009790/draft) later manually, in case there was a race condition, but it's still not finding an account based on the name
> * So if I understand correctly, an account doesn't actually get created in this path? Same applies to the YC path.

contact to account creation is a separate flow within salesforce. we create the contact, then have a separate flow that checks if an account with that domain exists. if yes links contact to that id, if no creates a new contact and passes org id from contact to the newly created account.

> * I though I could fetch a contact instead but it doesn't seem possible in Salesforce zaps/actions, plus it seems that contacts don't have the relevant Clearbit/Clay fields

they do but it may not be super obvious from the name, see last column in [this clay table](https://app.clay.com/workspaces/268768/workbooks/wb_gk4Xv8p2nvvH/tables/t_guuzJZCqQrh3/views/gv_yjWwigwdaVFk)
for contact enrichment mapping and [this one](https://app.clay.com/workspaces/268768/tables/t_7iMf8fsecn9y/views/gv_jZx6ZGRNjipM) for account mapping from clay to salesforce (account one seems to have stopped bc the table got too big, I'll fix that today and backfill)

> * Possible fixes I can think of so far
>   
>   * A) Modify step 7 in [Website to Salesforce](https://zapier.com/editor/248748195/run/0087bd10-677c-af53-9923-87f46d4c0051/248748195/run-details?sidebar=runs&filterRun=true) to use PostHog Org ID to find a record and create an account if not found - feels cleanest in terms of separation of concerns and potentially fixes an issue with missing out on creating new accounts (unless that's intended or happens in the background already)

if we just create a contact for the submitted form and make sure to populate `Posthog_Org_ID__c` on contact object salesforce will handle the rest. we can then run another zap that gets triggered every time a new contact created that has source set as startup and YC and pull the enriched fields from salesforce (we can add a 5 min delay for first round, and a 24 hour delay for the post-Clay enrichment)

pawel-cebula:
> contact to account creation is a separate flow within salesforce. we create the contact, then have a separate flow that checks if an account with that domain exists. if yes links contact to that id, if no creates a new contact and passes org id from contact to the newly created account.

Ah that explains it - I was testing with my PostHog email so it must've got matched to an existing account based on the domain, so when I tried searching via org ID it didn't find it as a new account wasn't created.

I suppose avoiding creating a new account if there's an existing one with same domain makes sense from sales perspective, even if there might be a different org in PostHog. We just won't be able to enrich automatically since the org IDs won't match and we won't find the account when trying to pull enrichment data. But that should be fine since it's unlikely this issue will apply for startups and YC companies that have just started recently.

Alternatively, we could find account based on the domain? I'm getting a weird result when I do that with `posthog.com`, with an address in Ukraine, but I suppose it should work for actual users with non-public email domains?

image

image

> if we just create a contact for the submitted form and make sure to populate Posthog_Org_ID__c on contact object salesforce will handle the rest. we can then run another zap that gets triggered every time a new contact created that has source set as startup and YC and pull the enriched fields from salesforce (we can add a 5 min delay for first round, and a 24 hour delay for the post-Clay enrichment)

Which of the fields would it be in contact creation action? I've tried refreshing it but still not seeing org ID. Is it one of those named `Posthog Org`?

image

> [this one](https://app.clay.com/workspaces/268768/tables/t_7iMf8fsecn9y/views/gv_jZx6ZGRNjipM) for account mapping from clay to salesforce (account one seems to have stopped bc the table got too big, I'll fix that today and backfill)

Is it possible to enrich it with founding date (rather than just year) and funding raised?

minekansu:
> I suppose avoiding creating a new account if there's an existing one with same domain makes sense from sales perspective, even if there might be a different org in PostHog. We just won't be able to enrich automatically since the org IDs won't match and we won't find the account when trying to pull enrichment data. But that should be fine since it's unlikely this issue will apply for startups and YC companies that have just started recently.

this is a great point actually which makes me think lets create a new org if there isnt one with matching org id. potential downsides 1. contact mapping may be off but thats something we can fix retroactively (we run de-duplication scripts every saturday so we can update the logic there) 2. enriching same company multiple times and paying clay for it (startup volumes are low so i dont think this is a concern atm). this might be better for other connections to zendesk and vitally as well as some of them rely on org ids
 
> Alternatively, we could find account based on the domain? I'm getting a weird result when I do that with `posthog.com`, with an address in Ukraine, but I suppose it should work for actual users with non-public email domains?

domains also definitely work but you convinced me creating new accounts is a better solution :)
 
> Which of the fields would it be in contact creation action? I've tried refreshing it but still not seeing org ID. Is it one of those named `Posthog Org`?

this is the right one. looks like the other field was created by our contractor by mistake, i asked them to review and remove, thanks for catching that!
 
Screenshot 2025-04-18 at 10 32 33 AM


> Is it possible to enrich it with founding date (rather than just year) and funding raised?

should be doable, taking a look at this now!

minekansu:
looks like clay enrichment only has founded year not exact date (we can be a bit generous here with the exact cutoff i think). i've added `total_funding_raised__c ` to salesforce and clay, running backfill enrichment now including this info, should be done in 2-3 hours!

pawel-cebula:
Awesome thanks! On my end I'll make sure to pass the org ID when creating contact and then will use that org ID to find the account and pull clearbit/clay data from it.

Agreed that we can be generous on the founding year - enrichment will mainly help to catch obvious discrepancies between submitted info and enriched data.

pawel-cebula:
@minekansu 

Here are the fields I'm using for Clearbit (copying from [here](https://zapier.com/editor/249925520/published/249925530/fields))
image

For `Clay Raised` I'm using `Total Funding Raised C` but not sure which one to use for `Clay Founded`, as the only one that comes up (`Founded Year C`) is already used as the Clearbit one above
image

minekansu:
current data flow is as follows:

Clearbit instant data fill populates `Founded_year__c` and `Clearbit_Raised__c` 
Clay updates `Founded_year__c` if clearbit was blank and updates `total_funding_raised__c`

pawel-cebula:
@minekansu Just wanted to clarify if the new account creation on Salesforce (if there's no accounts with matching posthog org ID) is something that you'll be implementing on your end (in those existing contact/account creation Salesforce automations) or if it should be done via the Zap that tries to find the account using posthog org ID?

The former would probably be best so that this logic is in one place? 

But if it's the latter, could you take a look [here](https://zapier.com/editor/291009790/draft/293525888/fields) and edit step 5 to populate the fields in Salesforce that we want?
- Tick `Create Salesforce Record if it doesn’t exist yet?`
- Review the fields we want to populate when creating an account - I picked some but not sure which ones we want

image

minekansu:
@pawel-cebula done! added this salesforce flow to create new org if there are no existing orgs with matching org id at time of contact creation https://posthog.lightning.force.com/builder_platform_interaction/flowBuilder.app?flowId=301Vx00000Is6m3IAB

PR #292, opened 2025-02-20
RFC: capture to disk

frankh:
I've been banging on about this long enough i figured i should write an RFC and we can make a decision one way or the other


requests-for-comments/2025-02-20-capture-to-disk.md
```diff
@@ -0,0 +1,49 @@
...
+
+Pros:
+1. Massively reduced latency - under 10ms P50 easily
+2. No Single Point of Failure
+   - if Kafka is down we still write events. Each pod can have a disk large enough for e.g. an hours worth of traffic, giving us time to spin up an MSK cluster, or warpstream, or whatever we can to fix things.
+   - short-lived blips in Kafka are invisible to the user. e.g. patching, or a warpstream pod dying etc.
+
+Cons:
+1. Durability - durability will necessarily be reduced. There is _some_ chance of a disk failure that would lead to events we have confirmed being lost. This is extremely unlikely as disks are constantly being flushed, but it is possible. We could potentially use RAID to mitigate this.
+2. Ordering - we can no longer guarantee ordering between 2 events that are sent in quick succession. A client could receive a 200 from capture, send a followup event, receive a 200 but from a different pod. These could flush out of order
```
MichaelKutsch-ph, line 26: we have this issue already imho because of the nature of a distrbuted async system. I don't know how likely it is that this happens as this is usually tied on end-user-actions, right?

requests-for-comments/2025-02-20-capture-to-disk.md
```diff
@@ -0,0 +1,49 @@
...
+
+The sqlite database will live in a persistent volume on the capture pod - this will have a reclaim policy of **RETAIN** (important!). This means in the case of a pod dying, or the underlying node stopping responding, the persistent volume will remain untouched. We can then have a service which periodically scans for orphaned disks and flushes them to Kafka.
+
+In the capture pod's shutdown handler we can have it flush the database and if it successfully flushes all the events it will update its volumes reclaim policy to **DELETE**, so we don't have hundreds of volumes lying around in normal circumstances.
+
+## Alternatives considered
+
+We recently implemented a "fallback to S3" failure mode for capture - the idea being that if an event fails to write to Kafka we write it to S3 instead
+
+This is more of an emergency fallback and it has significant drawbacks, especially in the case of short-lived blips or slowdowns in Kafka, where we may end up with handfuls of events in S3 that end up out of order.
```
MichaelKutsch-ph, line 47: I would be very interested in the rationale of the requirement for ordering. 

***assumption area***
I'm not very deep in this topic tbh, but as far as I understand it it's mostly because of "we want to get the last info on persons". 

If this logic is changed to full idempotency (meaning, we only persist "latest" based on timestamp?) the order would not matter. 
I think this would be required anyhow because how would you re-process older events in case of s3 failover
***end assumption area ***

requests-for-comments/2025-02-20-capture-to-disk.md
```diff
@@ -0,0 +1,49 @@
...
+In the capture pods we should simply write the events to disk and return a 200 immediately, then asynchronously persist them to Kafka.
+
+Pros:
+1. Massively reduced latency - under 10ms P50 easily
+2. No Single Point of Failure
+   - if Kafka is down we still write events. Each pod can have a disk large enough for e.g. an hours worth of traffic, giving us time to spin up an MSK cluster, or warpstream, or whatever we can to fix things.
+   - short-lived blips in Kafka are invisible to the user. e.g. patching, or a warpstream pod dying etc.
+
+Cons:
+1. Durability - durability will necessarily be reduced. There is _some_ chance of a disk failure that would lead to events we have confirmed being lost. This is extremely unlikely as disks are constantly being flushed, but it is possible. We could potentially use RAID to mitigate this.
```
MichaelKutsch-ph, line 25: pretty sure MSK runs with the same disks :D

frankh: but MSK has replication

oliverb123: ~capture pods could replicate to each other over grpc~

requests-for-comments/2025-02-20-capture-to-disk.md
MichaelKutsch-ph: where and when is this distinct ID generated?

frankh: on the client, a distinct id refers to a single user (so e.g. in the browser, a distinct ID is created and persisted in cookies(?))

requests-for-comments/2025-02-20-capture-to-disk.md
```diff
@@ -0,0 +1,49 @@
...
+
+![Proposed Architecture](./images/capture-to-disk-architecture.png)
+
+I propose we add a sqlite layer to the capture service (open to other data stores, but sqlite is solid and easy to work with).
+
+Events will be written to sqlite and we immediately return a 200 to the client.
+
+In the background we read from sqlite and write batch to Kafka. This has additional advantages of being able to write large batches slowly, making warpstream even cheaper (e.g. we could have a linger of 1000ms, and warpstream batches of 8MB)
+
+The sqlite database will live in a persistent volume on the capture pod - this will have a reclaim policy of **RETAIN** (important!). This means in the case of a pod dying, or the underlying node stopping responding, the persistent volume will remain untouched. We can then have a service which periodically scans for orphaned disks and flushes them to Kafka.
```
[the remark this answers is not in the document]
frankh: yeah, but its a rare situation

the way things are architected at the moment we cannot have both ordering guarantees AND no single point of failure

requests-for-comments/2025-02-20-capture-to-disk.md
MichaelKutsch-ph: Coming to the point of my comment below: 
because only we can guarantee in kafka that events are in order (it could well be that two events are sent almost at the same time with the same distinct id and do not land in the right order in kafka because of latency), we'd have to make sure to only write "latest" already. 

Maybe I'm mistaken (just asking outside perspective questions), what I want to say is: as far as my understanding goes: we _should_ be able to handle this already, to a) handle those edge cases and b) make it possible to reprocess events at any given point in time without producing wrong results in persons, right?

requests-for-comments/2025-02-20-capture-to-disk.md
MichaelKutsch-ph, line 1: What we're seeing here is that we want to use warpstream, but we consider its latency as being too high to work on our scale and to serve our customer demands?

Your proposed solution is very smart, ofc it adds complexity and more moving parts where things could go wrong. 

If latency is our issue, maybe we can re-architect the overall setup?

What about having two small MSK clusters that can hold ~12-24 hrs each of raw capture data ((and everything else that we need that needs to be up public facing but nothing else!), one of em is a hot standby, both of them replicate their data to s3 via kafka connect?
There are solutions out there like [conduktor](https://conduktor.io/blog/when-disaster-hits-kafka-are-you-ready) that provide auto-failover. 

Benefits I see: 
- auto-failover logic in infrastructure, not in code
- MSK outage -> switch to warpstream
- one kafka node outage doesn't take us down, easy upgrades of MSK as two cluster setup
- public facing latency stays as is
- medium risk in terms of tech complexity
- less maintenance of critical software parts as the interfaces for capture stay the same


Downsides
- still having MSK
- costs more than WS only
- added complexity, more hops


While writing this: why aren't we trying to move the highest risk thing first? :D 
Can't we move the processing topics to WS already and reduce the pressure dramatically on MSK?

@benjackwhite @oliverb123 @frankh

MichaelKutsch-ph: We have to move the topics anyhow at some point, but we will de-risk the whole operation

oliverb123: I'm pro capture->msk, everything else to WS, but if we want to move off MSK entirely /eventually/ we should also figure out a way to do fast writes on capture regardless of backing queue - to me, this RFC can be a longer-term thing, not solving an immediate problem (and we can be a bit more out-there with it as a result)

PR #293, opened 2025-02-25
Add RFC for query migrations

thmsobrmlr:
## TL;DR

We wanted a safe way to change the shape of saved insight queries without breaking old ones (like database migrations, but for query JSON). This RFC proposed how, we built it, and the doc now records the outcome: the backend upgrades old queries to the latest schema version whenever it sees them, and the frontend asks the backend to upgrade queries it can't trust (from URLs or notebooks).

## Why

The RFC predates the implementation. It's been updated to mark the chosen option (a) and document what was actually built (`posthog/schema_migrations/`, per-kind `version` field, read-time upgrades, `/query/upgrade` endpoint, insights backfill), so it can be merged as an accurate record.

### DIFF
--- a/requests-for-comments/2025-02-25-query-migrations.md
+++ b/requests-for-comments/2025-02-25-query-migrations.md
@@ -14,10 +14,11 @@ Queries are persisted in various places:
 4. Notebook nodes (and canvas element).
 5. The activity log, when changing a query.
 6. Queries that are converted from `filters` e.g. when users create an insight by directly calling the api.
+7. ...? _(let me know)_
 
-Some of these places are available frontend-side (queries from a url) and some are available backend-side (queries from an insight model).
+Some of these places are available frontend-side (e.g. queries in a url) and some are available backend-side (e.g. queries in an insight model).
 
-This makes it difficult to migrate a query with a not backwards compatibly change e.g. converting a boolean to an enum. Backward compatible changes i.e. adding an optional field can still be done.
+This makes it difficult to migrate a query with a non-backwards compatibly change e.g. converting a boolean to an enum. Backward compatible changes e.g. adding an optional field are no problem.
 
 Any errors in a query cause Pydantic to fail the validation and thus insights with the error will fail to compute.
 
@@ -65,11 +66,15 @@ We could add only a backend side migration and let the frontend call a `/query/v
 
 I'm favoring option a, where we add a backend side migration and call that from the frontend, as it seems to be the easiest (and as such, most scalable) solution for future developers adding query migrations.
 
+### Option d) Something easy I'm not seeing right now?
+
+- Let me know.
+
 ## Implementation
 
 -   Get in touch with customers to deprecate `filters` based insights (already happening). Remove all remaining `filters` conversions.
+-   Implement the above.
 
-tbd:
-
+To be discusssed:
 -   Should we add a version field to the queries, so that we know which migrations to run
 -   Should we keep migrations forever or should we have a system in place that discards frontend side queries that are older than a certain time
--- a/requests-for-comments/2025-02-25-query-migrations.md
+++ b/requests-for-comments/2025-02-25-query-migrations.md
@@ -2,13 +2,9 @@
 
 ## Problem statement
 
-_Who are we building for, what are their needs (include example cases), why is this important?_
-
 We'd like to evolve the [query schema](https://github.com/PostHog/posthog/blob/master/frontend/src/queries/schema/index.ts) while keeping backwards compatibility for existing queries.
 
 ## Context
-
-_What are our competitors doing, what are the technical constraints, what are customers asking for, what does the data tell us, are there external motivations (e.g. launch week, enterprise contract)?_
 
 Queries are persisted in various places:
 
@@ -26,8 +22,6 @@ This makes it difficult to migrate a query with a not backwards compatibly chang
 Any errors in a query cause Pydantic to fail the validation and thus insights with the error will fail to compute.
 
 ## Design
-
-_What are the key user experience and technical design decisions / trade-offs?_
 
 Migrating the schema from one version to the next should be as easy as a database migration for developers.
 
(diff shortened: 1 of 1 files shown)


requests-for-comments/2025-02-25-query-migrations.md
```diff
@@ -0,0 +1,80 @@
...
+- Let me know.
+
+## Implementation
+
+-   Get in touch with customers to deprecate `filters` based insights (already happening). Remove all remaining `filters` conversions.
+-   Implement the above.
+
+To be discusssed:
+-   Should we add a version field to the queries, so that we know which migrations to run
+-   Should we keep migrations forever or should we have a system in place that discards frontend side queries that are older than a certain time
```
Gilbert09, line 80: When you talk about a "migration" here, what does that look like in code? Are we gonna have a history of the model changes over time?

requests-for-comments/2025-02-25-query-migrations.md
```diff
@@ -0,0 +1,80 @@
...
+
+This makes it difficult to migrate a query with a non-backwards compatibly change e.g. converting a boolean to an enum. Backward compatible changes e.g. adding an optional field are no problem.
+
+Any errors in a query cause Pydantic to fail the validation and thus insights with the error will fail to compute.
+
+## Design
+
+Migrating the schema from one version to the next should be as easy as a database migration for developers.
+
+### Option a) Add a backend side migration and allow the frontend to use this in a new endpoint
```
Gilbert09, line 29: I've also experienced a bunch of issues from having conversions on both the frontend and backend, and so I think this makes sense - we'd need to make sure this endpoint was very fast though

requests-for-comments/2025-02-25-query-migrations.md
```diff
@@ -0,0 +1,80 @@
...
+-   Easiest to maintain
+
+#### Cons
+
+-   We can only add optional fields
+-   Technical debt will increase
+
+I'm favoring option a, where we add a backend side migration and call that from the frontend, as it seems to be the easiest (and as such, most scalable) solution for future developers adding query migrations.
+
+### Option d) Something easy I'm not seeing right now?
```
Gilbert09, line 69: Have we considered something like a library/package with the conversion code compiled to both Python and JS? Not sure if something like this exists, but it would help reduce code duplication and things getting out of sync

requests-for-comments/2025-02-25-query-migrations.md
```diff
@@ -0,0 +1,80 @@
...
+
+- Let me know.
+
+## Implementation
+
+-   Get in touch with customers to deprecate `filters` based insights (already happening). Remove all remaining `filters` conversions.
```
Gilbert09, line 75: Big 👍 on getting `filters` removed asap

Gilbert09 (approved):

requests-for-comments/2025-02-25-query-migrations.md
```diff
@@ -0,0 +1,80 @@
...
+
+Any errors in a query cause Pydantic to fail the validation and thus insights with the error will fail to compute.
+
+## Design
+
+Migrating the schema from one version to the next should be as easy as a database migration for developers.
+
+### Option a) Add a backend side migration and allow the frontend to use this in a new endpoint
+
+We could add only a backend side migration and let the frontend call a `/query/validation` endpoint that also migrates queries to newer versions.
```
Twixes, line 31: Since queries are always retrieved from the backend, in most cases we can do something sneakier that avoids the roundtrip: In all of the call sites from the "Queries are persisted in various places" list, we'd do the `/query/validation` transformation internally before the query ever gets to the user. This would mean the frontend _never_ seeing old-style queries.

With one exception: filters in the URL  – where the roundtrip would still be necessary, and would be quite painful (e.g. in some cases the frontend would not be able to support the query anymore, so it would _have_ to wait for the validation round trip – a few hundred milliseconds of a wait doesn't sound so bad, but it can definitely add up to a slow experience!)

requests-for-comments/2025-02-25-query-migrations.md
thmsobrmlr: 💯  I'm thinking it's alright if we have a quick way to return queries that don't need to be touched. We'd then keep backend side queries up-to-date with a backfill job and frontend side queries that are important should be touched (and with that upgraded) often as well.

requests-for-comments/2025-02-25-query-migrations.md
thmsobrmlr: True, we could transpile code from one language to the other or run the migrations inside of a VM. I'm worrying that this is brittle and would blow up bundle size (if using python as source of truth) however.

requests-for-comments/2025-02-25-query-migrations.md
thmsobrmlr: Yup, we definitely want to handle this backend side as much as possible. I'll probably find out how easy that is for notebooks, activity logs, etc when working on it 😆 

We also have queries stored in local storage, so these need to do the roundtrip as well.

orian: Why would migrating the query take few hundred ms? Correct me if wrong, but the `/query/validation` should parse => apply migrations => serialize, it should not take more than 50-100ms?

It would also be easy to cache those.

requests-for-comments/2025-02-25-query-migrations.md
```diff
@@ -0,0 +1,80 @@
...
+
+- Let me know.
+
+## Implementation
+
+-   Get in touch with customers to deprecate `filters` based insights (already happening). Remove all remaining `filters` conversions.
+-   Implement the above.
+
+To be discusssed:
+-   Should we add a version field to the queries, so that we know which migrations to run
```
orian, line 79: We may start to version out API and put the version in each request. This would allow us to discover all situations where someone is calling our trunk API with old app version, whatever reason.

requests-for-comments/2025-02-25-query-migrations.md
Twixes: Validation itself should be quick, but there's also significant overhead to every individual API request, plus internet speeds vary – piece by piece these things add up to a slow app experience, translating to lower retention → less 💵.

PR #300, opened 2025-03-14
Add RFC - Data Warehouse for testing w/ production like data

phixMe:


requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+    - Source containers (Postgres, MySql, Snowflake)
+    - Source destinations (Postgres, MySql, Snowflake)
+    - Clickhouse (Reuse in dev environment as query layer)
+    - Temporal Runner(Reuse in dev environment as query layer)
+- S3
+- GitHub Actions (Invoke when changing subset of our directories, triggering agent)
+
+## Concerns/Open Questions
+- How feasible is it to recreate instances of more hosted type databases, like Snowflake, would something like
+  LocalStack be an option for us?
```
tomasfarias, line 79: I think feasibility will vary from "there's an alternative that kind of works" (Snowflake, Redshift, S3) to "there are no alternatives" (BigQuery).

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+   queries inherited from production.
+
+7. Reverse of #1 to clean up all fragments of delta tables that have been copied into the dev environment and buckets
+   that have. Cleanup fragments in Clickhouse that have been generated by this process as well.
+
+[Data Warehouse Testing Production Flow](./images/dw-testing-production-flo.png)
+
+## Out of Scope
+- Temporal Scheduling Logic - We want to ensure that sync works, not that temporal invokes a schedule at the correct
+  interval.
```
tomasfarias, line 64: At some point, we have to trust our service providers to do what they say they will do. This is rightfully out of scope.

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+- HogQl queries (joins)
+
+> Preface: The main idea behind this strategy is to reverse the direction of our standard syncing procedure in order to
+> determine if we can reconstruct our block storage data (delta → s3) back into the original data store it came from (
+> Postgres, MySql, etc…), and query it once it’s reconstructed as before once it has been put back together.
+
+> Note: The diagram below can be used as an aide.
+
+## Steps
+1. The first step would be to transfer data from a configurable list of sub-buckets under
```
tomasfarias, line 36: How will said "configurable list of sub-buckets" be configured? I assume that we'll need to store the configuration used in each test run to ensure repeatability?

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+3. Run an on demand temporal workflow to test our standard syncing operation as if it’s any other production workflow.
+
+4. The results of these workflows will end up back in s3 via delta as an output. We should be able to run queries at
+   this point.
+
+5. Compare the results of the original copied s3 payloads to the s3 payloads that went through the temporal processor
+   and were synced. This is mostly a parity check to ensure that our syncing process is working as expected.
+
+6. Run test case queries on data that has moved from s3 → external_data_source → s3 either with custom queries or test
+   queries inherited from production.
```
tomasfarias, line 55: How we will select queries for testing? Are we going to store them to ensure repeatability?

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+## Steps
+1. The first step would be to transfer data from a configurable list of sub-buckets under
+   `posthog-s3-datawarehouse-us-east-1` into an ephemeral s3 bucket from a list of deltalake tables from prod → dev
+   environment. Using this approach we can avoid looking at credentials and our Postgres instance. This also means that
+   we have very minimal interactions with production as a whole.
+
+2. Reconstruct stateful instances from the delta tables in our CI flow. This would require different processes for
+   various providers. For example, for MySql we could take an official image, deploy it with ArgoCD, then have a program
+   move the data that has been temporarily copied into our dev s3 bucket into entities we consider external data
+   sources.
```
Gilbert09, line 44: I worry about how well we'll be able to do this. For example, we transform some data during the pipeline and so those transformations won't be tested here. Plus, we're adding an extra layer of logic in the tests that could independently fail (and be fairly costly to maintain imo)

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+common theme is that many are SQL based and our data eventually end up in s3 via delta tables and are queryable via
+HogQl, so that feels like the area to hone our focus on for biggest impact. Other data sources like HubSpot, Salesforce,
+and Chargebee are less generic and also would have a smaller blast radius in the event syncing incremental or full table
+syncing gets regressed. These sources are still important and should be incorporated into the test plan as it matures.
+
+## Higher Level Goals
+- Determinism - We do not want to randomly sample sources/schemas in order to test against. This list of tests should
+  grow over time, but remain consistent for feature development.
+- Repeatability - A failure one run should fail the same way the next run, A success should also be repeatable.
+- Not Ignorable - A scheduled out of band test can and will be ignored, so this test sequence should run as part of CI.
```
Gilbert09, line 17: We certainly wouldn't want to run this on every PR - I imagine this will be a lengthy process. Maybe it can run when we add a certain label to a PR

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+   move the data that has been temporarily copied into our dev s3 bucket into entities we consider external data
+   sources.
+
+3. Run an on demand temporal workflow to test our standard syncing operation as if it’s any other production workflow.
+
+4. The results of these workflows will end up back in s3 via delta as an output. We should be able to run queries at
+   this point.
+
+5. Compare the results of the original copied s3 payloads to the s3 payloads that went through the temporal processor
+   and were synced. This is mostly a parity check to ensure that our syncing process is working as expected.
```
Gilbert09, line 52: How would we do the comparing here? How do we make sure there's parity?

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+- S3
+- GitHub Actions (Invoke when changing subset of our directories, triggering agent)
+
+## Concerns/Open Questions
+- How feasible is it to recreate instances of more hosted type databases, like Snowflake, would something like
+  LocalStack be an option for us?
+- It is too much work to do this for some many of our integrations, would we become test automation specialists instead
+  of software engineers? How can we maintain this balance over time added more and more source types?
+- What would be the best way to simulate other sources like Salesforce?
+- Should we use a test account for something like Stripe?
```
Gilbert09, line 83: I don't think this would be any better than what we're already doing imo

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+## Tooling
+- ArgoCd
+    - Source containers (Postgres, MySql, Snowflake)
+    - Source destinations (Postgres, MySql, Snowflake)
+    - Clickhouse (Reuse in dev environment as query layer)
+    - Temporal Runner(Reuse in dev environment as query layer)
+- S3
+- GitHub Actions (Invoke when changing subset of our directories, triggering agent)
+
+## Concerns/Open Questions
```
Gilbert09, line 77: This tests one-half of the pipeline - the T and the L in "ETL", but we're not really testing the "E" - the extract layer

Gilbert09, line 77: Have you considered whether testing becomes a magnitude easier if we were to change our pipeline setup? One think I've been thinking about is how the whole pipeline is a single process, we could split this into two and have an event-based system where we extract data in one workflow, fire the data into a kafka topic, then another process consumes from the topic to load data into S3.

In this model, we'd be able to add an extra consumer on the kafka topic to test PR changes or something similar. My idea isn't fully fleshed out, but would be good to consider different pipeline models here too

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+day. At this time, we don't have enough confidence in our (mostly unit) test suite and development workloads don't
+closely mimic the realities of production scenarios. Not all the sources have similar distributions of use, but one
+common theme is that many are SQL based and our data eventually end up in s3 via delta tables and are queryable via
+HogQl, so that feels like the area to hone our focus on for biggest impact. Other data sources like HubSpot, Salesforce,
+and Chargebee are less generic and also would have a smaller blast radius in the event syncing incremental or full table
+syncing gets regressed. These sources are still important and should be incorporated into the test plan as it matures.
+
+## Higher Level Goals
+- Determinism - We do not want to randomly sample sources/schemas in order to test against. This list of tests should
+  grow over time, but remain consistent for feature development.
```
rossgray, line 15: I think, as long as we can reproduce the test, randomly sampling our sources/schemas could be a good way of detecting issues. If we always run tests on the same subset of data sources we may never come across certain types of edge cases

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Concerns/Open Questions
+- How feasible is it to recreate instances of more hosted type databases, like Snowflake, would something like
+  LocalStack be an option for us?
+- It is too much work to do this for some many of our integrations, would we become test automation specialists instead
+  of software engineers? How can we maintain this balance over time added more and more source types?
+- What would be the best way to simulate other sources like Salesforce?
+- Should we use a test account for something like Stripe?
+- Would we like to test out complicated joins on queries between multiple sources and events? 
+
```
rossgray, line 85: Have you considered any alternative approaches?

For example, maybe a way to compare the data we have in PostHog vs the source is to occasionally run some reconciliation queries, which change depending on the source type. Eg for BigQuery you could run some queries to get row counts in the source vs PostHog, to check breakdowns by a certain column, etc. Then we wouldn't need to spin up our own BigQuery instance.

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+    - Clickhouse (Reuse in dev environment as query layer)
+    - Temporal Runner(Reuse in dev environment as query layer)
+- S3
+- GitHub Actions (Invoke when changing subset of our directories, triggering agent)
+
+## Concerns/Open Questions
+- How feasible is it to recreate instances of more hosted type databases, like Snowflake, would something like
+  LocalStack be an option for us?
+- It is too much work to do this for some many of our integrations, would we become test automation specialists instead
+  of software engineers? How can we maintain this balance over time added more and more source types?
```
rossgray, line 81: Definitely something to keep in mind. If the tests are too flaky and unreliable we could risk sinking a lot of time into them and/or lose confidence in them

requests-for-comments/2025-03-14-dw-testing-production-data.md
rossgray: > Plus, we're adding an extra layer of logic in the tests that could independently fail (and be fairly costly to maintain imo)

This is probably my main concern here

rossgray: Also sounds like we'd likely need quite a bit of support from the infra team here (at least for the initial work)

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,89 @@
...
+We now have reasonable traction with our data warehouse product and the cost of making breaking changes with
+incremental updates is growing as we accrue more and more external sources, self-managed sources, and saved queries each
+day. At this time, we don't have enough confidence in our (mostly unit) test suite and development workloads don't
+closely mimic the realities of production scenarios. Not all the sources have similar distributions of use, but one
+common theme is that many are SQL based and our data eventually end up in s3 via delta tables and are queryable via
+HogQl, so that feels like the area to hone our focus on for biggest impact. Other data sources like HubSpot, Salesforce,
+and Chargebee are less generic and also would have a smaller blast radius in the event syncing incremental or full table
+syncing gets regressed. These sources are still important and should be incorporated into the test plan as it matures.
+
+## Higher Level Goals
```
EDsCODE, line 13: What's the specific goal that would be great to solve here? Could you add specific issues we would want to protect against/detect with live testing (e.g. credentials failing because something with our live key for a source went wrong/expired)

phixMe: I will split out the higher level goals and testing goals into different subsections.

requests-for-comments/2025-03-14-dw-testing-production-data.md
rossgray: I suppose it depends on what the queries are. If they're just running aggregations like row counts then the actual data egress would be minimal. However it would obviously incur additional compute on the user's end.

Though maybe we could mitigate this:
- choosing a random source to perform the test on, so that each source would only likely have an additional query per week or something
- have an opt-in check box which would be something like "allow us to perform a data integrity check on your data once a week"

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,109 @@
...
+- Determinism - We do not want to randomly sample sources/schemas in order to test against. This list of tests should
+  grow over time, but remain consistent for feature development.
+- Repeatability - A failure one run should fail the same way the next run, A success should also be repeatable.
+- Not Ignorable - A scheduled out-of-band test can and will be ignored, so this test sequence should run as part of CI.
+  We shall have a designated label in CI called `dw-integration-suite` that will trigger CI to run these tests.
+- Uses "real" data - Not anchored to "test" datasets that fail to capture the madness of the real world.
+- No Perf Degradation - We don't wish to interfere with any production workloads or queries.
+- No Cost/CPU Impact - We don’t wish to run any out of band testing workloads that would impact a live service, with the
+  possible exception of testing our own services that are in the purview of PostHog.
+
```
rossgray, line 34: I think (a relatively) easy win here would be to not even use customer's data but to use our own test data.

A lot of the issues I come across when working on data imports for example is just testing I don't break something when I update a particular source. Even if it's just a simple refactor, there's not currently an easy way of testing my changes. For example, if I make a change to a Snowflake source, I need to ensure I have a Snowflake account set up, then set up a new Snowflake source locally, put in my credentials, run a sync and confirm it works.

What we do in batch exports is we have a set of end to end tests that actually test against a real instance of the system (eg Snowflake). This means I can run this specific set of tests and have a lot of confidence I haven't broken something.

Of course, this won't cover all edge cases related to actual production data, but I feel it would cover > 90% of issues, and we can always build on our test suite when we come across a bug. For example, if we come across a bug with table names that contain a numeric character, we can just add a test for that.

This could also be a first step (i.e. we can think about how to incorporate customer's data later if so wish).

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,109 @@
...
+## Concerns/Open Questions
+
+- How feasible is it to recreate instances of more hosted type databases, like Snowflake, would something like
+  LocalStack be an option for us?
+- It is too much work to do this for some many of our integrations, would we become test automation specialists instead
+  of software engineers? How can we maintain this balance over time added more and more source types?
+- What would be the best way to simulate other sources like Salesforce?
+- Should we use a test account for something like Stripe?
+- Would we like to test out complicated joins on queries between multiple sources and events? 
+
```
MichaelKutsch-ph, line 105:
```suggestion
- using real prod data in a non-prod environment (-> PII, data-deletion-requests), data must be sanitized and tested or generic test data has to be generated
```

requests-for-comments/2025-03-14-dw-testing-production-data.md
```diff
@@ -0,0 +1,109 @@
+# RFC - Test Pipeline Changes on Real Customer Data w/o Affecting Actual Prod Data
```
MichaelKutsch-ph, line 1: I'm challenging the "real customer data". Using _real_ customer data has significant effects on security and data privacy things (and we have to check off any GDPR and SOC2 concern). 

I think the only way is to properly sanitize this data or generate generic test data. Using real customer data sounds like danger zone to me.

requests-for-comments/2025-03-14-dw-testing-production-data.md
MichaelKutsch-ph, line 1: If I zoom out, it sounds like you're asking for a proper integration environment.

MichaelKutsch-ph: or we simply use `dev` for this, given we don't copy unsanitized customer-data to it

PR #309, opened 2025-04-15
RFC: Consistent scripts to rule them all

haacked:
Proposal for adopting and adapting a consistent set of scripts across our OSS repositories:

> Being able to get from git clone to an up-and-running project in a development environment is imperative for fast, reliable contributions.

### DIFF
--- a/requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
+++ b/requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
@@ -1,0 +1,65 @@
+# Request for comments: Consistent scripts to rule them all
+
+## Problem statement
+
+*Who are we building for, what are their needs (include example cases), why is this important?*
+
+PostHog has 235 (at the time of writing) public repositories. Each of these repositories has a unique way to get the project up and running locally.
+Each repository has a unique way to run tests, build the project, and so on. For those of us who work on the many client SDK libraries, this adds a
+lot of friction as we switch from project to project and have to learn a whole new set of incantations. Not only that, it makes it more challenging
+for external contributors to get involved as they have to learn a whole new process for each repository they work on.
+
+And as we continue to grow, this challenge will only grow.
+
+## Solution
+
+Adopt and adapt GitHub's approach of [scripts to rule them all](https://github.blog/engineering/engineering-principles/scripts-to-rule-them-all/) to
+fit our needs.
+
+> Being able to get from git clone to an up-and-running project in a development environment is imperative for fast, reliable contributions.
+
+I'm not suggesting we do _exactly_ what they do, but a similar approach would be a big benefit. Not every repository will need every script. Some
+repositories will need scripts custom to the environment. That's all fine. The goal is to have a baseline set of scripts that we can use to get a
+project up and running in a development environment and a known location to look for those scripts.
+
+## Context
+
+For context, here's the list of scripts referenced in the GitHub blog post ([samples available in a repository](https://github.com/github/scripts-to-rule-them-all)):
+
+* `script/bootstrap` - Install dependencies (Ex. npm packages, brew packages, etc. Usually run once after cloning the repository)
+* `script/setup` - Setup the project. Seems like this could just be merged in bootstrap
+* `script/update` - Update project after a pull such as migrations, etc.
+* `script/server` - For web apps, starts the application. For client SDKs, might start an example site.
+* `script/test` - Run tests (Ex. `npm test`, `bundle exec rspec`, etc.). Also includes linting, formatting, etc.
+* `script/cibuild` - used by the CI system to run tests, build, etc.
+* `script/console` - is used to open a console for your application.
+
+## Design
+
+For `posthog/posthog`, we already have a set of scripts in the `bin` directory. One issue with using `bin` across the board is that some environments
+add `bin` to the `.gitignore` file by default because that's where they compile binaries to. However, we've already set this precedence, so it may make
+sense to stick to it. For example, [posthog/posthog-dotnet](https://github.com/PostHog/posthog-dotnet/tree/main/bin) already has a `bin` directory.
+
+* `bin/setup` - Install or upgrade dependencies (Ex. npm packages, brew packages, etc. Usually run once after cloning the repository).
+* `bin/update` - Updates dependencies after a pull. This could simply call `bin/setup`.
+* `bin/build` - Build the project, for projects that are compiled such as C#, Java, etc.
+* `bin/start` - Start the project. For SDKs, this might start an example server.
+* `bin/test` - Run tests (Ex. `npm test`, `bundle exec rspec`, etc.). Also includes linting, formatting, etc.
+* `bin/fmt` - Optional: Format/lint code. This can be called by `test`.
+
+## Feedback Requested
+
+* What other scripts do we need?
+* What should the scripts be named?
+* Should we use `bin` or something else?
+* Should we also define repository templates for new repositories?
+
+## Next Steps
+
+I'll leave this open for two weeks to gather feedback.
+
+After that, if there's consensus on the approach:
+
+1. I'll create a PR to the Handbook with documentation on our approach.
+2. I'll create a scripts repository that contains some template scripts (and script utilities).
+3. I'll create a PR to each SDK repository to add the new scripts as I go.


rafaeelaudibert (approved):
+1 from me, but suggest we keep the `bin` folder for now. If we ever have to build something we can use `out`

requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
```diff
@@ -0,0 +1,72 @@
...
+PostHog has 235 (at the time of writing) public repositories. Each of these repositories has a unique way to get the project up and running locally.
+Each repository has a unique way to run tests, build the project, and so on. For those of us who work on the many client SDK libraries, this adds a
+lot of friction as we switch from project to project and have to learn a whole new set of incantations. Not only that, it makes it more challenging
+for external contributors to get involved as they have to learn a whole new process for each repository they work on.
+
+And as we continue to grow, this challenge will only grow.
+
+## Solution
+
+Adopt and adapt GitHub's approach of [scripts to rule them all](https://github.blog/engineering/engineering-principles/scripts-to-rule-them-all/) to
```
marandaneto, line 16: TIL github approach for scripts

requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
```diff
@@ -0,0 +1,72 @@
...
+* `bin/start` - Start the project. For SDKs, this might start an example server.
+* `bin/test` - Run tests (Ex. `npm test`, `bundle exec rspec`, etc.). Also includes linting, formatting, etc.
+* `bin/fmt` - Optional: Format/lint code. This can be called by `test`.
+
+## Alternative Ideas Considered
+
+I've noticed a lot of our repos have a `make` file. This is especially common for mobile development. An alternate proposal would be to standardize on `make` rather than these `bin` scripts.
+
+My recommendation is to still standardize on `bin` scripts so that it's consistent across the board. Some of the bin scripts are pretty complicated and it'll be hard to port to `make`.
+If some repos have `make` files, that's fine. The bin scripts can call the `make` commands.
```
marandaneto, line 55: the only downside here is that we have a bit of duplication

requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
```diff
@@ -0,0 +1,72 @@
...
+For `posthog/posthog`, we already have a set of scripts in the `bin` directory. One issue with using `bin` across the board is that some environments
+add `bin` to the `.gitignore` file by default because that's where they compile binaries to. However, we've already set this precedence, so it may make
+sense to stick to it. For example, [posthog/posthog-dotnet](https://github.com/PostHog/posthog-dotnet/tree/main/bin) already has a `bin` directory.
+
+* `bin/setup` - Install or upgrade dependencies (Ex. npm packages, brew packages, etc. Usually run once after cloning the repository).
+* `bin/update` - Updates dependencies after a pull. This could simply call `bin/setup`.
+* `bin/build` - Build the project, for projects that are compiled such as C#, Java, etc.
+* `bin/start` - Start the project. For SDKs, this might start an example server.
+* `bin/test` - Run tests (Ex. `npm test`, `bundle exec rspec`, etc.). Also includes linting, formatting, etc.
+* `bin/fmt` - Optional: Format/lint code. This can be called by `test`.
```
marandaneto, line 48: I think `fmt` and `lint` should be different tasks

requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
```diff
@@ -0,0 +1,72 @@
...
+I've noticed a lot of our repos have a `make` file. This is especially common for mobile development. An alternate proposal would be to standardize on `make` rather than these `bin` scripts.
+
+My recommendation is to still standardize on `bin` scripts so that it's consistent across the board. Some of the bin scripts are pretty complicated and it'll be hard to port to `make`.
+If some repos have `make` files, that's fine. The bin scripts can call the `make` commands.
+
+## Feedback Requested
+
+* What other scripts do we need?
+* What should the scripts be named?
+* Should we use `bin` or something else?
```
marandaneto, line 61: I'd rather use just `scripts`, `bin` is used by some other languages/frameworks (eg a dart project will have its executable and other entry points in the bin folder) as part of the folder structure and our scripts will be among many different files, a bit confusing

marandaneto:
nice, thanks for the initiative @haacked 
As you mentioned, I did that with the mobile SDKs using makefiles (more common for mobile development), so I can't agree more.

requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
haacked: Maybe `lint` should just go into the `bin/build` task? Hear me out.

It seems to me that linting for a dynamic language serves a similar purpose to compilation in a compiled language in that it points out obvious errors.

So for a language like python, `bin/build` could run both `mypy` and `flake8`. Thoughts?

haacked: I'm open to just having it be a separate script as well.

requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
haacked: It's too bad Swift doesn't have union types. In practice, a feature flag result will only be `null`, `bool`, or a `string`. If `getFlagValue` returned something other than one of those types, that would be unexpected. In that case, I think returning `false` is appropriate OR we should throw an exception. What do you think?

requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
```diff
@@ -0,0 +1,72 @@
...
+
+* `script/bootstrap` - Install dependencies (Ex. npm packages, brew packages, etc. Usually run once after cloning the repository)
+* `script/setup` - Setup the project. Seems like this could just be merged in bootstrap
+* `script/update` - Update project after a pull such as migrations, etc.
+* `script/server` - For web apps, starts the application. For client SDKs, might start an example site.
+* `script/test` - Run tests (Ex. `npm test`, `bundle exec rspec`, etc.). Also includes linting, formatting, etc.
+* `script/cibuild` - used by the CI system to run tests, build, etc.
+* `script/console` - is used to open a console for your application.
+
+## Design
```
robbie-c, line 37: Love the idea!

I'd love to also standardise the release process for new versions of a library, I don't think you need to do that in this PR though

haacked: > I'd love to also standardise the release process for new versions of a library, I don't think you need to do that in this PR though

Same! It's on my list of things to do. With RELEASING, I think I'll document the many ways we do it today and propose a couple ideas for how we can consolidate it into a common approach. Obviously, not all things can be released in exactly the same way, but we can probably come up with a framework for releasing things that's fairly consistent.

marandaneto: You can inspire from (or even use it) https://github.com/getsentry/craft
We built craft at sentry exactly for this reason

requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
marandaneto: I think for speedy, i prefer the `lint` to be its own thing, but it really depends on the tooling.
what i did in the past and worked really well is that i have a cmd for each step (fmt, lint, etc), and a cmd that does it all (eg build) that will run all its dependencies (fmt lint, etc).
eg https://github.com/PostHog/posthog-android/blob/9906ac8ac333a0e4990b64dceeca892453d81303/Makefile#L6-L7
^ runs everything else, so in CI I just do that, but locally if i want to check something, i can just `make format`

requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
marandaneto: I'd also recommend having the release and contributing markdown in the repos
eg https://github.com/PostHog/posthog-android/blob/main/RELEASING.md
https://github.com/PostHog/posthog-android/blob/main/CONTRIBUTING.md
so that open source contributors (and new employees) know how to build, run, and release things.

pauldambra (approved):

requests-for-comments/2025-04-15-scripts-to-rule-them-all.md
haacked: I'll let you and @rafaeelaudibert debate `bin` vs `scripts`. 😆

rafaeelaudibert: I vouched for `bin` because we're already using and I don't wanna hurt our muscle memory.

It is true `bin` is used in stuff like Ruby/Rails as well to include default scripts, but isn't this just the same as we already have here? The only difference is that they might have some extra files inside `bin` that don't exist in Django/JS land, for example

If that's not the case, and they're more than just scripts, I'm fine changing to scripts. it will hurt in the short-term, but it doesnt really matter in the long-term

marandaneto: Some languages/frameworks have `bin` as part of .gitignore btw, you'd need to remove from gitignore and add all the other files that should not be checked under the source control.
no strong opinions here, I just think that `scripts` would be cleaner and avoid this known pitfall.
for muscle memory, you could either set up your own alias or even have a symbolic file linked to the correct one so if you do eg ./lint in the root folder, it just works it does not matter if its within the bin or scripts really

ivanagas:
If anyone has final opinions, I opened a `posthog.com` PR for this: https://github.com/PostHog/posthog.com/pull/12127

gewenyu99:
Not an opinion, but I'll be adding generation scripts for SDK references parsed from doc strings. Do we want to standardize the executable for this? Do we want this to be a part of build scripts?

PR #310, opened 2025-04-16
RFC: PostHog Builder Community Incubator

ivanagas:
An idea for a PostHog way to do events and community.

### DIFF
--- a/requests-for-comments/2025-04-15-community-incubator.md
+++ b/requests-for-comments/2025-04-15-community-incubator.md
@@ -4,6 +4,8 @@ Decider: Ian
 
 Decision date: April 25
 
+Decision: Yes! Daniel Z is leading :)
+
 ## Problem
 
 We know we want to do events at PostHog.
--- a/requests-for-comments/2025-04-15-community-incubator.md
+++ b/requests-for-comments/2025-04-15-community-incubator.md
@@ -1,0 +1,66 @@
+# Request for comments: PostHog Builder Community Incubator
+
+Decider: Ian
+
+Decision date: April 25
+
+## Problem
+
+We know we want to do events at PostHog.
+
+- We’ve been talking about doing more for over a year.
+- We are hiring someone to “do events”
+- It’s going to be their goal to create 500 superfans.
+
+The problem is we don’t really know what types of events we want to do. We’ve partnered on and sponsored some events, but nothing of our own. The standard options don’t really work for what we want to do.
+
+1. **Casual meetup:** Doesn’t create super fans and we want to encourage building, not networking.
+2. **Accelerator:** Won’t get to 500 super fans quick enough. There are a lot of other accelerators to go through. 
+3. **Conferences:** Too expensive and undifferentiated, basically the event version of a whitepaper. 
+
+Our best option is “hackathons” and we’ve seen some good results partnering with other companies on (such as the [ElevenLabs hackathon](https://hackathon.elevenlabs.io/)).
+
+The tricky part about hackathons is that PostHog as a product is fundamentally not a good fit with hackathons. You don’t use PostHog to build the v0 of a product, you use it to build version 0.1. This is a different position to companies like ElevenLabs and Supabase who’s events we look up to. 
+
+We also want something “we would actually want to go to.” So what should this be?
+
+## Solution: The PostHog Builder Community Incubator
+
+Inspired by “something we’d actually want to go to,” I thought about the events I have actually wanted to go to and there’s one that comes to mind: [Novus](https://www.novus.build/). 
+
+Novus is a weekly builder meetup in Vancouver that happens every Saturday. It’s 3 hours, starting with a quick round of intros, then 2.5 hours of building, and then demos at the end. It attracts exactly the type of people PostHog wants to appeal to: builders working on startups, prototypes, and side projects.
+
+![image](https://github.com/user-attachments/assets/ba029e04-c850-4bfc-915d-a732a79ad501)
+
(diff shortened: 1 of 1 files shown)


andyvan-ph:
Sounds like great idea. No notes. :D

joethreepwood (commented):
I'm very pro for this idea - which I'm saying at the start so this isn't interpretted as not being "Yes, and". However, I want to understand the logistics you're proposing properly. I'll go through it in order. 

## Payment
How do you expect that payment to work operationally? There seem to be three options. 

- **We send them money upfront.** Highest risk for us. Easiest for them. Open to exploitation. 
- **We reimburse.** More control for us. Less ease for them. Adds process. 
- **We pay for them.** Slows things way down for everyone. We get less value from them. 

There's a possible hybrid model. We give 50% up front, and refund the rest. That's the best of both worlds, but also the worst of both worlds.

This is my big concern right now, so would love to know how you see it. 

Also, what happens if they go over/under budget? You're proposing communicating it as "We _pay_ you" when this isn't actually what would happen, right? We will probably need to be annoyingly clear that this is actually a grant and _not_ a payment. 

## Contracts
I hate to be this guy, but I do think we'd need a contract or something here. Something simple, just to cover things like basic liability, use of funds, cancellation policies, etc. We could do this easily ofc, but flagging so we know that's a required step.

I'm an over-worrier and I'll try to avoid taking this to an extreme (e.g. What if someone dies? What if our host says something bad?) but we should at least think about a basic protection for the funds and clarity on what we will/won't provide. 

## Scale
Small note: We should geo-target this to specific cities/areas we want to hit. This'll give it focus, ensure we can get better attendance (e.g. no tiny towns in Utah), and up quality (e.g. no spammed applications from India). 

Suggest we'd start with some standard candidates: New York, Toronto, SF, London, Paris, Berlin. 

## Involvement
Just so we're clear, it sounds like we'd have very light involvement in this? We'd basically fund a space and food, give some basic branding, and that's it. We're not committing to having someone in attendance, or that they work on PostHog products or anything? That sounds good to me and is the best way to go fast, but I want to confirm. 

So, yeah, like the idea, but I want to understand the details.

ivanagas:
My thoughts:

1. We send them money at the end of the four weeks after they've successfully done it. All of these events I'm familiar with don't have any expenses. They borrow space and rely on volunteers to run it, so nothing to reimburse or pay for. This would be a grant to encourage them to get started basically.
2. Yes. We can have terms on getting accepted into the program and grant. Good catch. 
3. Geo, I'm pretty open tbh. If they can get 20 people out for 4 weeks in tiny town in Utah that sounds like a success to me. A bunch of big cities already have something similar as well (Toronto and Montreal have [Builder Sundays run by Shopify](https://builder-sundays.myshopify.com/) for them). We should point to meetups we know already exist instead of splintering off.  
4. Yes as you say it. No one from PostHog needs to attend, no one needs to use PostHog.

joethreepwood:
> We send them money at the end of the four weeks after they've successfully done it. All of these events I'm familiar with don't have any expenses. They borrow space and rely on volunteers to run it, so nothing to reimburse or pay for. This would be a grant to encourage them to get started basically.

Gotcha, so we _are_ paying them and this money would be expressly for the purpose of that. From the original description I'd expected it to be: "Here's $800 to cover costs, thanks for volunteering" and not "We'll pay you $800 to do this and won't cover any other costs"

@fraserhopper This may be premature, but I'm curious if this would create any people ops impact in terms of contracts, etc. I'm hoping it's easy and as simple as a quick thing for them to sign, but would help to know what the process would need to be here. 

One thing that's on my mind here is that in a busy city someone could potentially turn this into a full-time job, running multiple events each month for $800 a pop and getting a handful of attendees each time. That's potentially great for us (we'd probably consider hiring such a person), but worth thinking about from an ops POV too. 

> Geo, I'm pretty open tbh. If they can get 20 people out for 4 weeks in tiny town in Utah that sounds like a success to me. A bunch of big cities already have something similar as well (Toronto and Montreal have [Builder Sundays run by Shopify](https://builder-sundays.myshopify.com/) for them). We should point to meetups we know already exist instead of splintering off.

Yeah, sounds alright. I think we should _definitely_ start with limiting it to specific countries at least. I agree it's fine if we get a small town, but I can also see us getting bombarded with low quality applications from India as happens with some other iniatives. If we can just focus it on US, Canada, UK, EU, I think that's a fine starting point.

ivanagas:
> One thing that's on my mind here is that in a busy city someone could potentially turn this into a full-time job, running multiple events each month for $800 a pop and getting a handful of attendees each time. That's potentially great for us (we'd probably consider hiring such a person), but worth thinking about from an ops POV too.

In my mind, someone would "graduate" from this program after the first month and they could go on running these events without our involvement (or money). We'd help get the community off the ground and then they can continue on if they want. 

We could also keep supporting some communities (like a growth/continuity fund).

Cool with limiting countries. We can say this is for payment reasons :)

ivanagas:
My plan is for them to run multiple events to get the grant and help seed a community, I think it has a chance to work, but maybe I'm wrong 🤷

joethreepwood:
For sure, we should find out! Wasn't meaning to shoot it down, just flag
it's something we should keep an eye on as it grows

fraserhopper:
Sorry this was a bit buried. 

My gut here is we do a bit of external trust & feedback > process here. I would actually suggest we maybe even give the money up front here - like we might get burned but the right people will far exceed $800 if they feel like we are trusting them. So if we are starting small (eg 5 people x $800) learn a bit then we can react. 

If this starts to gain traction and we are doing it with lots of people we may want to figure out something more structured and @hector-r-759 can figure out the implications of this legally/contractually/indemnity side. 

I think in the beginning thought the risk is fairly low, it's when it starts to scale I would think we want a bit more protection.

hector-r-759:
@fraserhopper agreed.  If we're talking a pilot here where the exposure is ~$800 per event just to see how things take off, I don't think we need to over-complicate things at this stage.  We would want to document the payment, but I'm not really sure I see a big legal/risk exposure to us at this stage, especially if its sort of given up front, with little to no-strings attached and little to no PostHog involvement.  

The biggest risk is probably just straight up losing the money if you give it up front, but as long as we're prepared for that potential outcome, it seems fine.  We weren't going to sue to recover $800 anyway, so more robust legal protections for recovering damages, use of funds, etc. are practically speaking, not all that useful anyway.  That being said, obviously if things take off and we're starting to talk about more material amounts, it would probably make sense to start getting something more robust in place.

PR #324, opened 2025-06-18
RFC for generated SDK specs and reference documentation

gewenyu99:
RFC to generate specs to make pages like this possible:

https://github.com/user-attachments/assets/52bcbc4e-7a3d-48e0-8453-15e196b6fb0f

See RFC inside


requests-for-comments/2025-06-18-sdk-specs.md
```diff
@@ -0,0 +1,301 @@
...
+     * 
+     * @returns The newly initialized PostHog instance
+     */
+```
+
+This is **very similar** to what we already do, but standardizing this to be parsable will go a long way. This also ensures that when SDK code is updated, so will the specs be updated.
+
+### Specs generated
+
+There isn't a good standard I like. So we'll create `hogRef` inspired by [Supabase's SDK refs](https://raw.githubusercontent.com/supabase/supabase/82715efc250bd1e706491effab91968ea100479a/apps/docs/spec/supabase_dart_v2.yml).
```
rafaeelaudibert, line 199: Is it likely other people will want to consume this `hogRef` file or is this internal-only for our website? We'd probably want to do something more mainstream if this is supposed to be consumed by our AI overlords, for example

requests-for-comments/2025-06-18-sdk-specs.md
```diff
@@ -0,0 +1,301 @@
...
+#### Categories
+We will display methods by category. Init, config, and identify persons categories are always at the top. The rest sorted by product and alphabetical.
+
+### Proposed workflow
+We will automate generation, but manually version pin known-good versions to prevent breaking the website because a ref is generated in a format posthog.com can't parse.
+- References are served as jsons from GitHub
+- Each time a PR is merged to main, run a workflow to re-build json specs
+    - for JS, this means using `api-extractor` from Microsoft and a parse script maintained by us to create `hogRef`
+- Use Gatsby to fetch and build pages from version pinned jsons from GitHub
+- We will manually bump sdk version pins (or list of versions) manually
```
rafaeelaudibert, line 292: When and who?

requests-for-comments/2025-06-18-sdk-specs.md
```diff
@@ -0,0 +1,301 @@
...
+
+## Effect on existing docs
+Where possible, for each SDK doc, hand-write usage guides for set up and key use cases only. This replaces the potentially giant pages currently used to document SDKs.
+
+## Ownership
+If you update the SDK, you're expected to update the doc comment.
```
rafaeelaudibert, line 298: Very reasonable. I'd love to see in this RFC, however, how you plan on delegating ownership in actually writing/reviewing the initial set of docs. I wouldn't expect the Docs team to be able to document all of the public interfaces we have for `posthog-js` - in the same way I wouldn't expect myself to know what each method does even though I've contributed 4k lines to it 😆

rafaeelaudibert: Ah, I now see you have an outstanding PR, but that doesnt cover everything and you're simply "asking for help". I believe it makes sense for you/someone to delegate this task to people explicitly or else you'll have a hard time getting people to do this because it's time-consuming to do it "on the side" but it's possible to do it if we consider it first-class work.

requests-for-comments/2025-06-18-sdk-specs.md
gewenyu99: I'm open to other spec suggestions, but I actually didn't find a good, language agnostic spec that's available. 

I would love to not have to invent our own thing, but I did this at Appwrite, Supabase is doing their own thing, afaik, Stripe is doing their own thing, too.

requests-for-comments/2025-06-18-sdk-specs.md
gewenyu99: Good point, let me specify

requests-for-comments/2025-06-18-sdk-specs.md
```diff
@@ -0,0 +1,309 @@
...
+- Certain information like [default properties](https://posthog.com/docs/data/events#default-properties) is only available in source code.
+- Without consistently formatted docs, LLMs have a harder time understanding our SDKs.
+
+## Quick note on scope
+We'll tackle generation for our most Popular SDK before tackling general Open API spec accuracy and all SDKs. Starting with PostHog JavaScript Web ofc.
+
+## Design 
+There are 3 separate problems to solve here:
+1. Documenting SDKs with doc comments + generated types
+2. Generating a common intermediate spec schema for all SDKs
```
oliverb123, line 20: I don't _quite_ understand the need for a common schema here? I hate to go full 🦀, but my "gold standard" for docs are the [tokio ones](https://docs.rs/tokio/latest/tokio/), which are entirely written as doc-comments in the source code. I feel like a huge amount of the benefit could be had here just through auto-generating web-hosted docs from docstrings/doc comments/etc, and the whack-a-mole of supporting the myriad ways different languages work when trying to generate a unified schema might drive us mad.

oliverb123: A kind of related point: I feel like for most devs, most of the time, having to open a web browser to see how to use something is the worst case scenario. This is most untrue when someone is first evaluating posthog, but in those cases, I think quickstart tutorials are more relevant than SDK docs.

Once someone is using us, they _really_ want docs to be available in their IDE

gewenyu99: I see your point. I think entirely doc-comment driven docs are nice in some sense, but they're missing out in some key points in SEO and GEO (fancy LLM SEO).

I think the tokio docs are fine for a library, not really enough for product docs. 

I think a shared schema is important for consistency sake. We need our SDK docs to look like they belong to the same product with strong consistency. Something Clerk/Stripe/Supabase and other SaaS type products do well.

This is something I've also done at Appwrite, where we generated all our SDKs and SDK docs from the same source.

I think if they have to look in their SDK to get reference docs, we've already lost. I believe there are two types of readers of docs:
- Those who will figure it out at all cost -> they will happily dig in our source code. We don't need to worry about them in the first place
- Those who google and ask Chatjipity, and when they don't find reference docs they yell at us impatiently on Twitter.

Unfortunately, this effort for reference docs is to please the second type of person.

requests-for-comments/2025-06-18-sdk-specs.md
```diff
@@ -0,0 +1,309 @@
...
+This has potential to cause huge churn and debt. We can't keep chasing after fast evolving SDK versions manually.
+
+### Why this is a problem
+- Docs will frequently drift out of date.
+- This causes our hand rolled SDK docs pages to be obscenely long and impossible to read. See [JavaScript Web](https://posthog.com/docs/libraries/js) and [React Native](http://localhost:8001/docs/libraries/react-native).
+- Certain information like [default properties](https://posthog.com/docs/data/events#default-properties) is only available in source code.
+- Without consistently formatted docs, LLMs have a harder time understanding our SDKs.
+
+## Quick note on scope
+We'll tackle generation for our most Popular SDK before tackling general Open API spec accuracy and all SDKs. Starting with PostHog JavaScript Web ofc.
```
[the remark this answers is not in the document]
gewenyu99: What we did before, is that generally there is platform specific setup information we can roll into the beginning of the docs. This is usually done by having a custom header for the specs that cover platform specific setup, and rest of references sharing simple reference type info.

requests-for-comments/2025-06-18-sdk-specs.md
gewenyu99: And I guess to add to this, I do also believe that this is only half for the purpose of utility, half is for looking cool 😉 

We believe docs are also an important part of marketing. I want someone to see our docs and feel some degree of polish and appreciate the details. This is certainly over-engineered. We can just use typedoc to generate MD/HTML pages. But I think that's not the 10x'd docs we're looking for.

requests-for-comments/2025-06-18-sdk-specs.md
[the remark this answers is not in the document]
oliverb123: > Do you think that changes anything for the mobile SDKs?

Really can't speak to mobile SDKs, you'd have to ping the mobile folks. A platform header/intro makes a lot of sense

gewenyu99:
I'm gonna merge this. We've shipped a few and I think there's been good value here.

PR #353, opened 2025-08-22
RFC: Should we start a second newsletter focused on AI for engineers?

ivanagas:
Should we start a second newsletter focused on AI for engineers?

> **Note:** This is quite a speculative RFC, I more want to get the idea out there and have discussion about its potential.

### DIFF
--- a/requests-for-comments/2025-08-22-second-ai-newsletter.md
+++ b/requests-for-comments/2025-08-22-second-ai-newsletter.md
@@ -1,0 +1,46 @@
+# RFC: Should we start a second newsletter focused on AI for engineers?
+
+> **Note:** This is quite a speculative RFC, I more want to get the idea out there and have discussion about its potential.
+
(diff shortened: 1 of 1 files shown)


miqcie (commented):
Seeing cross-publisher subscribers with big Substacks like Towards AI Newsletter (130K+) and Last Week in AI (hundreds of thousands) would help size the current audience overlap.

Unfortunately Substack's audience overlap data is practically nonexistent ([support docs](https://support.substack.com/hc/en-us/articles/1500006075182-Audience-analytics), [community complaints](https://www.indiehackers.com/post/does-substack-offer-any-subscriber-overlap-analytics-8d6e42ff0f)).

Consider testing with a special edition in the existing newsletter first. Tag engaged readers in CDP to measure interest and if the addition of a FT copywriter for a separate publication is worth it.

PR #363, opened 2025-09-10
Customer reach out email for breaking change to feature flags

andyzzhao:
Email draft to inform customers of a breaking change coming to feature flag evaluation. I intend to send this email to specific team ids affected. 

US: https://metabase.prod-us.posthog.dev/question/1585-list-of-team-ids-affected (~250)
EU: https://metabase.prod-eu.posthog.dev/question/512-affected-team-ids-in-eu (~100)

https://github.com/PostHog/posthog/issues/37854

### DIFF
--- a/requests-for-comments/2025-09-09-ff-eval-fix.md
+++ b/requests-for-comments/2025-09-09-ff-eval-fix.md
@@ -34,7 +34,7 @@ However, please review your affected feature flags if:
 3. Your analytics or A/B tests depend on specific variant distributions
 
 ## Fix Release Timeline
-- **PostHog Servers**: Fix deployment [2025-09-15]
+- **PostHog Servers**: Fix deployment [2025-09-12]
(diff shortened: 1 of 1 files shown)


joethreepwood:
Thanks for doing this on the right repo, @andyzzhao -- though FYI we have an issue template that works a little better for these. Email comms currently all come through me. 

This target looks quite small and we have some other comms lined up at the moment. What I'd suggest is that we aim to send this email next week, giving two weeks notice of the change. We may want to follow up with a reminder email too. 

Does that timeline work for you, or do we need to go faster? If it works, I can spin this up into a draft soon -- the main question is whether you want it to come from you (personal touch, you'll get all replies) or from the company account (impersonal, better design).

andyzzhao:
Thanks @joethreepwood! I made an issue following the template here: https://github.com/PostHog/meta/issues/364. We can move our conversation there.

PR #416, opened 2025-10-21
New event template

joethreepwood:
We have a growing number of event issues, which is great. 

I'm not fussed how they're organized, but we do need a way to make sure that invites we sent (especially for large events) get added to the Messaging Calendar going forward - and the most direct way to do that is via a template with a simple checklist. 

I also had to rename the community event template, so I'll whip through the main repo now and see if we need to redirect there, but this should be easy.

### DIFF
--- a/.github/ISSUE_TEMPLATE/event-plan.md
+++ b/.github/ISSUE_TEMPLATE/event-plan.md
@@ -1,7 +1,7 @@
 ---
 name: 'Event Plan'
 about: For PostHog team to organize events
-title: 'Event: [City or Event Name]'
+title: '[Event] - Event Name Here '
 labels: [marketing]
 assignees: 'Assign yourself, Daniel Zaltsman'
 ---


joethreepwood:
Idea is really that the checklist here acts as a prompt to add to the messaging calendar and, since you create issues anyway, it seemed this would be pretty low-hassle. 

Once events are added to the calendar there's an automation that triggers to let Marketing and Sales teams know in their team channels on Slack. 

Idea here is partly to give us greater visibility of when messages send (so that we don't send a big invite, a changelog, incident comms, and a product announcement all on the same day). We also want to give Sales the chance to raise red flags about some things, but I don't think that case will impact events at all.

joethreepwood:
> Looks like you're doing two things in this PR editing the template and making a new md file. The md file - what will that be?

So, we currently have a file called `event-plan.md` which is an event template intended for people in the community to create their own events. 

This PR creates a _new_ event template for us to create event issues from and uses that to replace the current `event-plan.md` file. The _old_ template, for the community, moves to a new template saved as `community-event.md`.

This isn't an area where I feel super strongly, tbh. We'll need to cross this bridge at some point, but if you've got a better idea for how we can make sure we add things to the messaging calendar then I'm all for it -- even if that idea is just "Trust me, I'll remember".

PR #493, opened 2026-02-20
Logs clickhouse cluster data loss post mortem

frankh:
Post mortem for the logs cluster data loss incident

### DIFF
--- a/post-mortems/2026-02-20-logs-data-loss.md
+++ b/post-mortems/2026-02-20-logs-data-loss.md
@@ -59,4 +59,4 @@ We lacked specific monitoring for the integrity of "cold" data stored in S3. Our
 
 1.  **Immediate Configuration Audit:** Disable Zero Copy Replication on all clusters immediately. Conduct a full audit of the Logs ClickHouse configuration and ensure no experimental features are used in production.
 2.  **Implement S3 Object Protection:** Enable S3 Versioning on the underlying storage buckets. This ensures that even if the database application issues a destructive command due to a bug, the underlying data objects can be recovered.
-3.
+3. Before a product is made Generally Available, we spot check configurations and our data integrity strategies to find and correct for potential single points of failure


post-mortems/2026-02-20-logs-data-loss.md
```diff
@@ -0,0 +1,58 @@
...
+
+This new cluster uses S3 disks in ClickHouse and data parts are automatically uploaded to S3 after 24 hours - this is to allow us to scale to the enormous amount of volume required for Logs, our own internal log consumption is over 500MB/s, or about 1PB/month (compressed it's about ~100TB/month)
+
+A bug in ClickHouse caused it to unexpectedly start deleting almost all of the data parts in S3. The database is replicated with two replicas, however very early on in the project we had set "Zero Copy Replication" on in the cluster. This is an experimental feature that ClickHouse do not recommend in production, for exactly this reason: A bug that should have caused a single replica to be deleted instead deleted the data for both replicas.
```
oliverb123:
```suggestion
A bug in ClickHouse caused it to unexpectedly attempt to delete almost all of the data parts in S3. The Logs database is replicated, with two replicas, however very early on in the project we had enabled "Zero Copy Replication" in the Logs cluster nodes. This is an experimental feature that ClickHouse **does not recommend** in production, for exactly this reason: a bug that should have caused a single replica to be deleted instead deleted the data everywhere.
```

oliverb123 (commented):
Buncha small things. One judgement call on whether to make the differing cluster ownership between logs and CH explicit (I lean yes but would defer to frank, or CH, or ben, or anyone with a stronger opinion really). Removed mentions of CH from the topline - explicitly and entirely taking ownership goes a long way ime, and while I've no doubt it wasn't intended to, opening paragraph read a bit like pointing a finger (to an external audience - if this was an internal-only review I think it would have been totally fine).

PR #514, opened 2026-04-01
Wizard and in-app onboarding

gewenyu99:
An initial draft to facilitate discussions on how to make Wizard x in-app onboarding the coooolest experience ever.


requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,121 @@
...
+- **State push** — A new `wizard-state-push.ts` that hooks into `WizardStore.emitChange()` and debounces fire-and-forget POSTs to the PostHog API with the current task list, run phase, and event plan. Zero impact on the TUI — if the push fails, the wizard doesn't notice.
+- **Upfront insight and dashboard planning**: Right now we do it at the end with MCP tools. It should be planned up front when we decide which events to instrument. It can also just be a terraform file and a single sync call.
+- **Upfront event schema planning**: Plan it up front, we can probably push this to the web app for the user to look at.
+- **Split the big query**: Right now the Wizard has a locked down, linear, single query. We can split it into the various natural sections of the wizard's workflow. This will give us deterministic hard breaks we can display in the web app/UI.
+
+### PostHog API
+- **Wizard sessions endpoint** — `POST /api/projects/{id}/wizard/sessions/` to create a session, `PUT .../state/` to push state updates, `GET .../?status=active` to poll for active sessions (for the future). Simple CRUD, state is probably just a JSONB column overwritten on each push. This will not be reused.
```
MattBro: I'd like to consider websockets/SSE instead of polling, I think that'd be simpler and better user experience

gewenyu99: 👍 if supporting socket isn't gonna be a pain I'm game

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Wizard
+- **A new subcommand/option**, maybe `npx @posthog/wizard --sync` that syncs progress to the PostHog API. The wizard will already have an oauth session or API key.
+- **State push** — A new `wizard-state-push.ts` that hooks into `WizardStore.emitChange()` and debounces fire-and-forget POSTs to the PostHog API with the current task list, run phase, and event plan. Zero impact on the TUI — if the push fails, the wizard doesn't notice.
+- **Upfront insight and dashboard planning**: Right now we do it at the end with MCP tools. It should be planned up front when we decide which events to instrument. It can also just be a terraform file and a single sync call.
+- **Upfront event schema planning**: Plan it up front, we can probably push this to the web app for the user to look at.
+- **Split the big query**: Right now the Wizard has a locked down, linear, single query. We can split it into the various natural sections of the wizard's workflow. This will give us deterministic hard breaks we can display in the web app/UI.
```
MattBro: We could probably descope this for v1 since the current hook points give us enough to start with.

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,121 @@
...
+We can split the screen into a progress list of what the wizard is doing, and the configuration steps that are available to the user.
+
+The final step can be taking the user to the dashboards and insights that the wizard created.
+
+## What we need to build to ship a V1
+
+I think we can start a V1, where as the user runs the wizard, we sync the state to the web app so they don't feel completely disjoint. There are many super cool things we can do with this I'll discuss in the potential future work section.
+
+### Wizard
+- **A new subcommand/option**, maybe `npx @posthog/wizard --sync` that syncs progress to the PostHog API. The wizard will already have an oauth session or API key.
```
MattBro: We should make this the default, otherwise most people won't do it

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,121 @@
...
+
+### PostHog API
+- **Wizard sessions endpoint** — `POST /api/projects/{id}/wizard/sessions/` to create a session, `PUT .../state/` to push state updates, `GET .../?status=active` to poll for active sessions (for the future). Simple CRUD, state is probably just a JSONB column overwritten on each push. This will not be reused.
+- **Session expiry** — Mark sessions as abandoned if `run_phase = 'running'` and no update for 10+ minutes.
+
+### PostHog Web App
+- **Wizard-aware onboarding page** — When an active wizard session is detected, replace the generic "Install" step with live wizard progress (task list with pending/in-progress/completed states).
+- **Split-screen onboarding** — Show wizard progress on one side, and configuration steps the user can work through in parallel (Configure, Session replay, Import data, etc.) on the other.
+- **Completion handoff** — When the wizard finishes, surface the dashboards and insights it created, and skip any onboarding steps it already completed.
```
MattBro: > skip any onboarding steps it already completed

this implies the web app understands the concrete steps. Might consider descoping for v1 to keep it more freeform.

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,121 @@
...
+- **Upfront event schema planning**: Plan it up front, we can probably push this to the web app for the user to look at.
+- **Split the big query**: Right now the Wizard has a locked down, linear, single query. We can split it into the various natural sections of the wizard's workflow. This will give us deterministic hard breaks we can display in the web app/UI.
+
+### PostHog API
+- **Wizard sessions endpoint** — `POST /api/projects/{id}/wizard/sessions/` to create a session, `PUT .../state/` to push state updates, `GET .../?status=active` to poll for active sessions (for the future). Simple CRUD, state is probably just a JSONB column overwritten on each push. This will not be reused.
+- **Session expiry** — Mark sessions as abandoned if `run_phase = 'running'` and no update for 10+ minutes.
+
+### PostHog Web App
+- **Wizard-aware onboarding page** — When an active wizard session is detected, replace the generic "Install" step with live wizard progress (task list with pending/in-progress/completed states).
```
fercgomes: I like the idea of, when the Wizard notifies the API that it has begun, it already opens the onboarding page with progress tracking, so there is a seamless transition to the post onboarding flow.

MattBro: What happens when a user finishes onboarding normally and THEN runs the wizard? What if they've already completed steps the wizard is trying to do? Who wins on conflicts?

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Wizard state
+
+Todo items decided by the agent have three states:
+
+```typescript
+export enum TaskStatus {
+  Pending = 'pending',
+  InProgress = 'in_progress',
+  Completed = 'completed',
```
MattBro, line 73: Probably want a failed state here

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,121 @@
...
+- **Upfront insight and dashboard planning**: Right now we do it at the end with MCP tools. It should be planned up front when we decide which events to instrument. It can also just be a terraform file and a single sync call.
+- **Upfront event schema planning**: Plan it up front, we can probably push this to the web app for the user to look at.
+- **Split the big query**: Right now the Wizard has a locked down, linear, single query. We can split it into the various natural sections of the wizard's workflow. This will give us deterministic hard breaks we can display in the web app/UI.
+
+### PostHog API
+- **Wizard sessions endpoint** — `POST /api/projects/{id}/wizard/sessions/` to create a session, `PUT .../state/` to push state updates, `GET .../?status=active` to poll for active sessions (for the future). Simple CRUD, state is probably just a JSONB column overwritten on each push. This will not be reused.
+- **Session expiry** — Mark sessions as abandoned if `run_phase = 'running'` and no update for 10+ minutes.
```
MattBro: Might be a bit long considering most runs are shorter I think?

gewenyu99:
## Notes from our #quick-call

Maybe:
- Create new account, no need to go to the web at all. (email to register with in the wizard and goes right to the setup)
- But they can't pay use :sadge:
- Signup spam?
- Keep in mind we pay for usage.

Sync state to web:
- If we do that sync, we may as well allow following on the browser

Rafa's 3 approaches in UI
- 1. is still fullscreen onboarding, not letting explore, display progress, explain events, also show visuals of what's possible. In the end prompt for things we couldn't before.
- 2. Don't have onboarding anymore, roam around the app, wizard side bar as progress
- 3. Maybe have both, being able to have that in a side bar or smthing, floating window, resumable in full screen.
- Probably build both with a skip button and run an experiment
- Hogtoks while we do wizard run.
- **What if we do a full tiktok feed. Videos are just demos**
- Important to try to capture their attention to build a **mental model**
- @fercgomes has great comments about people really wanting to just skip the installation
- @rafaeelaudibert points out that instead of having people skill and scroll reels, give them reels to scroll.

The important part is just pushing events from wizard to onboarding.

## Action items

For @PostHog/team-growth to follow up on:
- Implementation details on this RFC for the app side!
- Especially important to consider data schema with care so we can continue building on this in the future!

Ask for wizard: 
- Accept commands/tasks from the web app for the future
- Run wizard for runs beyond just a first time run. 
	- Audit runs
	- Do just 1 task/integration
	- Smooth handoff to wizard from web

requests-for-comments/2026-04-01-wizard-web-communication.md
edwinyjlim: @daniloc would this overlap or help with your sandboxed wizard work? if we have a dedicated `/wizard` API endpoint, we can fetch wizard event status and render it anywhere

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,121 @@
...
+
+We hook into the following points:
+
+| Trigger | Wiring |
+|--------|--------|
+| The wizard agent starts | `RunPhase.Running` |
+| The wizard agent shares what it plans to do (event list) | `.posthog-events.json` → `setEventPlan` |
+| The wizard agent creates or updates todos | `TodoWrite` in `handleSDKMessage` → `syncTodos` |
+| The wizard agent is working a todo | `syncTodos` with `status: 'in_progress'` |
+| The wizard agent finishes a todo | `syncTodos` with `status: 'completed'` |
```
edwinyjlim: probably stating the obvious here, but we should make the task emitting service as flexible and reusable as possible

if we can emit task status across any kind of Wizard run – like migrations, specific product installs, troubleshooting, – by supporting arbitrary to-do lists, we’re cooking 

@gewenyu99 @daniloc prob goes hand in hand with the work to make the wizard harness more modular

gewenyu99: Yah definitely. The idea is that we queue up next tasks that's very free form. The in-app onboarding should basically display what ever tasks are being performed and in queue.

for the strict onboarding, we _can_ try to have more structure for consistency on the bigger segments of tasks tho. But the key being that fundamentally, we should just reflect the task queue

requests-for-comments/2026-04-01-wizard-web-communication.md
sarahxsanders: +1

requests-for-comments/2026-04-01-wizard-web-communication.md
sarahxsanders: I think we need to split the big query for a variety of reasons, this just being one of them. I am team split it!

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,121 @@
...
+
+The Wizard needs a pretty good round of refactoring before we can really start messing with its behavior, especially with more input from the web app. But when that's possible, there are many super cool things we can do:
+
+- **Migrations**: If they set up Product Analytics, and we detect they're using a competitor for feature flags, we can 1:1 migrate them. In the web app we can show them how much they'd save if they mirgate. If they would like to try, we can just do it.
+- **Two-way communication** — The wizard could poll the web app for commands, or the web app could push commands to the wizard. This would allow the wizard to wait for user input, or to be paused and resumed by the web app.
+  - Example 1: On web, the user can tweak which events to instrument, what they look like, which insights and dashboards to create, etc.
+  - Example 2: On web, if the user configures remote proxies or makes other config suggestions, we can sync this to the wizard. Maybe via polling.
+- **Test correctness** - We can figure out a way to fire test events from the wizard to the web app to test correctness of the wizard's implementation. This builds confidence.
```
sarahxsanders: Something like a replay-based verification has been on my mind for a bit. The wizard instruments, boots up a dev server + headless browser, performs user flows, and verifies the events are in the web app.

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,121 @@
...
+- **Wizard sessions endpoint** — `POST /api/projects/{id}/wizard/sessions/` to create a session, `PUT .../state/` to push state updates, `GET .../?status=active` to poll for active sessions (for the future). Simple CRUD, state is probably just a JSONB column overwritten on each push. This will not be reused.
+- **Session expiry** — Mark sessions as abandoned if `run_phase = 'running'` and no update for 10+ minutes.
+
+### PostHog Web App
+- **Wizard-aware onboarding page** — When an active wizard session is detected, replace the generic "Install" step with live wizard progress (task list with pending/in-progress/completed states).
+- **Split-screen onboarding** — Show wizard progress on one side, and configuration steps the user can work through in parallel (Configure, Session replay, Import data, etc.) on the other.
+- **Completion handoff** — When the wizard finishes, surface the dashboards and insights it created, and skip any onboarding steps it already completed.
+- **Error state** — When the wizard errors, show the error type and last known state so support can diagnose without terminal screenshots.
+
+## Potential future work
```
sarahxsanders, line 182: some things I've been thinking about when it comes to web app levers:
- code review for existing PH code: review existing integrations to make them more robust, lives in the web app as a process they can trigger 
- diff mode: run the wizard when you first set up, then again a few months later. it has memory of your first run, recognizes you added pages since the last run, and offers to instrument the delta. could be even cooler as a background process watching for new commits.

requests-for-comments/2026-04-01-wizard-web-communication.md
daniloc: I would not be mad about it!

requests-for-comments/2026-04-01-wizard-web-communication.md
gewenyu99: Yeah we don't need this for the V1 version, will decouple

requests-for-comments/2026-04-01-wizard-web-communication.md
gewenyu99: This is highly sus depending on the day. Claude is very slow some days, and larger projects take 10+

requests-for-comments/2026-04-01-wizard-web-communication.md
gewenyu99: Yes, although, this concept doesn't really exist in the Wizard. The agent atm doesn't really understand when it "fails" and chooses to abort. This was just recently added, though

requests-for-comments/2026-04-01-wizard-web-communication.md
gewenyu99: Do we even bother showing onboarding twice? We can just let the wizard run and have the web app ignore its updates. I _assume_ people don't want to see onboarding twice. (We certainly don't have any button to redo-onboarding.)

requests-for-comments/2026-04-01-wizard-web-communication.md
gewenyu99: Yeah we can descope. The list of tasks is also decided by the agent, which will vary run to run. I think that's fine and healthy for now.

requests-for-comments/2026-04-01-wizard-web-communication.md
rafaeelaudibert: Besides "failed", a "canceled" state might be a useful distinction. I understand this doesn't make sense at the wizard level, but it will at the DB level.

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,198 @@
...
+| Anything the agent does during the run (plans events, writes todos, changes a todo's status, re-plans) | `TodoWrite` in message triggers → `setTodos()` → `syncTodos()` — full state overwrite |
+| The wizard agent finishes or errors out | `WizardStore` phase change → `RunPhase.Completed` / `RunPhase.Error` → final push |
+
+**Every `TodoWrite` intercept is a full overwrite of the list.** The agent is free to add, remove, rename, reorder, or completely re-plan the todos mid-run, and we just replace whatever we previously pushed with the new list — no diffing, no merging on our side. Consumers should treat each push as the new source of truth.
+
+Within a single run the individual items still transition through the normal lifecycle the agent drives (`pending` → `in_progress` → `completed` or `failed`), so if a todo survives from one push to the next its status moves forward as you'd expect. But nothing guarantees a given todo id exists in the next push at all.
+
+Each hook point emits a generic task-stream update — the onboarding wizard is the first consumer, but the schema itself is not onboarding-specific.
```
rafaeelaudibert, line 95: Can the todos have some structure or are they just unstructured JSON? Thinking of how to model them in the DB

rafaeelaudibert: Ah, it's explained below, forget about this :)

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,198 @@
...
+}
+```
+
+Every run is a new `session_id`. The wizard never updates an old session — re-running `onboarding` + `posthog_integration` is a new row with a newer timestamp, not a mutation of the previous run. Consumers get the current view by picking the latest session for a given `(workflow_id, skill_id)` pair, and keep the older rows as run history.
+
+The wizard only emits `workflow_id: 'onboarding'` payloads for v1, but the schema is designed so additional workflows and skills can plug in later without a migration.
+
+### What the wizard will ship for V1
+
+- **Sync on by default** — whenever the wizard has an OAuth session or API key, it pushes to the PostHog API with no extra flag needed. `--no-sync` is the escape hatch for users who explicitly want to stay offline.
```
rafaeelaudibert: "Stay offline" isn't really possible because of the LLM, right? Maybe we should just have a very simple `--no-telemetry` or `POSTHOG_TELEMETRY=false` instead?

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,198 @@
...
+- **Task stream push module** — a new `task-stream-push.ts` that hooks into `WizardStore.emitChange()` and debounces fire-and-forget pushes with the current task list, run phase, event plan, and `run_type`. Zero impact on the TUI — if the push fails, the wizard doesn't notice. Onboarding is the first consumer, but the module stays generic so future migration / audit / single-task runs reuse it.
```
rafaeelaudibert, line 127: One thing we'll need here is to make sure we include timestamps here, and latest update wins.

requests-for-comments/2026-04-01-wizard-web-communication.md
gewenyu99: 👍 Will do

requests-for-comments/2026-04-01-wizard-web-communication.md
gewenyu99: Good call

requests-for-comments/2026-04-01-wizard-web-communication.md
gewenyu99: Done

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,200 @@
...
+---
+
+## Part 2 — Suggested: PostHog web app onboarding handling
+
+This part is a suggested implementation as discussed in the #quick-call with the growth team, not a commitment.
+
+### Suggested API shape
+
+- **Wizard sessions endpoint** — `POST /api/projects/{id}/wizard/sessions/` to create a session and persist the final state. State can be a JSONB column overwritten on each push. Since the agent will not emit a concrete list of todos, we'll need to persist the entire task stream in the session. This _may_ change in the future as we rework task queue. It _may_ become a known, fixed, finite list of todos.
+- **Push transport over WebSockets/SSE** — Instead of the web app polling for active sessions, state updates stream to connected clients. The CRUD endpoint remains for session create and final-state persistence. Simpler end-to-end, and unblocks rendering wizard progress anywhere in the app (see Edwin / daniloc's sandboxed wizard idea).
```
fercgomes, line 146: afaik, we don't have any infra for WS, just SSE. @rafaeelaudibert may know more?

rafaeelaudibert: We have both, but this sounds more like SSE territory

requests-for-comments/2026-04-01-wizard-web-communication.md
```diff
@@ -0,0 +1,200 @@
...
+
+---
+
+## Part 2 — Suggested: PostHog web app onboarding handling
+
+This part is a suggested implementation as discussed in the #quick-call with the growth team, not a commitment.
+
+### Suggested API shape
+
+- **Wizard sessions endpoint** — `POST /api/projects/{id}/wizard/sessions/` to create a session and persist the final state. State can be a JSONB column overwritten on each push. Since the agent will not emit a concrete list of todos, we'll need to persist the entire task stream in the session. This _may_ change in the future as we rework task queue. It _may_ become a known, fixed, finite list of todos.
```
fercgomes, line 145: If this will overwrite the current session state, doesn't it make more sense for this to be a PATCH?

gewenyu99: True!

PR #530, opened 2026-04-24
Update launch plan

joethreepwood:
Adds sales enablement item. Idea is for us to reference the enablement for what items will look like. 

More info: https://github.com/PostHog/requests-for-comments-public/issues/512


Lizzieepton (approved):
Sounds good to me

cleo-pleurodon (approved):

PR #534, opened 2026-04-24
Better SDKs for LLMs

gewenyu99:
Can we make SDKs better for LLMs? Here are quirks we found the Wizard and other LLM tools commonly tripping over.

I know many of these are not worth fixing and are very very nit. But this is just to surface the problems we noticed!

### DIFF
--- a/requests-for-comments/2026-04-23-better-sdk-for-llms.md
+++ b/requests-for-comments/2026-04-23-better-sdk-for-llms.md
@@ -1,0 +1,346 @@
+# Request for comments: How to make PostHog SDKs better for LLMs
+
+## Problem statement
+
+LLMs are becoming the default interface for developers to interact with PostHog, includingh PostHog's SDKs. There are various quirks of the PostHog SDKs that make it harder for LLMs (and humans) to implement PostHog correctly the first time. This can add friction for LLMs implementing PostHog.
+
+It's important to highlight that in practice, LLMs will recover from most of these issues. But it may take a few turns (API round trips) to do so. On long conversations with context windows in the hundreds of thousands, this makes it significantly slower and increases cost.
+
+This RFC covers the following:
+- Catalogs the inconsistencies we've hit, grouped by area, and marks each with how it should be fixed.
+- Proposes various action items to fix the inconsistencies.
+- Proposes a new "commandments" style of documentation and skill to guide LLMs on how to use the SDKs correctly.
+
+Many of these are very nit-picky. This RFC is really to surface the problems, I acknowledge that some of these are not worth the effort to fix.
+
+## Proposed action items
+
+In the [problem areas](#problem-areas) section, we've identified a number of problems that we've found necessary to explicitly explain to LLMs to make them work correctly. 
+- Some of these can be fixed with tweaks to the SDKs
+- Many of these are just language-specific quirks that we can't engineer around
+
+What I propose is to do two things. Obviously, I'll suggest areas to improve the SDKs themselves. I also propose that we move some of the "commandments" that are special notes for LLMs to not fumble into the SDK repos themselves. I really think some of these "commandments" for LLMs to read are going to stay, but we should move them to closer to the SDKs and load them into the docs and skills during build.
+
+Many of these are very nit-picky and are not worth the effort to fix. At least we'll ship some commandments for LLMs to read to help them avoid the quirks.
+
+## Problem areas
+
+Here are some problem areas we've identified while building the PostHog Wizard and various skills. These are problems that we've found necessary to explicitly explain to LLMs to make them work correctly. LLMs (specifically Claudes's various models) frequently trip over these areas when implementing PostHog.
+
+Within each section, entries are ordered from highest to lowest LLM-tripping severity — silent failures (no error surfaces, LLM has no feedback signal) are listed before loud failures (clear errors the LLM can recover from in a round or two).
+
+**Legend** proposed solutions:
+- `[DOCS]` fix with better docs
+- `[SDK]` fix with code
+- `[BOTH]` needs both
+
+### Naming and config shape
+
+#### Feature-flag timeout has five names and two units
+
+ `feature_flag_request_timeout_ms` (JS/Node/PHP), `featureFlagsRequestTimeoutMs` (RN), `feature_flags_request_timeout_seconds` (Python), `feature_flag_request_timeout_seconds` (Ruby). Default 3 s everywhere except RN (10 s).
+
+**Issue:** LLMs struggle with synonyms in general. LLMs may take a round or two to figure out which option is the correct one.
+
+**Suggested solution:** `[SDK]` Settle on one option name and one unit (suffixed with `_ms` or `_seconds`), align the default to 3 s, and keep the deprecated names as aliases.
+
+#### Host option has four names
+
+`[SDK]` `api_host` (JS/React/Next), `host` (Node/Python/Ruby/PHP/RN/iOS/Android/Flutter), `Endpoint` (Go), `HostUrl`/`Host` (.NET), `POSTHOG_HOST` (Flutter manifest). `host` is dominant — deprecate the rest.
+
+**Issue:** LLMs will often hallucinate and use the wrong option name. It will easily catch it later, but waste 2-3 rounds reading SDK code to find the correct option name. This can get expensive quickly.
+
+**Suggested solution:** `[SDK]` Find a single, common name for the host option. Keep the deprecated options for backwards compatibility.
+
+#### `sendFeatureFlags` vs `sendFeatureFlagEvents` collision
+
+Two unrelated options that sound identical:
+- `sendFeatureFlags` / `send_feature_flags`: per-capture, attach matched flags to the event.
+- `sendFeatureFlagEvents` / `send_feature_flag_events` / `sendFeatureFlagEvent` (singular on Android/RN/iOS): per-SDK, auto-emit `$feature_flag_called` on flag read.
+
+**Issue:** This is confusing for humans and LLMs alike. LLMs may take a round or two of extra thinking to distinguish between the options.
+
+**Suggested solution:** `[DOCS]` Document the distinction with a side-by-side example on every SDK's config page, and `[SDK]` consider renaming one to remove the collision.
+
+#### `personProfiles` has five encodings
+`[DOCS]` `'identified_only'` (JS/RN string), `.identifiedOnly` (Swift), `PersonProfiles.IDENTIFIED_ONLY` (Kotlin), no option on server SDKs (use `$process_person_profile: false` per event). Idiomatic per language — needs one comparison table.
+
+**Issue:** LLMs porting config between SDKs will reuse the JS string form (`'identified_only'`) on Swift or Kotlin. Catchable at compile time, but wastes a round.
+
+**Suggested solution:** `[DOCS]` Add a single cross-SDK comparison table showing the `personProfiles` encoding per SDK and the server-side `$process_person_profile` equivalent.
+
+#### Ruby gem name ≠ require name
+`[BOTH]` `gem 'posthog-ruby'` requires `require 'posthog'`, not `require 'posthog-ruby'`. Either alias the require or put the right line in the quickstart.
+
+**Issue:** LLMs commonly write `require 'posthog-ruby'` to match the gem name. Loud error, but a near-universal first-attempt trip.
+
+**Suggested solution:** `[BOTH]` Alias `require 'posthog-ruby'` to `require 'posthog'` and update the Ruby quickstart to show the correct line.
+
+#### `evaluation_contexts` / `evaluation_environments` both active
+Legacy name works across all SDKs and the `/flags` API. Autocomplete surfaces both. Either set a deprecation flag or document the alias.
+
+**Issue:** Both names work, so LLMs pick whichever matches their training data. No immediate failure, but generated code is inconsistent across a codebase.
+
+**Suggested solution:** `[DOCS]` Pick one name as canonical, document the alias, and set a removal date for the other.
+
+#### snake_case and camelCase mixed within a single SDK
+
+`[SDK]` `posthog-js` ships `get_distinct_id`, `set_config`, `opt_out_capturing` next to `getFeatureFlag`, `isFeatureEnabled`, `captureException`, `startSessionRecording`. React Native inherits both and adds its own.
+
+**Issue:** Low severity, but sometimes LLMs will hallucinate and use the wrong option name. It will easily catch it later, but waste 2-3 rounds reading SDK code to find the correct option name. This can get expensive quickly.
+
+**Suggested solution:** `[SDK]` (very low priority) Avoid mixing casing within a single SDK.
+
+---
+
+### Capture call shapes
+
+#### Python `capture` treats first positional arg as `distinct_id`
+`posthog.capture('event_name', {...})` silently becomes `distinct_id='event_name'` with the props dict as the event name. No error.
+
+**Issue:** Many silent failures. The code appears to work, no exception is raised, but every event is sometimes malformed (event name stored as distinct_id). LLMs have no signal to correct themselves unless the user notices downstream.
+
+**Suggested solution:** `[DOCS]` Call this out explicitly in the Python quickstart, and emit a warning when the first positional arg looks like an event name (contains dots, starts with `$`, etc) `[SDK]` I wonder if we can standardize the shape of the server SDKs to either all use context for identity or share the same shape for distinct_id first? Maybe named objects as the expected capture call shape.
+
+#### Person properties have three incompatible shapes
+- `$set` / `$set_once` inside `properties` (JS/Node/Python/Ruby/PHP/Go/Flutter/.NET)
+- Top-level `userProperties` / `userPropertiesSetOnce` args (Android/iOS/RN)
+- `personPropertiesToSet` / `personPropertiesToSetOnce` (.NET `IdentifyAsync`)
+
+Porting JS `{ $set: {...} }` to iOS silently makes `$set` a regular event property.
+
+**Issue:** LLMs translating a JS `$set` snippet to iOS/Android/RN silently produce wrong data — `$set` becomes a regular event property and person properties never update. No error at call site. LLMs see a ton of JS examples and sometimes will infer the wrong shape.
+
+**Suggested solution:** `[DOCS]` Add a cross-SDK comparison table for setting person properties, and call out the iOS/Android/RN top-level args explicitly.
+
+#### Ruby `capture_exception` is positional while `capture` is hash-based
+`[SDK]` `client.capture_exception(exception, distinct_id, additional_properties)`. Passing `distinct_id:` as a kwarg silently binds it into `additional_properties`. Align on hash args like `capture`.
+
+**Issue:** If the LLM passes `distinct_id:` as a kwarg (natural given `capture`'s hash shape), Ruby silently folds it into `additional_properties`. No error; events attribute to the wrong user.
+
+**Suggested solution:** `[SDK]` Align `capture_exception` on hash args like `capture`, and warn when kwargs bind into `additional_properties` unexpectedly.
+
+#### PHP uses camelCase keys inside PHP arrays
+`PostHog::capture(['distinctId' => ..., 'event' => ..., 'properties' => [...]])`. Not `distinct_id`. Unintuitive for PHP devs.
+
+**Issue:** LLMs default to snake_case (`distinct_id`, matching the rest of the ecosystem); PHP silently drops those keys. No error; events look fine but distinct_id is missing.
+
+**Suggested solution:** `[DOCS]` Document the camelCase keys explicitly on the PHP commandments and quickstart docs. Maybe warn when snake_case keys are used.
+
+#### Python has no `identify()` method
+`[SDK]` Every other SDK does. `posthog.identify(...)` raises `AttributeError`. Use `set()` or `identify_context()` inside `new_context()`.
+
+**Issue:** LLMs assume `posthog.identify(...)` exists because every other SDK has it. Python raises `AttributeError` — loud, catchable, but wastes a round.
+
+**Suggested solution:** `[SDK]` Add a `posthog.identify()` method to the Python SDK that wraps `set()` / `identify_context()` so the API matches every other SDK.
+
+---
+
+### identify / alias / reset
+
+#### Invalid distinct_ids return 200 OK with a refused-merge warning
+`null`, `true`, `"distinct_id"`, `""`, etc.: event ingests attached to one side only, warning logged in the ingestion UI. SDKs should reject at call site.
+
+**Issue:** LLMs generating `distinct_id='distinct_id'` (literal placeholder) or `distinct_id=None` get 200 OK with no exception. Maybe at least for the common cases, we can set some warnings.
+
+**Suggested solution:** `[SDK]` Validate and warn on invalid distinct_ids (null, booleans, empty strings, literal `"distinct_id"`) at call site with a clear error.
+
+#### `identify()` means different things client-side vs server-side
+Client SDKs: associates all future events with the distinct_id. Server SDKs: updates persons properties only; every capture still needs `distinct_id` explicitly. Scattered in docs; belongs in every server-side quickstart.
+
+Also, calling Identify on the server side DOES NOT establish a stable ID, because it doesn't merge the anonymous ID with a stable ID (one doesn't exist on the server). Identify actually works exactly like alias, it seems, on the server.
+
+**Issue:** LLMs reuse client-side identify patterns on the server — one `identify()` call, then bare `capture()`s. Server-side, distinct_id is required per capture; events silently attribute to the wrong user or get dropped.
+
+**Suggested solution:** Maybe we should all use the identify_context pattern like python everywhere.
+
+#### `alias()` argument order and shape varies
+- JS: `alias(newAlias, existingId)`
+- Node/Ruby: `{ distinctId, alias }`
+- Python: `alias(previous_id=newAlias, distinct_id=existingId)` — `previous_id` exists only here
+- Go: `Alias{ DistinctId, Alias }` struct
+- Mobile: `alias(newAlias)`, current identity implicit
+
+**Issue:** Argument order inverts between JS (`newAlias, existingId`) and Python's explicit form. LLMs mixing SDKs will merge the wrong direction and can irreversibly join two unrelated users.
+
+**Suggested solution:** Maybe we should standardize this :P
+
+#### `reset(true)` vs `reset()` scope differs per SDK
+Web: `true` also resets device_id. RN: `reset()` clears feature-flag cache. No table.
+
+**Issue:** LLMs assume `reset()` is symmetrical across SDKs. Porting logout flows misses that web needs `reset(true)` to clear device_id or that RN also purges the flag cache. Hard to catch without behavioral testing.
+
+**Suggested solution:** `[DOCS]` Add a comparison table showing exactly what `reset()` and `reset(true)` clear on each SDK.
+
+#### `$merge_dangerously` is the only escape for two already-identified users
+`[DOCS]` Irreversible. Not a first-class SDK method. Needs a dedicated troubleshooting page.
+
+**Issue:** Not a first-class method, so LLMs rarely know it exists. Asked to merge two identified users, they'll propose `identify()` or `alias()` and silently fail.
+
+**Suggested solution:** `[DOCS]` Add a dedicated troubleshooting page for recovering already-identified users, including the `$merge_dangerously` caveats. Maybe we should also rethink how merging really works. It's super confusing to also know which users are truly already identified.
+
+---
+
+### Defaults that differ silently
+
+#### `flushInterval` unit varies
+ Node/RN: ms (default 10000). Python: seconds (default 0.5). iOS: seconds (default 30). Android: `flushIntervalSeconds` (30). Go: `time.Duration`. Pasting `flushInterval: 30` between SDKs changes the rate 1000×. Suffix the name with the unit (Android does).
+
+**Issue:** Copy-pasting `flushInterval: 30` from Node to Python changes the flush rate by 1000×. No warning; the code runs. LLMs pattern-match on option names and miss the unit sometimes, resulting in difficult to debug issues.
+
+**Suggested solution:** `[SDK]` Suffix the option name with its unit (`flushIntervalMs` / `flushIntervalSeconds`) everywhere, and keep the legacy names as aliases.
+
+#### Autocapture defaults diverge on mobile
+`[DOCS]` iOS: lifecycle + screen views on, `captureElementInteractions` off. RN: `PostHogProvider` needs `autocapture` passed explicitly. Android: lifecycle + screens on. Web: everything on.
+
+**Issue:** LLMs assume autocapture is either "on" or "off" across platforms. The per-platform subfeature matrix is not intuitive; generated code over- or under-captures silently.
+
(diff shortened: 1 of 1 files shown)


requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+**Issue:** LLMs writing experiment code tend to reach for `getFeatureFlagPayload` because "payload" sounds most specific. That method doesn't emit the exposure event, so the experiment has zero data. Silent and severe.
+
+**Suggested solution:** `[DOCS]` Document the exact set of methods that emit `$feature_flag_called` on each flag/experiment reference page and call out the experiment-exposure implication.
+
+#### `useFeatureFlag` (RN) returns `undefined` during load; React returns `false`
+`[DOCS]` Same-looking code, different branches. RN has three states (`undefined`/`false`/`true`), React has two.
+
+**Issue:** LLMs port `if (flag) {...}` from React to RN. In RN, `undefined` is a third state and the loading branch silently takes the `false` path. Surfaces as flicker or wrong default treatment.
+
+**Suggested solution:** ?
```
dustinbyrne, line 216: Typically undefined/nullish types returned by a flags API indicates the flag is not found, rather than evaluating to false. I think we should standardize to `undefined`/`false`/`true` here.

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+---
+
+### Framework integration
+
+#### Next.js: `PostHogProvider` + `instrumentation-client.ts` creates two instances
+`[BOTH]` Singleton initialized in `instrumentation-client.ts` vs. provider-wrapped instance are independent. `posthog.capture(...)` goes to one, hooks to the other. Session replay fragments. Flag hooks already fall through to the singleton — the provider is redundant for most apps.
+
+**Issue:** LLMs combine two common quickstart patterns (`instrumentation-client.ts` + `PostHogProvider`) and silently create two SDK instances. Session replay fragments; `posthog.capture(...)` and hooks hit different instances. Very hard to notice.
+
+**Suggested solution:** (? I don't really know but maybe) `[BOTH]` Make `PostHogProvider` detect and reuse the `instrumentation-client.ts` singleton, and update the Next.js quickstart to show one instance only.
```
dustinbyrne, line 286: I believe this is a non-issue in `@posthog/next`, though it hasn't really been publicized much yet due to it's maturity

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+`[DOCS]` Imports from `layout.tsx`, `+layout.svelte`, or SSR Astro pages evaluate on the server; `window` is undefined. Needs a generic "SSR safety" note.
+
+**Issue:** LLMs import `posthog-js` at the top of `layout.tsx` or `+layout.svelte`. `window` is undefined on the server; build or first render fails.
+
+**Suggested solution:** `[DOCS]` Add a generic "SSR safety" note covering Next.js, SvelteKit, and Astro, with the dynamic-import / `typeof window` patterns.
+
+#### `@posthog/react`: `usePostHog()` can return `undefined`
+`[DOCS]` Provider renders children before init. Optional-chain everywhere or gate on the value.
+
+**Issue:** LLMs assume `usePostHog()` always returns a client. First-render guards are missing; null-pointer errors when a hook runs before the provider initializes.
```
dustinbyrne, line 319: This is tricky, because we explicitly cast away the `undefined` even if it actually _is_ undefined. Personally I'd like to see the return type accurately reflect `PostHog | undefined`.

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+
+#### Django: custom middleware on top of `PosthogContextMiddleware` double-tags
+The SDK already ships middleware. Initialize in `AppConfig.ready()`, not `settings.py`.
+
+**Issue:** LLMs write a custom request-tagging middleware on top of `PosthogContextMiddleware`. Events are double-tagged. Not immediately obvious in dashboards.
+
+**Suggested solution:** `[DOCS]` Document that the SDK ships middleware and that init belongs in `AppConfig.ready()`, not `settings.py`.
+
+#### Rails has no context middleware
+`[SDK]` Django's `PosthogContextMiddleware` reads session/distinct tracing headers. `posthog-rails` has no equivalent. Rails devs porting from Django end up with uncorrelated frontend/backend events.
```
dustinbyrne, line 331: This is something I'm actively working on

dustinbyrne:
Very detailed write up, thank you for this! I've read through it once and I generally agree with what's written here. I'll take another pass over this early next week.

requests-for-comments/2026-04-23-better-sdk-for-llms.md
marandaneto: or just return a NoOp PostHog instance if undefined which would be back compatible

marandaneto:
@ioannisj @turnipdabeets would love your eyes here as well

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+
+### Naming and config shape
+
+#### Feature-flag timeout has five names and two units
+
+ `feature_flag_request_timeout_ms` (JS/Node/PHP), `featureFlagsRequestTimeoutMs` (RN), `feature_flags_request_timeout_seconds` (Python), `feature_flag_request_timeout_seconds` (Ruby). Default 3 s everywhere except RN (10 s).
+
+**Issue:** LLMs struggle with synonyms in general. LLMs may take a round or two to figure out which option is the correct one.
+
+**Suggested solution:** `[SDK]` Settle on one option name and one unit (suffixed with `_ms` or `_seconds`), align the default to 3 s, and keep the deprecated names as aliases.
```
marandaneto, line 45: mobile needs longer timeouts due to more unstable internet

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+
+#### Host option has four names
+
+`[SDK]` `api_host` (JS/React/Next), `host` (Node/Python/Ruby/PHP/RN/iOS/Android/Flutter), `Endpoint` (Go), `HostUrl`/`Host` (.NET), `POSTHOG_HOST` (Flutter manifest). `host` is dominant — deprecate the rest.
```
marandaneto, line 49: sometimes you need a prefix because the word is reserved or ambiguous, eg a manifest needs a prefix otherwise it'll clash with other stuff

requests-for-comments/2026-04-23-better-sdk-for-llms.md
marandaneto: some languages have a `Duration` type and its not just a number, in this case it'd be just `feature_flag_request_timeout` without a suffix because you specify when setting a value like `= Duration(3, Seconds)`

requests-for-comments/2026-04-23-better-sdk-for-llms.md
marandaneto: agree to keep a single common name if possible

requests-for-comments/2026-04-23-better-sdk-for-llms.md
marandaneto: agree to keep a single common name if possible
the snake case or camel case depends on the language/sdk convention

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+
+**Suggested solution:** `[SDK]` Find a single, common name for the host option. Keep the deprecated options for backwards compatibility.
+
+#### `sendFeatureFlags` vs `sendFeatureFlagEvents` collision
+
+Two unrelated options that sound identical:
+- `sendFeatureFlags` / `send_feature_flags`: per-capture, attach matched flags to the event.
+- `sendFeatureFlagEvents` / `send_feature_flag_events` / `sendFeatureFlagEvent` (singular on Android/RN/iOS): per-SDK, auto-emit `$feature_flag_called` on flag read.
+
+**Issue:** This is confusing for humans and LLMs alike. LLMs may take a round or two of extra thinking to distinguish between the options.
```
marandaneto, line 61: yeah not all SDKs have the global option, this is mainly for autocaptured events/flags evaluation that you dont have the chance to pass your own choice eg enabling session replay behind a flag, surveys, etc that you dont call `capture|getFeatureFlag` yourself so you need to enable or disable globally otherwise there's no opt out unless you disable the whole feature

agree with the renaming singular/plural and with or without the 'events' suffix

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+
+**Suggested solution:** `[DOCS]` Document the distinction with a side-by-side example on every SDK's config page, and `[SDK]` consider renaming one to remove the collision.
+
+#### `personProfiles` has five encodings
+`[DOCS]` `'identified_only'` (JS/RN string), `.identifiedOnly` (Swift), `PersonProfiles.IDENTIFIED_ONLY` (Kotlin), no option on server SDKs (use `$process_person_profile: false` per event). Idiomatic per language — needs one comparison table.
```
marandaneto, line 66: I think server side SDKs should add a `personProfiles` enum equivalent since $ properties arent human friendly

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+**Issue:** LLMs commonly write `require 'posthog-ruby'` to match the gem name. Loud error, but a near-universal first-attempt trip.
+
+**Suggested solution:** `[BOTH]` Alias `require 'posthog-ruby'` to `require 'posthog'` and update the Ruby quickstart to show the correct line.
+
+#### `evaluation_contexts` / `evaluation_environments` both active
+Legacy name works across all SDKs and the `/flags` API. Autocomplete surfaces both. Either set a deprecation flag or document the alias.
+
+**Issue:** Both names work, so LLMs pick whichever matches their training data. No immediate failure, but generated code is inconsistent across a codebase.
+
+**Suggested solution:** `[DOCS]` Pick one name as canonical, document the alias, and set a removal date for the other.
```
marandaneto, line 84: we dont do dates, we'd remove all deprecated fields when we cut a new major version and this is done as needed

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+**Issue:** LLMs porting config between SDKs will reuse the JS string form (`'identified_only'`) on Swift or Kotlin. Catchable at compile time, but wastes a round.
+
+**Suggested solution:** `[DOCS]` Add a single cross-SDK comparison table showing the `personProfiles` encoding per SDK and the server-side `$process_person_profile` equivalent.
+
+#### Ruby gem name ≠ require name
+`[BOTH]` `gem 'posthog-ruby'` requires `require 'posthog'`, not `require 'posthog-ruby'`. Either alias the require or put the right line in the quickstart.
+
+**Issue:** LLMs commonly write `require 'posthog-ruby'` to match the gem name. Loud error, but a near-universal first-attempt trip.
+
+**Suggested solution:** `[BOTH]` Alias `require 'posthog-ruby'` to `require 'posthog'` and update the Ruby quickstart to show the correct line.
```
marandaneto, line 77: @gewenyu99 what do you mean by the ruby quickstart?
i just found this one using the wrong import https://github.com/PostHog/posthog/pull/56482

marandaneto: agree with the alias

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+
+**Suggested solution:** `[DOCS]` Pick one name as canonical, document the alias, and set a removal date for the other.
+
+#### snake_case and camelCase mixed within a single SDK
+
+`[SDK]` `posthog-js` ships `get_distinct_id`, `set_config`, `opt_out_capturing` next to `getFeatureFlag`, `isFeatureEnabled`, `captureException`, `startSessionRecording`. React Native inherits both and adds its own.
+
+**Issue:** Low severity, but sometimes LLMs will hallucinate and use the wrong option name. It will easily catch it later, but waste 2-3 rounds reading SDK code to find the correct option name. This can get expensive quickly.
+
+**Suggested solution:** `[SDK]` (very low priority) Avoid mixing casing within a single SDK.
```
marandaneto, line 92: for SDKs that already have both cases, we should add a linter and stick to one
deprecating the 'wrong ones' and adding the new ones with the canonical case may increase the problem tho, so not sure if we should do this

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+---
+
+### Capture call shapes
+
+#### Python `capture` treats first positional arg as `distinct_id`
+`posthog.capture('event_name', {...})` silently becomes `distinct_id='event_name'` with the props dict as the event name. No error.
+
+**Issue:** Many silent failures. The code appears to work, no exception is raised, but every event is sometimes malformed (event name stored as distinct_id). LLMs have no signal to correct themselves unless the user notices downstream.
+
+**Suggested solution:** `[DOCS]` Call this out explicitly in the Python quickstart, and emit a warning when the first positional arg looks like an event name (contains dots, starts with `$`, etc) `[SDK]` I wonder if we can standardize the shape of the server SDKs to either all use context for identity or share the same shape for distinct_id first? Maybe named objects as the expected capture call shape.
```
marandaneto, line 103: > I wonder if we can standardize the shape of the server SDKs to either all use context for identity or share the same shape for distinct_id first

not sure i understand here, do you mean making backend SDKs stateful instead of stateless so they dont need to pass distinct id at all? only when they call identify?

not all languages have named objects, but for sure better docs/method documentation would help

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+
+#### Person properties have three incompatible shapes
+- `$set` / `$set_once` inside `properties` (JS/Node/Python/Ruby/PHP/Go/Flutter/.NET)
```
marandaneto, line 106: also similar to https://github.com/PostHog/requests-for-comments-public/pull/534/changes#r3146688842

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+**Issue:** LLMs default to snake_case (`distinct_id`, matching the rest of the ecosystem); PHP silently drops those keys. No error; events look fine but distinct_id is missing.
+
+**Suggested solution:** `[DOCS]` Document the camelCase keys explicitly on the PHP commandments and quickstart docs. Maybe warn when snake_case keys are used.
+
+#### Python has no `identify()` method
+`[SDK]` Every other SDK does. `posthog.identify(...)` raises `AttributeError`. Use `set()` or `identify_context()` inside `new_context()`.
+
+**Issue:** LLMs assume `posthog.identify(...)` exists because every other SDK has it. Python raises `AttributeError` — loud, catchable, but wastes a round.
+
+**Suggested solution:** `[SDK]` Add a `posthog.identify()` method to the Python SDK that wraps `set()` / `identify_context()` so the API matches every other SDK.
```
marandaneto, line 135: surprised python does not have identify

marandaneto:
very good examples of how we can improve things here
Lots of them, I believe were hit because we are porting the JS SDK skill/docs to every other SDK which do not always match 1:1 or even have very different behavior
A few things that I believe would move the needle here are

* fix/improve many of the items listed here
  * focus on silent fails, the ones that take more turns to get fixed, etc
  * avoid breaking changes
  * avoid increasing the public API because it can get worse
* having a skill per SDK
  * example https://github.com/getsentry/sentry-for-ai/blob/main/skills/sentry-flutter-sdk/SKILL.md	
* all public methods and configurations should have proper description, explanation, code snippet example and link to docs
  * example https://github.com/PostHog/posthog-js/blob/e7ab0cebcbdd8bf60aa2290b987ad090fa87a7c7/packages/browser/src/posthog-core.ts#L2338-L2378
* skill per SDK instead of ported from a canonical skill (eg JS SDK)
* plan a new major version for all SDKs with an API convention (so SDKs align in behavior, naming convention, etc)
  * requires designing a new public API
  * migration plan
  * quite a bit of work

requests-for-comments/2026-04-23-better-sdk-for-llms.md
ioannisj: Agreed with the points above. I think language convention are important and sometimes different defaults make sense for different platforms. Would comments or a matric on related config per SDK help the LLM here maybe?

requests-for-comments/2026-04-23-better-sdk-for-llms.md
ioannisj: At least this would mark them for removal on next major though

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+
+**Issue:** Low severity, but sometimes LLMs will hallucinate and use the wrong option name. It will easily catch it later, but waste 2-3 rounds reading SDK code to find the correct option name. This can get expensive quickly.
+
+**Suggested solution:** `[SDK]` (very low priority) Avoid mixing casing within a single SDK.
+
+---
+
+### Capture call shapes
+
+#### Python `capture` treats first positional arg as `distinct_id`
```
ioannisj, line 98: Looks like it has been fixed since? Based on the comment [here](https://github.com/PostHog/posthog-python/blob/3bd841f6a5f9a145d6cbb54edfa1072b427b8e81/posthog/__init__.py#L272)

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+
+#### `flushInterval` unit varies
+ Node/RN: ms (default 10000). Python: seconds (default 0.5). iOS: seconds (default 30). Android: `flushIntervalSeconds` (30). Go: `time.Duration`. Pasting `flushInterval: 30` between SDKs changes the rate 1000×. Suffix the name with the unit (Android does).
+
+**Issue:** Copy-pasting `flushInterval: 30` from Node to Python changes the flush rate by 1000×. No warning; the code runs. LLMs pattern-match on option names and miss the unit sometimes, resulting in difficult to debug issues.
+
+**Suggested solution:** `[SDK]` Suffix the option name with its unit (`flushIntervalMs` / `flushIntervalSeconds`) everywhere, and keep the legacy names as aliases.
+
+#### Autocapture defaults diverge on mobile
+`[DOCS]` iOS: lifecycle + screen views on, `captureElementInteractions` off. RN: `PostHogProvider` needs `autocapture` passed explicitly. Android: lifecycle + screens on. Web: everything on.
```
ioannisj, line 194: iirc PostHogProvider autocapture is the equivalent of `captureElementInteractions` on iOS

gewenyu99:
I think the SDKs are in good hands :D 

One parting comment is to really consider how LLMs tokenize variable names and function names.

createTotallyCoolApp and create_totally_cool_app both will become some version of these individual tokens:

`create, totally, cool, app` 

Which will then be vectorized. This means makeTotallySickApp with tokens `make, totally, sick, app` would be a pretty confusing "synonym" set of tokens. 

Early versions of GPT really commonly confused similar variables like these because `create` and `make` are similar words ([example of weird behavior](https://news.ycombinator.com/item?id=44689778#:~:text=Obviously%20yes.,consistent%20programmers%20are%20with%20naming)). So when possible, really think about the names you choose both as a whole, and how it can be very similar to another name when vectorized.

It's obviously a pretty gross approximation of how the cool models really do it, but historically LLMs struggled with trivial things like counting the number of `r`s in [strawberry](https://aclanthology.org/2025.naacl-long.172/). Models are getting better, but it's this kind of weird stuff we do need to pay extra attention to during code review. Making models distinguish `person` `persons` `people` `peoples` in variable names ins totally avoidable. You can take a function name/variable name, ask GPT/Claude what it thinks this variable does. If it's way off, you may want to rename it :D

requests-for-comments/2026-04-23-better-sdk-for-llms.md
turnipdabeets: I feel like keeping the language conventions would help the LLM since it's working in the user's code base that expects camel case (for example) and comments could help for different defaults but maybe we can still unify the naming even if there are in different cases. ms vs s and also flag vs flags

requests-for-comments/2026-04-23-better-sdk-for-llms.md
turnipdabeets: Good call out on name clashes, we should probably confirm each platform before locking in the unified name.

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+**Issue:** LLMs will often hallucinate and use the wrong option name. It will easily catch it later, but waste 2-3 rounds reading SDK code to find the correct option name. This can get expensive quickly.
+
+**Suggested solution:** `[SDK]` Find a single, common name for the host option. Keep the deprecated options for backwards compatibility.
+
+#### `sendFeatureFlags` vs `sendFeatureFlagEvents` collision
+
+Two unrelated options that sound identical:
+- `sendFeatureFlags` / `send_feature_flags`: per-capture, attach matched flags to the event.
+- `sendFeatureFlagEvents` / `send_feature_flag_events` / `sendFeatureFlagEvent` (singular on Android/RN/iOS): per-SDK, auto-emit `$feature_flag_called` on flag read.
+
```
turnipdabeets, line 60: This makes me think of the logs too. There's the new logs feature coming but we also have "logs" in session replay as well. `capture` seems to be a naming we use to capture events and so the natural name for logs would be `captureLogs` but it's already being used by session replay which might confuse the LLM to think logs are being captured or set up but really it's session replay. Obviously it'll be easier to use a different naming across the board on a new feature vs renaming session replay but it may be confusing either way.

requests-for-comments/2026-04-23-better-sdk-for-llms.md
turnipdabeets: What about wrapping a shared config with idiomatic configs? Then the idiomatic config can have the correct casing plus ignore optional fields that it doesn't use. This would still require a major release though.

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+**Issue:** Not a first-class method, so LLMs rarely know it exists. Asked to merge two identified users, they'll propose `identify()` or `alias()` and silently fail.
+
+**Suggested solution:** `[DOCS]` Add a dedicated troubleshooting page for recovering already-identified users, including the `$merge_dangerously` caveats. Maybe we should also rethink how merging really works. It's super confusing to also know which users are truly already identified.
+
+---
+
+### Defaults that differ silently
+
+#### `flushInterval` unit varies
```
turnipdabeets, line 186: I wonder if having a documented const file with the different defaults would help. All in one place with names like FLUSH_DEFAULT_RN etc... at least in posthog-js since it handles many SDKs

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+**Issue:** LLMs cherry-pick `PostHog::Client.new(...)` examples inside a Rails app, bypassing `PostHog::Rails.configure`. `auto_capture_exceptions` and middleware hooks silently stop working.
+
+**Suggested solution:** `[DOCS]` Document the Rails-vs-plain-Ruby difference and warn that `PostHog::Client.new` inside Rails bypasses `PostHog::Rails.configure`.
+
+#### React Native: `PostHogProvider` must be inside `NavigationContainer`
+`[DOCS]` Counter to the "providers at the top" convention. React Navigation v7 hooks don't fire otherwise.
+
+**Issue:** "Providers at the top" is strong convention. LLMs default to placing `PostHogProvider` above `NavigationContainer`; screen-tracking hooks then silently no-op.
+
+**Suggested solution:** `[DOCS]` Call this out explicitly on the RN + React Navigation integration page with a working example? Maybe we could find a better pattern for RN.
```
turnipdabeets, line 307: `working example` - will beefing up the example apps help AI? We'd need evidence, but I did notice that some examples apps have more examples than others.

gewenyu99: @turnipdabeets I don't mean example apps in the SDKs necessarily. 

When we build context for the Wizard, we feed it integration specific example code snippets to help it understand the end shape of certain integrations.

this can be especially helpful for complex stuff like feature flags.

Random examples _do not_ help, it needs to have purpose and be fed to agent as explicit context.

requests-for-comments/2026-04-23-better-sdk-for-llms.md
```diff
@@ -0,0 +1,346 @@
...
+**Suggested solution:** `[DOCS]` Document that the SDK ships middleware and that init belongs in `AppConfig.ready()`, not `settings.py`.
+
+#### Rails has no context middleware
+`[SDK]` Django's `PosthogContextMiddleware` reads session/distinct tracing headers. `posthog-rails` has no equivalent. Rails devs porting from Django end up with uncorrelated frontend/backend events.
+
+**Issue:** Django porters look for a Rails equivalent and find nothing. Not silent — LLMs notice the absence — but they produce subpar generated code without any context propagation.
+
+**Suggested solution:** `[SDK]` Add a `posthog-rails` context middleware equivalent to Django's `PosthogContextMiddleware`.
+
+---
```
turnipdabeets, line 337: We could also try publishing skills/ txt files per SDK so we don't need to wait for making some breaking changes. The trick is keeping that document updated.

requests-for-comments/2026-04-23-better-sdk-for-llms.md
gewenyu99: But yeah, having an example to reference is often a big step up when explicitly provided to agents in our experience 👍 We'd just need to reference it in the skills!

requests-for-comments/2026-04-23-better-sdk-for-llms.md
gewenyu99: Yeah, I think this would help a ton. That is friendly to agent exploration

requests-for-comments/2026-04-23-better-sdk-for-llms.md
gewenyu99: Yeah this can be a tricky one! Luckly, most loggers work like `logger.log()` so we don't necessarily need to clobber the capture word more

requests-for-comments/2026-04-23-better-sdk-for-llms.md
gewenyu99: ahhhh okay, nice

requests-for-comments/2026-04-23-better-sdk-for-llms.md
gewenyu99: I recall writing the changelog here: https://posthog.com/tutorials/python-v6-migration#1-replace-identify-calls

I am not sure if it's gone gone or this was the advice. We also [removed it from the docs](https://posthog.com/docs/libraries/python).

gewenyu99: I think there was push for SDKs to swing toward context based identity, but it wasn't finished

requests-for-comments/2026-04-23-better-sdk-for-llms.md
gewenyu99: Ahhhh, yeah this isn't necessarily possible.

So in python, you'd tag an entire call stack with a `context()` and label through that. 

For example:

```py
from posthog import new_context, tag, set_context_session, identify_context

with new_context():
    tag("transaction_id", "abc123")
    tag("some_arbitrary_value", {"tags": "can be dicts"})

    # Sessions are UUIDv7 values and used to track a sequence of events that occur within a single user session
    # See https://posthog.com/docs/data/sessions
    set_context_session(session_id)

    # Setting the context-level distinct ID. See below for more details.
    identify_context(user_id)

    # calls a function down the call stack
   	do_something()
 
def do_something():
	# This event is captured with the distinct ID, session ID, and tags set above
    posthog.capture("order_processed")
```

I'm not sure this makes sense either, this was the push behind the [contexts here](https://posthog.com/tutorials/python-v6-migration#new-feature-contexts)

requests-for-comments/2026-04-23-better-sdk-for-llms.md
gewenyu99: Ahhh when I say quick start anything, it's usually referencing wizard things. This was aggregated mostly from wizard examples

https://github.com/PostHog/context-mill/blob/f1c7b89795377d7f086513d342c774eb160c653d/basics/ruby/README.md?plain=1#L90

requests-for-comments/2026-04-23-better-sdk-for-llms.md
gewenyu99: Yeah it's a balance we have to strike. Try to make everything consistent where possible, write a skill for the rest

requests-for-comments/2026-04-23-better-sdk-for-llms.md
gewenyu99: If we can standardize and explicitly update examples in docs etc. to check for the nullish case, that'd be good.

Languages with looser typing will have less than obvious behavior when a third `undefined` type creeps in

requests-for-comments/2026-04-23-better-sdk-for-llms.md
gewenyu99: So I subtly am trying to push for y'all to publish and maintain skills in the SDK repos (so it's close to the source) for the quirks per SDK that you can't fix another way ;)

I know it's rough. But we _will_ unfortunately need these. What might be good is to build some agents that trigger on release to look at quirks like these. LLM models are good at finding inconsistencies that impact LLM performance.

gewenyu99: We can probably ~abuse~ use the Inkeep agents we already use for docs for this? I'm down to try this out.

marandaneto:
there's also this issue https://github.com/PostHog/posthog/issues/48234 with a bunch of differences across SDKs

requests-for-comments/2026-04-23-better-sdk-for-llms.md
marandaneto: ah gotcha, yeah, i used to call that a 'scoped' call

requests-for-comments/2026-04-23-better-sdk-for-llms.md
dustinbyrne: I like this idea

marandaneto:
thanks for the RFC @gewenyu99 
we've taken a few ideas from here and addressed some of them, and did a couple of other things
our goal was to improve the SDK usage by LLM but also not to disrupt SDK users with so many breaking changes
closing this for now, lets watch how LLMs trip with newer versions and take from there

PR #551, opened 2026-05-26
Make dashboard MCPs better

gewenyu99:
## Summary

The MCP shape of dashboards are not optimal. They're for hooking up UI, not agents.

## Checklist

### DIFF
--- a/requests-for-comments/2026-05-26-mcp-dashboard-specs.md
+++ b/requests-for-comments/2026-05-26-mcp-dashboard-specs.md
@@ -1,0 +1,486 @@
+# Request for comments: Agent friendly insight and dashboard MCP
+
+## Problem statement
+
+Agents are bad at building PostHog dashboards. For a dashboard with 5 insights, they need 12+ calls to the PostHog MCP. Run all 5 queries, turn them into insights, and lay them out.
+
+If an agent knows what insights it wants and what queries those insights should run, it should be able to one shot the entire dashboard in a single tool call. Today it cannot.
+
+The MCP is UI/API shaped. Not agent shaped.
+
+The proposal in this RFC, the MCP tool for dashboards and insights accepts a JSON bundle whose objects are the existing Terraform resource shapes (`posthog_dashboard`, `posthog_insight`, `posthog_dashboard_layout`). One document, one call, three resources written transactionally, server side. The agent never runs `terraform`. We do not invent a new spec format.
+
+## What the MCP looks like today
+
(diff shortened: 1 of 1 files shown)


requests-for-comments/2026-05-26-mcp-dashboard-specs.md
```diff
@@ -0,0 +1,486 @@
...
+    ]
+  }
+}
+```
+
+One tool call. Five insights, two markdown headers, seven tiles total, all laid out. Every field inside `dashboard`, each `insights[ref]`, and each `layout.tiles[*]` matches the existing Terraform resource. The only thing the MCP wrapper adds is the named refs (`signup_funnel`, `weekly_active_users`, etc.) and the bundled apply. The funnel uses the typed `kind: "Funnel"` shape. The four trend tiles use `kind: "HogQL"`.
+
+Compare that to the wizard session. The agent there wrote roughly the same number of bytes per tile, but distributed across 5 separate calls, with `kind: "InsightVizNode"` and `kind: "TrendsQuery"` and `kind: "EventsNode"` ceremony around every series, plus dashboard id threading on every call. No headers, because adding them would have been yet another tool call. The bundle collapses all of that into one document the agent writes once.
+
+If anything in the bundle is wrong, the agent gets the full list of problems back in one response, not five interleaved errors.
```
vdekrijger, line 486: I've been thinking about this problem during my trip home from Barbados as well; this was my initial direction / exploration: https://github.com/PostHog/posthog/pull/58781. _It's mostly AI code atm as I have been busy with other things, but it's on my list to pick up and dust off again._ 

There is some overlap (e.g. we need to create dashboards in 1 tool call vs the back and forth it has now), but I also wanted to focus on the up front experience by prompting users and gathering more details on what kind of dashboard they are looking for / have data for. My train of thought was that people:

1. Trust that whatever they have as source (e.g. data) is correct therefore only care about the results 
2. Can't be bothered to do a full exploration themselves, so they hand it off to / generally trust LLMs (there is too much complexity otherwise, in general)
3. Aren't always exactly sure what they are looking for, nor are people great at prompting what they want, so we need some interactivity during "complex" steps.

It's a pattern I noticed during the offsite, but also generally with people using LLMs (code / our AI / ...) and due to that the above PR came to live 😄 . 

Completely agree though that our current experience leaves room for improvement though!

gewenyu99: Yeah this makes sense. Keep in mind that:
-> The MCP tool should be deterministic, it does what you ask it to like an API
-> The skills, prompts, and agents is where you'd build this interactivity. You can of course, ship an mcp (`create-dashboards`) and another MCP tool that we can point to (`create-dashboards-skill`) that downloads a skill which helps you create the dashboard you want.

I'll trust you an Matt to figure out the optimal shape <3

PR #557, opened 2026-06-02
RFC: RBAC v2 - most-specific access wins

a-lider:
## Summary

Rethinks how the effective access level is resolved across resource, object, and default overrides. This RFC proposes a **most-specific-rule-wins** pattern, which also unblocks guest mode.

## Checklist


requests-for-comments/2026-05-30-rbac-v2-most-specific-access-wins.md
```diff
@@ -0,0 +1,164 @@
...
+- Update resource access control to use the "more specific wins" rule - member, then `max(roles)`, then default.
+- Allow the new access control UI to set Role and Member access for both projects and resources lower than the inherited level (default or role).
+- Update object access control to use the same "more specific wins" rule.
+- Add a new `shared_with` field for the default object access and migrate existing resources to `shared_with=project`.
+    - Add the new `shared_with` dropdown to object access settings in the side panel.
+- Make a specific object access apply when the user has no access to the project.
+- Make a specific resource permission apply when the user has no access to the project.
+
+The main risk is existing access control rules that are configured but don't apply and after the change they'll widen someone's access. 
+- Resource overrides will grant access regardless of a `No access` project.
```
yasen-posthog, line 137: Is there a migration plan for this case? Same as next point? What number of such cases do we have?

requests-for-comments/2026-05-30-rbac-v2-most-specific-access-wins.md
```diff
@@ -0,0 +1,164 @@
...
+
+But the API doesn't forbid setting a member's project access to `None` when the project default is `Member`. A project is treated as just another object, so it resolves the same way object access does (`resource=project`, `resource_id=project_id`).
+
+Recently we had a [ticket](https://posthoghelp.zendesk.com/agent/tickets/59200) where a customer had this issue.
+
+_Solution:_ allow selecting role/member project access lower than the default (inherited) level in the new UI.
+
+### Case 2 - object default vs role override
+
+Same as Case 1, but for an object - e.g. a dashboard where everyone has `Viewer` but one role has `No access`.
```
yasen-posthog, line 45: Got a flag on this that based on the current logic it may not be entirely true. Can you confirm?

requests-for-comments/2026-05-30-rbac-v2-most-specific-access-wins.md
```diff
@@ -0,0 +1,164 @@
...
+- Make a specific resource permission apply when the user has no access to the project.
+
+The main risk is existing access control rules that are configured but don't apply and after the change they'll widen someone's access. 
+- Resource overrides will grant access regardless of a `No access` project.
+- Object overrides that don't apply when project access is set to `None`. We could migrate existing defaults as shared with the project and add a warning on member/role overrides that this is an external member.
+Not many customers have these, so we can manually reach out before switching them to the new logic.
+
+Using more-specific access over additive rules is somewhat controversial, but I think it fits the PostHog model better, where an org has many projects and resources and users want to be able to both share and restrict a specific resource or object. 
+
+### Alternative solutions that I rejected
```
yasen-posthog, line 143: Have you thought about porting our existing data model to a proven AC decision framework like Zanzibar? I've been thinking about this for a while as an idea, because according to some basic research I did, we're pretty much there in terms of data model, we mainly need to pull in an OSS Zanzibar-like framework and create adapters for it in `UserAccessControl.access_level_for_object/resource`. However, because pure Zanzibar doesn't use `deny` action on purpose, we could use a flavor, such as SpiceDB or OpenFGA, or even CasBin if we want a lib-only approach.

These will help with:
- Cachable access graphs (Zanzibar is graph based)
- Familiar framework that agents can work with.
- A formal framework that has been proven at scale. Having a hand-rolled AC is great, but tbh my head almost exploded while reading the RFC and all the cases that we need to manually encode - I'd be much happier/worry-free if our implementation lives in a framework where all the edge cases are tested throughout and there are no bajillion stuff that can bite us in the ass.
- A framework/decision tree Enterprises are familiar with so there are no surprise behaviors.

Happy to hear what you think. If you think it's too heavy of a refactor or not worth it at all, lets discuss, I think the end state for AC should be in a standardized form, we should not be reinventing this specific wheel imo. Also, I'm not saying do it INSTEAD of the idea in the RFC, but maybe pick up as a next step?

yasen-posthog: Actually, the `shared_with` turns out is a very close alternative to something already existing in Zanzibar/ReBAC - parent->member relationships, but somehow limited to just 2 parents, while Zanzibar allows chains if we want proper AC hierarchy (eg you can split Replay into "replay" and "replay AI tools" and only allow access to AI tools if the user has access to the main replay obj, etc).

yasen-posthog (commented):
Overall, thorough RFC, thank you for putting it together! 

My main concern is if we're just trying to manually recreate something that already exists as a proven framework, in which case we need to step on the breaks and start thinking about a refactor into that robust foundation.

On the topic of Guest mode, I didn't fully understand how this change unlocks it, can you elaborate on that?

requests-for-comments/2026-05-30-rbac-v2-most-specific-access-wins.md
ReeceJones: tbh I think completely rebuilding our data model here and porting to an external framework is a bad idea, because:
1. We will give up a lot of control we have over the future of the access control product
2. We will spend a bunch of engineering time doing the migration, for very little customer benefit. From a ROI perspective, it is probably much better to fix our access control implementation than completely re-implement it.
3. Our data model is already mostly complete, it sounds like this RFC is mostly suggesting changes to the helpers that run on top of the data mdoel

requests-for-comments/2026-05-30-rbac-v2-most-specific-access-wins.md
a-lider: not a lot, have a note about this in case 6
> We'll need to notify customers who have projects with a `No access` default and resource overrides that will start applying. There are only 27 orgs and 40 teams affected (Metabase query: https://metabase.prod-us.posthog.dev/question/2114-rbac-projects-with-no-access-resource-defaults).

I think the right approach here would be to understand how existing customers use default resource overrides and what they expect. Then we can either remove the default overrides for them or explain how the change will affect them (grant resource access to org members without project access).

requests-for-comments/2026-05-30-rbac-v2-most-specific-access-wins.md
yasen-posthog: I disagree it's a bad idea. It may not be the right idea for now, but I think the direction should be towards a proven framework. As the focus shifts to getting more Enterprise customers, the edge-cases and hidden bugs will keep mounting and we'll end up with a spaghetti mess of a codebase. We need something that can scale, both implementation-wise and performance-wise (imagine orgs with millions of actors and resources).

Some examples that a graph framework will unlock:
- Dashboard->dashboard tile AC inheritance we now have custom logic for. Actually AC inheritance was one of the main blockers for Guest mode, cuz if you have N nodes in a Notebook, the AC needs to deeply propagate to all nodes, subnotes, subsubnodes, etc.
- Transitive/nested roles
- Visualizing the evaluation logic of a particular case (e.g for debugging, support, etc).

I don't think we should be as dismissive towards hardening one of our main package features. Optimally, for the future we need to be looking to limit any custom logic if the wheel's already invented for use at scale.

a-lider:
> On the topic of Guest mode, I didn't fully understand how this change unlocks it, can you elaborate on that?

cases 6 and 7 are basically guest mode - share just a specific resource or object without giving access to the whole project. The idea is that only the explicitly shared resources/objects resolve, and everything else errors out with "no access to project". So instead of a separate guest-mode path, a guest is just a member with `No access` to the project, plus explicit overrides.

a-lider:
Will try implementing it, resources that embed other resources (like notebooks or experiments) could be tricky

requests-for-comments/2026-05-30-rbac-v2-most-specific-access-wins.md
a-lider: I'll need to research this. Here's Google's paper on how Zanzibar works and how they got there: https://storage.googleapis.com/gweb-research2023-media/pubtools/5068.pdf

At first glance it won't solve the problems we have right now, it's additive by design, so deny-access based on a more specific rule won't work out of the box. And I worry it could turn into integration hell where we don't control the implementation and end up duct-taping edge cases around its limitations.

> Visualizing the evaluation logic of a particular case (e.g for debugging, support, etc).

Vizualization could be helpful, but it's just a way to explain how access applies based on the existing rules. I think the problem we have now is that the rules are confusing, and it's hard to reason about how access control is resolved for a user. 

> We will spend a bunch of engineering time doing the migration, for very little customer benefit. From a ROI perspective, it is probably much better to fix our access control implementation than completely re-implement it.

+++

> We need something that can scale, both implementation-wise and performance-wise (imagine orgs with millions of actors and resources).

Very long term, maybe, but we're definitely not at Google's scale yet, and the current solution can scale quite well. At some point, we might need to tweak the indexes or introduce caching.

requests-for-comments/2026-05-30-rbac-v2-most-specific-access-wins.md
a-lider: Double checked, yes, the more specific "None" for role wins over object default now. 

Image

PR #568, opened 2026-06-25
RFC: Browser SDK v1 -> v2 deprecations

dustinbyrne:
## Summary

Adds an RFC for the Browser SDK v2 public API cleanup, including:

- root API cleanup and camelCase migration
- feature flag API shape (`getFeatureFlag()` returns `FeatureFlagResult`)
- `disableRemoteFeatureFlags` naming
- removal of `advanced_disable_flags`, deprecated aliases, legacy globals, and preview flags
- defaults/versioning guidance
- migration catalog and implementation checklist

Decision maker: @Team Client Libraries

## Checklist

### DIFF
--- a/requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
+++ b/requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
@@ -49,7 +49,7 @@ if (result?.enabled) {
 - `posthog.getFeatureFlagPayload()` is removed.
 - `posthog.getFeatureFlagResult()` is not public; its semantics are folded into `getFeatureFlag()`.
 - `posthog.featureFlags` is internal and should not appear in public docs.
-- `posthog.featureFlags.override()` is removed from the public API. If client-side debug/test overrides remain supported, they need an explicit root-level public API.
+- `posthog.featureFlags.override()` is removed from the public API. The existing root `posthog.updateFlags(flags, payloads?, { merge })` API is the public way to inject or override flag values.
 
 ### Remote feature flags
 
@@ -138,7 +138,7 @@ These APIs are reachable in v1 but should not be public in v2.
 | v1 API | v2 API | Notes |
 | --- | --- | --- |
 | `posthog.featureFlags.getFeatureFlagPayload(key)` | No public child replacement | Use `posthog.getFeatureFlag(key)?.payload`. |
-| `posthog.featureFlags.override(flags, suppressWarning?)` | No public child replacement | Add an explicit root API only if debug/test overrides remain a supported use case. |
+| `posthog.featureFlags.override(flags, suppressWarning?)` | `posthog.updateFlags(flags, payloads?, { merge })` | Use the existing root API to inject or override flag values. |
 | `posthog.toolbar._loadEditor(params)` | `posthog.loadToolbar(params)` or `posthog.toolbar.loadToolbar(params)` | Toolbar is no longer called “editor” in the API. |
 | `posthog.toolbar.maybeLoadEditor(...)` | `posthog.toolbar.maybeLoadToolbar(...)` | Internal-ish but reachable. |
 | `posthog.sessionRecording.onRRwebEmit(rawEvent)` | No public replacement | Direct callers are relying on internals. |
@@ -399,6 +399,7 @@ Old queued calls such as `posthog.set_config(...)`, `posthog.register_once(...)`
 - [ ] Implement root `getFeatureFlag()` returning `FeatureFlagResult`.
 - [ ] Remove public root `getFeatureFlagPayload()` and `getFeatureFlagResult()`.
 - [ ] Make `posthog.featureFlags` internal and remove it from public docs/types.
+- [ ] Keep `posthog.updateFlags(flags, payloads?, { merge })` as the public root API for injecting or overriding flag values.
 - [ ] Rename `advanced_disable_feature_flags` to `disableRemoteFeatureFlags`.
 - [ ] Remove `advanced_disable_flags` and `advanced_disable_decide`.
 - [ ] Keep `defaults`; remove any synthetic initial `defaults: 'v2'` / `$config_defaults: 'v2'` behavior.
@@ -415,5 +416,5 @@ Old queued calls such as `posthog.set_config(...)`, `posthog.register_once(...)`
 - Known customer or wrapper SDK usage of removed aliases.
 - Any runtime aliases that are worth keeping for one compatibility release.
 - Whether nested config keys should be renamed in v2 or deferred.
-- Whether a public root-level feature flag override API is needed for debug/test workflows.
+- Any missing use cases that `posthog.updateFlags(flags, payloads?, { merge })` does not cover for debug/test flag overrides.
 - Any migration risks around changed default behavior, bot pageviews, `sanitize_properties` / `$set_once`, or split storage.
--- a/requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
+++ b/requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
@@ -3,7 +3,7 @@
 - **Status:** Proposed
 - **Decision maker:** @Team Client Libraries
 - **Owner:** @dustinbyrne
-- **Last updated:** 2026-06-25
+- **Last updated:** 2026-07-16
 
 ## Summary
 
@@ -11,78 +11,101 @@ Browser SDK v2 should be a clean public API boundary for the JavaScript browser 
 aliases, standardizes public JavaScript SDK naming on camelCase, makes feature flags return one consistent result
 object, and stops exposing internal implementation objects as public API.
 
-Event wire protocol and persisted identity/session semantics stay stable. This RFC makes the public API smaller, more
+Core event transport and persisted identity/session semantics stay stable. The explicitly listed `$browser_type` and
+`$config_defaults` property changes are the only event-property changes. This RFC makes the public API smaller, more
 consistent with newer SDKs, and easier to document and maintain.
 
 ## Goals
 
-- Keep the public SDK API focused on top-level supported methods.
-- Remove deprecated v1 aliases instead of carrying them into the next major version.
-- Standardize public JavaScript APIs and top-level config keys on camelCase instead of continuing the current mixed
+- Keep the public SDK API focused on app-facing methods.
+- Remove deprecated v1 aliases instead of carrying them into the next major version, except where ignoring an old
+  privacy or consent key could increase collection.
+- Standardize public JavaScript APIs and all SDK-owned option keys on camelCase instead of continuing the current mixed
   snake_case / camelCase surface.
 - Keep v2 feature flag reads on the root `posthog` object.
-- Keep internal objects (`posthog.featureFlags`, underscored `PostHog` internals, extension globals) out of the
-  documented public API.
-- Keep the `defaults` mechanism for future default-behavior versioning.
+- Keep internal objects (`posthog.featureFlags`, child implementation objects, underscored `PostHog` internals, and
+  legacy extension globals) out of both the documented public API and the public unmangled namespace; reserve only one
+  versioned private bridge for lazy extension bundles.
 
 ## Scope
 
-This RFC covers the browser SDK v2 public API and top-level config migration. Event names, event `$properties`, request
-body fields, query/body wire protocol fields, and persisted identity/session semantics stay stable. The observable
-behavior/config changes are listed below.
-
-Nested option-object keys stay unchanged in this RFC. Top-level config keys move to camelCase; nested keys remain as
-they are for now.
+This RFC covers the browser SDK v2 public API and config migration. All SDK-owned top-level and nested option keys move
+to camelCase from the first v2 release. Except for the explicit `$browser_type` addition and `$config_defaults` removal,
+event names, event `$properties`, request body fields, query/body wire protocol fields, third-party option objects passed
+through verbatim, and persisted identity/session semantics stay stable.
 
 Final npm package naming is tracked separately. This RFC applies to the browser v2 API whether it ships as `posthog-js`,
 `@posthog/browser`, or both during a transition.
 
+Product tours remains a shared extension that runs the same implementation in v1 and v2. Removing product tours code
+is not part of the v2 release and can be decided independently.
+
+Moving shared implementation into `@posthog/core` is also not a v2 prerequisite. We should evaluate it case by case
+against browser bundle size, the cost of maintaining v1, and the size of the shared API surface.
+
 Observable behavior/config changes covered by this RFC:
 
 - Drop `$bot_pageview` routing.
-- Keep `defaults` as the versioning mechanism for future default changes.
+- Adopt the latest staged v1 behaviors as the v2 baseline.
+- Remove the dated `defaults` mechanism; future behavior changes ship through explicit package and script-loader
+  versions.
 
 ## Decisions
 
 ### Feature flags
 
-v2 uses one root feature flag read API:
+v2 uses one synchronous root feature flag read API:
 
 ```ts
-const result = posthog.getFeatureFlag("new-onboarding");
+const result = posthog.getFeatureFlag('new-onboarding')
 if (result?.enabled) {
-  console.log(result.variant, result.payload);
+    console.log(result.variant, result.payload)
 }
 ```
 
-- `posthog.getFeatureFlag('name')` returns `FeatureFlagResult | undefined`.
-- `posthog.getFeatureFlagPayload()` is removed.
-- `posthog.getFeatureFlagResult()` is internal; its semantics are folded into `getFeatureFlag()`.
-- `posthog.featureFlags` is internal and excluded from public docs.
-- `posthog.featureFlags.override()` is removed from the public API. The existing root
-  `posthog.updateFlags(flags, payloads?, { merge })` API is the public way to inject or override flag values.
+- `posthog.getFeatureFlag('name')` returns `FeatureFlagResult | undefined`. It returns `undefined` when flags are not
+  available yet.
+- `posthog.getFeatureFlagPayload()`, `posthog.getFeatureFlagResult()`, `posthog.getAllFeatureFlags()`, and
+  `posthog.isFeatureEnabled()` are removed; their read semantics are folded into `getFeatureFlag()`.
+- `posthog.onFeatureFlags(callback)` remains the subscription API for initial availability and later updates, with the
+  callback signature `(flags: FeatureFlagResult[], errorsLoading: boolean) => void`.
+- `posthog.featureFlags` is private and absent from public docs, types, and the unmangled `posthog` namespace.
+- `posthog.featureFlags.override()` is removed. The root `posthog.updateFlags(flags, payloads?, { merge })` API is the
+  public way to inject or override flag values.
 
-### Remote feature flags config
+A `FeatureFlagResult` is an object even when `enabled` is `false`. Existing truthiness checks such as
+`if (posthog.getFeatureFlag('key'))` must migrate to `if (posthog.getFeatureFlag('key')?.enabled)`; otherwise a loaded
+disabled flag would take the truthy branch.
 
-- Rename browser `advanced_disable_feature_flags` to `disableRemoteFeatureFlags`.
-- Match the recent core / React Native / web-lite behavior: with `disableRemoteFeatureFlags: true`, reads use
-  bootstrapped or runtime-supplied flag values instead of remote flag evaluation.
-- Keep remote config available when `disableRemoteFeatureFlags` is enabled.
+### Remote feature flags
+
+- Replace browser `advanced_disable_feature_flags` with the positive `featureFlagEvaluation` boolean, which defaults to
+  `true`.
+- With `featureFlagEvaluation: false`, reads use bootstrapped or runtime-supplied values instead of remote flag
+  evaluation.
+- Remote config remains available when `featureFlagEvaluation` is disabled.
 - Drop `advanced_disable_flags` entirely. Remove the broader “kill remote config and flags APIs” switch.
 - Drop `advanced_disable_decide` with `advanced_disable_flags`.
 
-### Defaults
+### Default behavior
 
-- Keep `defaults` as the mechanism for future default-behavior versioning.
-- Users can omit `defaults` for the initial v2 baseline.
-- Future dated default sets continue to use `defaults: '<date>'`.
-- `$config_defaults` continues to reflect the configured `defaults` value.
+- Remove the `defaults` config key and `$config_defaults` event property in v2.
+- The first v2 release uses the baseline listed in [v2 baseline defaults](#v2-baseline-defaults).
+- Future default changes require an explicit package or script-loader version rather than a dated config switch.
 
 ### Async init and extension loading
 
-- v2 init is async.
+- v2 init is async and module callers await it before using the returned instance.
+- App-facing calls made through the snippet or root singleton while init is pending—including `capture()`, `identify()`,
+  registration, `setConfig()`, and consent methods—are queued in invocation order. Synchronous reads return their
+  documented not-ready value instead of being queued.
+- After base initialization, the SDK replays the full pre-init queue in order while network transmission remains paused.
+  It begins sending only after configuration, persistence, and consent operations in that queue have been applied, so a
+  queued opt-out can suppress earlier queued captures.
+- If init fails, queued calls are discarded and a warning is emitted; the SDK must not transmit a queued capture with
+  partially initialized privacy settings.
 - Extension initialization is deferred by default.
-- Remove `__preview_deferred_init_extensions` / `__previewDeferredInitExtensions` as a result.
+- Remove `__preview_deferred_init_extensions` / `__previewDeferredInitExtensions`.
 
 ### Bot pageviews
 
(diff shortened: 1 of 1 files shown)


requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+
+- Keep the public SDK API focused on top-level supported methods.
+- Remove deprecated v1 aliases instead of carrying them into the next major version.
+- Standardize public JavaScript APIs and top-level config keys on camelCase instead of continuing the current mixed
+  snake_case / camelCase surface.
+- Keep v2 feature flag reads on the root `posthog` object.
+- Keep internal objects (`posthog.featureFlags`, underscored `PostHog` internals, extension globals) out of the
+  documented public API.
+- Keep the `defaults` mechanism for future default-behavior versioning.
+
```
turnipdabeets, line 29: I'm curious what you think about moving things into core for this version? It seems like the place to put everything we want to share across SDKs but there are times where the JS SDK doesn't use core or there's a duplicating between core and the browser sdk.

dustinbyrne: We can look at de-duplicating on a case-by-case basis, but my top concerns would be:
- core doesn't necessarily care about byte sizes as much as browser
- we'll still be supporting posthog-js v1 for some time, so there are likely cases where it's better to keep the duplicated code shared with v1 instead of splitting
- the `PostHogCore` class also suffers from some of the things we're trying to remedy, like feature flags API surface bloat (`getFeatureFlagResult`, `getFeatureFlag`, `getFeatureFlagPayload`, `getFeatureFlagPayloads`, `getFeatureFlags`, `getFeatureFlagDetails`, `getFeatureFlagsAndPayloads`, `isFeatureEnabled`)

Even still, there's probably some smaller stuff that can move over. As long as we keep the public API surface small, we can also do this incrementally (especially as we phase out `posthog-js` v1).

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+## Goals
+
+- Keep the public SDK API focused on top-level supported methods.
+- Remove deprecated v1 aliases instead of carrying them into the next major version.
+- Standardize public JavaScript APIs and top-level config keys on camelCase instead of continuing the current mixed
+  snake_case / camelCase surface.
+- Keep v2 feature flag reads on the root `posthog` object.
+- Keep internal objects (`posthog.featureFlags`, underscored `PostHog` internals, extension globals) out of the
+  documented public API.
```
marandaneto: we need to check the official docs if we didnt leak this anywhere

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+- Keep the `defaults` mechanism for future default-behavior versioning.
```
marandaneto: mmm i see how this can be useful, but we are not blocked by not having major versions anymore with the script loader version approach, so not sure if we should keep it? that was the reason it was introduced at least afaik
@pauldambra would know more

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+
+This RFC covers the browser SDK v2 public API and top-level config migration. Event names, event `$properties`, request
+body fields, query/body wire protocol fields, and persisted identity/session semantics stay stable. The observable
+behavior/config changes are listed below.
+
+Nested option-object keys stay unchanged in this RFC. Top-level config keys move to camelCase; nested keys remain as
+they are for now.
+
+Final npm package naming is tracked separately. This RFC applies to the browser v2 API whether it ships as `posthog-js`,
+`@posthog/browser`, or both during a transition.
```
marandaneto, line 38: @posthog/browser imo

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+```ts
+const result = posthog.getFeatureFlag("new-onboarding");
+if (result?.enabled) {
+  console.log(result.variant, result.payload);
+}
+```
+
+- `posthog.getFeatureFlag('name')` returns `FeatureFlagResult | undefined`.
+- `posthog.getFeatureFlagPayload()` is removed.
+- `posthog.getFeatureFlagResult()` is internal; its semantics are folded into `getFeatureFlag()`.
```
marandaneto: i like that but we just deprecated `getFeatureFlagPayload` in favor of `getFeatureFlagResult`
now `getFeatureFlag` changes the return type and `getFeatureFlagResult` no more, might be confusing
whats about `isFeatureEnabled`?

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+- `posthog.featureFlags` is internal and excluded from public docs.
+- `posthog.featureFlags.override()` is removed from the public API. The existing root
+  `posthog.updateFlags(flags, payloads?, { merge })` API is the public way to inject or override flag values.
+
+### Remote feature flags config
+
+- Rename browser `advanced_disable_feature_flags` to `disableRemoteFeatureFlags`.
```
marandaneto: do we need the `remote`? and do we need the `disable`?
if its a clear boolean, i prefer flags named like eg `loadFeatureFlags = boolean = true`
no storng opinions since all flags are `disable` something, and false by default, so my idea goes against that, not sure if its worth the friction

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+
+Implementation changes:
+
+- Remove `__preview_capture_bot_pageviews`.
+- Remove `$bot_pageview` routing.
+
+## Why camelCase
+
+The browser SDK currently has a mixed public naming surface: snake_case methods and config keys sit next to newer
+camelCase APIs. v2 is the right boundary to standardize on idiomatic JavaScript naming across the public API.
```
marandaneto, line 133: +1

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+| -------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------- | ----------------------------------------------------------------------------- |
+| `posthog.people.set(props)` / `posthog.people.set('key', value)`           | `posthog.setPersonProperties(props)` / `posthog.setPersonProperties({ key: value })`                       | Callback argument is removed.                                    |
+| `posthog.people.set_once(props)` / `posthog.people.set_once('key', value)` | `posthog.setPersonProperties(undefined, props)` / `posthog.setPersonProperties(undefined, { key: value })` | Preserves `$set_once` behavior through the second argument.      |
+| `posthog.decideEndpointWasHit`                                             | `posthog.flagsEndpointWasHit`                                                                              | `/decide` naming is retired.                                     |
```
marandaneto: we also have a callback right, maybe we should change the signature of that callback as well

`onFeatureFlags([FeatureFlagResult] flags, errorsLoading)`

instead of `(featureFlags, featureFlagVariants, { errorsLoading } = {}`

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+| `posthog._calculate_event_properties(...)`                                 | `posthog.calculateEventProperties(...)`                                                                    | Deprecated since 1.241.0.                                        |
```
marandaneto: i think this should be a private API

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
marandaneto: maybe we name this method a bit different like `flagsLoaded` or something?

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+| `posthog.getFeatureFlagPayload(key)`                                       | `posthog.getFeatureFlag(key)?.payload`                                                                     | Payload is part of `FeatureFlagResult`.                          |
+| `posthog.getFeatureFlagResult(key)`                                        | `posthog.getFeatureFlag(key)`                                                                              | Result semantics move to the main API.                           |
+| `posthog.renderSurvey(surveyId, selector)`                                 | `posthog.displaySurvey(surveyId, { displayType: DisplaySurveyType.Inline, selector })`                     | `displaySurvey` also supports popovers and richer options.       |
+| `posthog.canRenderSurvey(surveyId)`                                        | `const result = await posthog.canRenderSurveyAsync(surveyId); result.visible`                              | Async API returns a `SurveyRenderReason`; read `result.visible`. |
```
marandaneto: should we drop the `async` since its clear now? `canRenderSurvey` becomes what `canRenderSurveyAsync` is today

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+
+## Child/internal API cleanup
+
+These APIs are reachable in v1 and become internal in v2.
+
+| v1 API                                                   | v2 API                                                                 | Notes                                                          |
+| -------------------------------------------------------- | ---------------------------------------------------------------------- | -------------------------------------------------------------- |
+| `posthog.featureFlags.getFeatureFlagPayload(key)`        | No public child replacement                                            | Use `posthog.getFeatureFlag(key)?.payload`.                    |
+| `posthog.featureFlags.override(flags, suppressWarning?)` | `posthog.updateFlags(flags, payloads?, { merge })`                     | Use the existing root API to inject or override flag values.   |
+| `posthog.toolbar._loadEditor(params)`                    | `posthog.loadToolbar(params)` or `posthog.toolbar.loadToolbar(params)` | Toolbar is no longer called “editor” in the API.               |
```
marandaneto: whats about making `posthog.toolbar` internal, so the only way is `posthog.loadToolbar`?

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+| `posthog.toolbar.maybeLoadEditor(...)`                   | `posthog.toolbar.maybeLoadToolbar(...)`                                | Internal-ish but reachable.                                    |
```
marandaneto: maybe the 'maybe` should be a param in `loadToolbar` so we kill `maybeLoadToolbar`

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+| `posthog.sessionRecording.onRRwebEmit(rawEvent)`         | No public replacement                                                  | Direct callers are relying on internals.                       |
```
marandaneto: do we need to expose this as a public api? whats the usage case? do we document this anywhere?

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+| `posthog.experiments._is_bot()`                          | `posthog.experiments.isBot()`                                          | Internal helper renamed during underscore cleanup.             |
```
marandaneto: same as posthog.experiments, should this be internal and expose `isBot` somewhere else?

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+| `set_initial_person_info()` | `setInitialPersonInfo()` |
+| `get_initial_props()`       | `getInitialProps()`      |
+| `update_config()`           | `updateConfig()`         |
+| `set_disabled()`            | `setDisabled()`          |
+| `set_cross_subdomain()`     | `setCrossSubdomain()`    |
+| `set_secure()`              | `setSecure()`            |
+| `set_event_timer()`         | `setEventTimer()`        |
+| `remove_event_timer()`      | `removeEventTimer()`     |
+| `get_property()`            | `getProperty()`          |
+| `set_property()`            | `setProperty()`          |
```
marandaneto: i think both should be internal and if there are use cases, we should expose a high level method in the posthog instance?

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+| `rate_limiting`                                | `rateLimiting`                           |
+| `fetch_options`                                | `fetchOptions`                           |
+| `request_queue_config`                         | `requestQueueConfig`                     |
+| `cookieless_mode`                              | `cookielessMode`                         |
+| `internal_or_test_user_hostname`               | `internalOrTestUserHostname`             |
+| `override_display_language`                    | `overrideDisplayLanguage`                |
+| `tracing_headers`                              | `tracingHeaders`                         |
+| `disable_beacon`                               | `disableBeacon`                          |
+
+Top-level keys that intentionally remain unchanged: `token`, `name`, `autocapture`, `rageclick`, `persistence`,
```
marandaneto: we can rename `token` to `projectToken`

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+| --------------------------------------- | --------- | ------------------------------------------------------------ |
+| `SurveyAppearance.descriptionTextColor` | Removed   | Deprecated and unused.                                       |
+| `SurveyAppearance.inputBackgroundColor` | Removed   | Use `inputBackground`.                                       |
+| `ActionStepType.tag_name`               | Removed   | Use `selector`.                                              |
+| `SurveyConfig.autoSubmitIfComplete`     | Removed   | No longer used; prefilled skip-submit behavior is automatic. |
+| `SurveyConfig.autoSubmitDelay`          | Removed   | No longer used.                                              |
+| `RemoteConfig.editorParams`             | Removed   | Use `toolbarParams`.                                         |
+| `RemoteConfig.toolbarVersion`           | Removed   | Obsolete toolbar metadata.                                   |
+
+## Public-ish underscored `PostHog` internals
```
marandaneto: we probably should not let this goes into the next major with the -ish, either its public or not

marandaneto:
nice work, left a few comments, just suggestions/ideas, most of them i dont have strong opinions but felt better while reading the rfc

marandaneto:
something that would be worth working on is making sure the sdk references work https://posthog.com/docs/references/posthog-js
for v1 and v2
we should generate a skill for the migration, and tell people install this skill and ask to migrate from v1 to v2

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+v2 uses one root feature flag read API:
+
+```ts
+const result = posthog.getFeatureFlag("new-onboarding");
+if (result?.enabled) {
+  console.log(result.variant, result.payload);
+}
+```
+
+- `posthog.getFeatureFlag('name')` returns `FeatureFlagResult | undefined`.
```
ioannisj: https://github.com/PostHog/requests-for-comments-public/pull/568/changes#r3480461991 but also this means that any existing checks like `if (posthog.getFeatureFlag('x'))` will always pass without any errors right?

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+### Defaults
+
+- Keep `defaults` as the mechanism for future default-behavior versioning.
+- Users can omit `defaults` for the initial v2 baseline.
+- Future dated default sets continue to use `defaults: '<date>'`.
+- `$config_defaults` continues to reflect the configured `defaults` value.
+
+### Async init and extension loading
+
+- v2 init is async.
```
ioannisj: What happens for to capture() while we are waiting for init? Do we have a pre-init queue that will capture those?

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+  documented public API.
+- Keep the `defaults` mechanism for future default-behavior versioning.
+
+## Scope
+
+This RFC covers the browser SDK v2 public API and top-level config migration. Event names, event `$properties`, request
+body fields, query/body wire protocol fields, and persisted identity/session semantics stay stable. The observable
+behavior/config changes are listed below.
+
+Nested option-object keys stay unchanged in this RFC. Top-level config keys move to camelCase; nested keys remain as
```
ioannisj: Is there a strong reason not to be consistent with cameCase with one go?

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
```diff
@@ -0,0 +1,447 @@
...
+posthog.getDistinctId();
+```
+
+Old queued calls such as `posthog.set_config(...)`, `posthog.register_once(...)`, or `posthog.get_distinct_id()` fail in
+v2 because the queue processor dispatches by exact method name.
+
+## Runtime behavior for removed aliases
+
+- Removed methods and properties are absent at runtime.
+- Removed config keys in plain JavaScript can still exist on user objects; the SDK ignores them.
```
ioannisj: Are all removed keys safe to ignore? consent, capture and privacy ones (if any) should warn at least for a few minor versions of v2 I think. Otherwise we may be capturing more than what's intended with nothing in the console to signal it?

marandaneto:
do we kill/remove all code from product tours? (prob not but worth checking)

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: this should probably read "out of the documented public API and unmangled namespace"

otherwise it's effectively public anyhow - not sure if it's documented, but i'd wager that some people are using it now even if it's not

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: hmm. maybe you're right. it seemed like it was still valuable in preventing additive changes from also turning on new, potentially unexpected behaviors.

worth dropping - we can always add it back in if need be

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: the core idea here is that we drop all of the flags APIs and just roll with `getFeatureFlag`. `FeatureFlagResult` is able to serve every previous flags API in a single return value

```ts
// OLD
if (posthog.isFeatureFlagEnabled('key')) {
  const payload = posthog.getFeatureFlagPayload('key')
  return { ...payload, classList: [ posthog.getFeatureFlag('key') ] }
}

// NEW
const flag = posthog.getFeatureFlag('key')
if (flag?.enabled) {
  return { ...flag.payload, classList: [ flag.variant ] }
}
```

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: `featureFlagEvaluation: true`?

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: agree - though, we might not need `onFeatureFlags` anymore if `getFeatureFlag` returns a promise - if they're not yet loaded, the promise would await the evaluation

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: agree

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: i'm actually not sure what this is useful for. i think we should drop it

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: yes

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: agree

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: `maybeLoadToolbar` shouldn't be public at the very least

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: to me it looks like it should be internal

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: yeah, looks to be internal. there's probably a better export for `isBot` if we need it

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: i think it might be best to keep it private then see if anyone asks for it

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: agree

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: 😂 

we should be private by default imo

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: that's right. calls like that would effectively work the same as they do now

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: no, you're right, we should be consistent from day one

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: yeah, this is a good point. maybe for some of these more sensitive properties we should just read both old and new keys

dustinbyrne:
product tours will be a shared extension (i.e., runs the same code in both v1 and v2) so we will be able to make that decision independently of the v2 release

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
marandaneto: yep i like the API, just not 100% sure if its not gonna be a confusing migration

requests-for-comments/2026-06-25-browser-sdk-v2-public-api.md
dustinbyrne: I suppose it depends on whether or not `capture` is sync or async. If it's sync, it can just buffer. If it's async, it can wait for the client to be ready if need be.

I do like the idea of keeping `capture` sync - it's just a quick write to a buffer. Something like `captureImmediate` for async waiting for an event to be flushed

arnohillen:
[arno's agent] The safety-critical allowlist covers config keys only: a stale snippet stub queueing posthog.opt_out_capturing() against a v2 loader gets silently dropped by exact-name dispatch, losing a consent signal and increasing collection, which contradicts the invariant in the safety-critical section.

arnohillen:
[arno's agent] opt_out_capturing_cookie_prefix matches the allowlist's every-opt_out_*-key rule but also sits in the removed-aliases table whose keys are ignored rather than translated, and since consentPersistenceName no longer appends the token it is ambiguous whether opt-out state under the old prefixed key is still read.

arnohillen:
<small>(created by claude code)</small>

[arno's agent] The safety allowlist omits enable_recording_console_log, custom_blocked_useragents, save_referrer, and save_campaign_params even though ignoring each increases collection: an ignored enable_recording_console_log: false falls back to the server-side console recording setting, ignored custom_blocked_useragents stops blocking those user agents, and ignored save_referrer/save_campaign_params: false resumes persisting referrer and campaign params.

arnohillen:
<small>(created by claude code)</small>

[arno's agent] The nested safety bullet covers only autocapture, exception, and dead-click keys, so a v1-spelled capturePerformance network_timing: false is warned-and-ignored and network performance capture reverts to its default, which increases collection contrary to this section's invariant.

arnohillen:
<small>(created by claude code)</small>

[arno's agent] The async-init queue covers only app-facing calls, and with extension initialization deferred by default the DOM-capture extensions (autocapture per posthog-js#4513, rageclick, dead clicks, heatmaps, replay) have no stated buffering or listener-installation semantics for interactions that happen during the deferral window.

arnohillen:
<small>(created by claude code)</small>

[arno's agent] The nested catalog misses rate_limiting.events_per_second/events_burst_limit and fetch_options.next_options, and keys added since 2026-07-16 (capture_dead_swipes family, sessionRecording.trigger_pending_buffer_interval_millis, top-level metrics and rewriteRequestPath) appear in no table, so the catalog needs a re-sync against main and a decision on the public posthog.metrics API.

arnohillen:
<small>(created by claude code)</small>

[arno's agent] $browser_type is already emitted in v1 on every event whenever opt_out_useragent_filter is true, so it is not a v2 addition; the only real event-taxonomy change here is dropping $bot_pageview.

arnohillen:
<small>(created by claude code)</small>

[arno's agent] advanced_disable_feature_flags_on_first_load survives as advancedDisableFeatureFlagsOnFirstLoad while its base key becomes featureFlagEvaluation, leaving a v2 option named after a key that no longer exists.

marandaneto:
close this issue once this is done https://github.com/PostHog/posthog-js/issues/155

PR #586, opened 2026-07-25
Wizard V2.5 and V3

gewenyu99:
## Summary

Speed unlocks many things. This 1:30 flow takes the old wizard anywhere from 5-10 minutes.

https://github.com/user-attachments/assets/48978bb4-27ef-466f-9fac-a760be6556fd 

## Checklist

- [ ] I have considered whether this RFC should instead be private, and confirmed it's appropriate for this public repo. (See the README — this repo is for non-product discussions that are safe to have in public. If the topic touches anything customer-sensitive, commercially sensitive, or otherwise non-public, it likely belongs in a private venue instead.)


requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,192 @@
...
+### 1. Self-driving first experience
+
+`@npx @posthog/wizard@latest` should do the following:
+- Install the SDK
+- Initialize it and hook up environment variables
+- Identify your users
+- Turn on all the features for self-driving
+- Ask you about external signal sources
+- Design and offer scouts to you
+- Kick you to PostHog Desktop or the Inbox.
```
rafaeelaudibert: The inbox for now

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,192 @@
...
+- Leave lots of useful context for _their_ agents to take over.
+
+For a self-driving users, this is what's needed to get started. We can assume they're AI pilled, they have their own agents, and they are comfortable with using the PostHog through the MCP.
```
rafaeelaudibert: Should we assume this from all users? I believe we actually should and we should be ok with not getting the customers who don't vibe with this approach

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,192 @@
...
+
+They don't need all the fluff on top. This _is_ just the thing we offer them.
+
+For **cloud runs**, we can surface all of these questions early and during the run, and just pass this information to the agent.
+
+This shape will change. Self-driving itself is fast changing. The new modular, multi-agent flows are flexible enough to adapt to these changes.
+
+### 2. Customizable, modular setup
+
+During onboarding, we already ask questions about which tools the user wants to use, what features they want to enable, and how they want to use the PostHog.
```
rafaeelaudibert: Well, the new self-driving onboarding built by @fercgomes doesn't. And I think we don't need to. We should, by default, instrument all of the base 4 products (PA, WA, ET, SR), and then only the others should be optionally suggested on top.

I agree we should do what you mention below after we assume the above should be enabled by default, though.

I also don't think some things make sense during the first connection: warehouse sources possibly not, for example. This is something we can sell to them eventually.

Unless, of course, we can keep their attention with us for the whole time we're installing PostHog in the background.

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,192 @@
...
+### 3. Option to defer to the user's agent
+We cannot really compete with Fable class models. That's fine. What if we offer that as an option? A basic installation.
+
+There are things we can do for you still:
+- Install
+- Environment variables
+- Enabling features
+- Hooking up identification (I'm convinced _we_ can do this uniquely lol)
+- Leaving skills for your agent
+- Leave meaningful instructions as handoff.
```
rafaeelaudibert: I'm afraid people will usually just pick this one all the time 😢

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,192 @@
...
+
+Have opinions? You mainly care about sign up conversion and time to checkout? We can do that for you, too.
+
+We'll finish with a GPT 5.6 Sol class audit. This, alone, turns 4/11 projects capturing successfully post instrumentation after a wizard run to 11/11. In our tests, this gives us 13% higher ingestion rate after a Wizard run in the same period.
+
+We can do this because GPT 5.6 Sol class models are faster, cheaper, and more capable than anything practical before.
+
+## The timeline
+
+I think we can and need to do this over the next quarter or so. All the prep work to make this Wizard redesign possible is already done.
```
rafaeelaudibert: What quarter is "next quarter"? This feels like something you'd start building now, correct?

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,192 @@
...
+
+### 4. Higher instrumentation quality
+ 
+What do you care about?
+
+Yes you, the user.
+
+Have no opinions? We'll scan your repo, hook up events for the core flows (and more than before) and set up insights and dashboards. Even leave suggestions for your agents.
+
+Have opinions? You mainly care about sign up conversion and time to checkout? We can do that for you, too.
```
rafaeelaudibert, line 109: I wonder if we should almost turn it into just something you connect to your Claude somehow? Or maybe we just build a Claude Code-like UI on top of it where it works in the background and its doing its thing but you can steer it somehow?

gewenyu99: Why are you here so early I'm stillll writing this

gewenyu99: Stopppppp read it again when I ping everyone I'm like half way through this

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: yes, my point exactly, the ICP has shifted. We're not the same company anymore.

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: But thank you and all the @PostHog/team-growth folks for giving me the ideas here and pushing me to take another look at what the Wizard does now and should do next.

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: This quarter. I think we can ship this literally now and I want to get on this now (assuming my team doesn't convince me otherwise xD)

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+The benchmark to beat in 2025 was an engineer reading our docs and trying to setup PostHog for Product Analytics themselves. The 2025 goal was simple: _get as many ingested, identified events into PostHog as possible_. 
+
+This is outdated in several ways:
+- Product Analytics is no longer the default assumed entry point
+- We're pushing to get people on to self-driving
+- The quality bar of an install has changed. Users (that are our ICP) are agent pilled and have Opus 5.
+
+The target has shifted. We can't keep shooting the same shot and hope it'll land.
+
+I'm weary of complacency. I'm weary that we're afraid to swing big because we have something that works and people use. But if we're repositioning entirely around self-driving and believe the ICP has shifted, maybe we can and should swing bigger.
```
sarahxsanders: a visionary if you will

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+| capture | 1:14 | 5:25 |
+| identify | 0:46 | 2:21 |
+| error-tracking | 0:22 | 1:29 |
+| build | 0:54 | 2:11 |
+| review | 1:55 | 4:49 |
+| dashboard | 0:27 | 1:23 |
+| report | 0:41 | 3:28 |
+| total (median run) | 7:58 (n=511) | 26:01 (n=85) |
+
+What would have taken us 26 minutes to run now takes less than 8 and costs ~$2.
```
sarahxsanders: 26 minutes to run: is this on a monorepo? what is product sized app here? just curious because don't most runs take 6-8 mins?

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+
+There are things we can do for you still:
+- Install
+- Environment variables
+- Enabling features
+- Hooking up identification (I'm convinced _we_ can do this uniquely lol)
+- Leaving skills for your agent
+- Leave meaningful instructions as handoff.
+
+### 4. Higher instrumentation quality
```
sarahxsanders, line 95: food for thought: it might be worth some deep research and due diligence on our part to understand how context engineering is changing with new models. the shape of what we've built for context to drive instrumentation quality may be changing. if we fail to stay on top of this, we could continue building context for a bygone era

example: https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+
+This isn't the case anymore. For self-driving, we need to:
+- Installed SDK across the repo (22% of the time monorepos)
+- Identification (some things never change) and group analytics
+- Enable all the features, like autocapture, session replays, error tracking, AIO, so on.
+- Connect data sources
+- Connect signal sources
+- Scouts
+- Bring you to PostHog to see the PRs flow
+- **Ask questions along the way to customize this setup**
```
sarahxsanders: I'm curious what _this_ part looks like, in the CLI AND in the cloud. how do you customize this setup in the cloud? 

also, what we need to be on top of security if we go anywhere near users being able to prompt the agent with text of their own (chatbot style)

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+
+If we are fully committing to self-driving, this is the flow we'd want to implement.
+
+Then there's cloud runs. This _is_ gonna be the fastest way to get to self-driving. Nothing beats a single-click setup that leads directly to a PR. This requires us to have that same autonomous experience, but with the questions to as up front.
+
+This default Wizard program isn't the right shape for a self-driving-first PostHog. 
+
+### The Wizard assumes users want Product Analytics (and isn't in a monorepo)
```
sarahxsanders: we have to be careful not to fall into the same trap with self-driving too though. especially if we launch v3 wizard with that in mind. what assumptions should we make, if any?

I think assuming it's not a monorepo is the easiest solve here

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+There's also practical niceties like:
+- A const file with all event definitions
+- Event schemas
+- Using PostHog through a warpper (often the preferred shape)
+- Group analytics
+- Reverse proxies
+
+Once again, these were great compared to an empty project and manual hookups. 
+
+These are significantly less useful if you look at what someone with Opus 5 or Fable can do with our docs.
```
sarahxsanders: what if we do an experiment with the most powerful models ourselves, and compare their integrations with just docs to the current state of the wizard, and outline gaps? this will give us the clearest vision for how to shape our context moving forward imo

right now it feels like a black box to me: "what _is_ Fable doing that the wizard isn't?" - time for some boots on the ground testing

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+
+Once again, these were great compared to an empty project and manual hookups. 
+
+These are significantly less useful if you look at what someone with Opus 5 or Fable can do with our docs. 
+
+## What makes a Wizard V3? (What we need to build next)
+
+If you're just curious about what we plan to ship next, skip directly to [what we'll be shipping](#what-we-will-be-shipping).
+
+1. Self-driving first experience
```
sarahxsanders: again, worried here that we're making the same assumptions with self-driving that we made with product analytics. how do we avoid this trap, while leading with self-driving?

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+2. Customizable, modular setup
+3. Option to defer to the user's agent
+4. Higher instrumentation quality
+
+### 1. Self-driving first experience
+
+`@npx @posthog/wizard@latest` should do the following:
```
sarahxsanders: what does the current self-driving wizard not do that's in this list? or is this list just what it must do, and we improve what it already does if it ticks something off this list?

sarahxsanders: aka - are we just migrating `self-driving` command to `wizard`

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+
+They don't need all the fluff on top. This _is_ just the thing we offer them.
+
+For **cloud runs**, we can surface all of these questions early and during the run, and just pass this information to the agent.
+
+This shape will change. Self-driving itself is fast changing. The new modular, multi-agent flows are flexible enough to adapt to these changes.
+
+### 2. Customizable, modular setup
+
+During onboarding, we already ask questions about which tools the user wants to use, what features they want to enable, and how they want to use the PostHog.
```
sarahxsanders: also, I think there are tools that make the user get the MOST out of self driving. we should enable these by default

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+- (Agent) Plans the flow and enqueues agents to instrument your projects (15s)
+- (Agent) We add and initialize the SDKs and hook up environment variables (30s - 1 min)
+    - At this point, you can just leave and you'll have a working setup to continue with your agent. You _can_ just leave.
+- (User) We pause after installing the SDKs, we ask you about features and teaches you about them.
+  - Do you want group analytics and what are the cohorts/groups you have? 
+  - What questions are you trying to answer with your product analytics integration? 
+  - Do you want to connect data warehouse sources? 
+  - Do you want self driving? (Do you know what it is???)
+  - Do you want some dashboards to start with?
+  - Do you want us to install PostHog Desktop?
```
sarahxsanders: these are all options/checkboxes? not actual "user prompting the agent back" interactions?

requests-for-comments/2026-07-24-wizard-v3.md
sarahxsanders: but free inference if they use us, they don't have to pay for their tokens :)

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+ 
+What do you care about?
+
+Yes you, the user.
+
+Have no opinions? We'll scan your repo, hook up events for the core flows (and more than before) and set up insights and dashboards. Even leave suggestions for your agents.
+
+Have opinions? You mainly care about sign up conversion and time to checkout? We can do that for you, too.
+
+We'll finish with a GPT 5.6 Sol class audit. This, alone, turns 4/11 projects capturing successfully post instrumentation after a wizard run to 11/11. In our tests, this gives us 13% higher ingestion rate after a Wizard run in the same period.
```
sarahxsanders, line 111: yesssssss, audit in the main integration run is GOATed

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+
+We can do this because GPT 5.6 Sol class models are faster, cheaper, and more capable than anything practical before.
+
+## Some mockups
+
+### Intro
+
+`npx @posthog/wizard` boots straight into the self-driving contract. No detection shown, no framework claim to be wrong about — the agent explores after you hit start, and nothing is edited until you confirm what it found (the exploration screen below is that plan). The second row is the agent-first escape hatch, covered in "Set up your agent" below. Auth is unchanged from today.
+
+```
```
sarahxsanders: would be happy to pair with you on this for the TUI mockups, I think if we're launching wizard v3, we need a best in class TUI design. and this is going to require being a bit more meticulous about getting UX right. we got hammered with a lot of UX papercuts with v2 because a lot of it was vibed and not well thought out. I'd like to avoid that with v3

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+  │    ↓ 28 more [N] for next page                                       │
+  │      Confirm                                                         │
+  ╰──────────────────────────────────────────────────────────────────────╯
+  Wizard todos   ✔ sdk  ✔ env  ✔ identify  ✔ features  ✔ github  ▶ sources  ◻ scouts
+
+ ────────────────────────────────────────────────────────────────────────────
+  ↑↓ navigate  enter toggle
+```
+
+And when the terminal is genuinely too small, say so — without stopping:
```
sarahxsanders: OOOOO nice

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+                    ⊂ |•o•| ⊃   Found a duplicate capture on the
+                                pricing page… whacking it.
+
+  ✔ sdk  ✔ env  ✔ identify  ✔ features  ✔ sources  ✔ scouts  ▶ audit  ◻ report
+
+ ──────────────────────────────────────────────────────────────────────────────────────────────
+  Status  Events  Logs  HN                            ←→ tabs  n nudge  q quit
+```
+
+The battle log cycles as the audit works — expressions track the fight (`|•o•|` shocked on a find, `|•ᴗ•|` cheeky after the kill):
```
sarahxsanders: this is so cuteee

sarahxsanders (commented):
overall, I think the sense of urgency here is warranted. a lot of this needs to be addressed this quarter and next. left some thoughts along the way, you can ignore the ones where I was questioning what things would look like >.< didn't realize you had mockups of UI

high level:
- not mentioned is how the shape of context engineering is changing with every new model. does v3 wizard also require a real deep dive into how well the current shape of the context mill is serving us/the robot too? 
- I'd really like us to put an emphasis on getting granular with the TUI this version as well. and promoting Windows (yes, I know) as a first-class citizen as well. I'd love a lot less TUI papercuts, and for us to instead launch v3 with a version of the TUI that allows us to grow WITH it. I worry if we don't plan for that, we are gonna constantly be on the hamster wheel of having to re-do it to fit our scope. also, component library would be 👍 I am very bullish on this, and equally interested in leading the efforts, per my quarterly goal hehe
- let's GET AFTER IT, this is good stuff

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+What does a perfect Wizard integration look like in Q3 2026?
+
+I don't know, but it's no longer the 2025 shape we built for.
+
+The benchmark to beat in 2025 was an engineer reading our docs and trying to setup PostHog for Product Analytics themselves. The 2025 goal was simple: _get as many ingested, identified events into PostHog as possible_. 
+
+This is outdated in several ways:
+- Product Analytics is no longer the default assumed entry point
+- We're pushing to get people on to self-driving
+- The quality bar of an install has changed. Users (that are our ICP) are agent pilled and have Opus 5.
```
edwinyjlim: true. what do we think quality means though – is it a wider or deeper or more customized integration the goal? probably all of the above

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+- Enable all the features, like autocapture, session replays, error tracking, AIO, so on.
+- Connect data sources
+- Connect signal sources
+- Scouts
+- Bring you to PostHog to see the PRs flow
+- **Ask questions along the way to customize this setup**
+
+If we are fully committing to self-driving, this is the flow we'd want to implement.
+
+Then there's cloud runs. This _is_ gonna be the fastest way to get to self-driving. Nothing beats a single-click setup that leads directly to a PR. This requires us to have that same autonomous experience, but with the questions to as up front.
```
edwinyjlim: agree that cloud is the end destination, but it's not clear to me what this means for us yet besides getting users to connect their GitHub repos and account via OAuth as soon as possible

requests-for-comments/2026-07-24-wizard-v3.md
edwinyjlim: totally

for self-driving, i'd think the assumption is to turn on _all_ tools and features to get as much signal as possible. or we could break things out into more precise use cases and scouts that only rely on certain tools

requests-for-comments/2026-07-24-wizard-v3.md
sarahxsanders: post huddle thoughts: I actually do think scouts/signals/self-driving becomes a byproduct if we can deliver them an artifact that shows them we SEE them and KNOW them. the true goal is to not deliver an instrumentation, it's to deliver the feeling of being known

we do this this way:
1. scan their codebase (already do this, sweet)
2. infer about their codebase - agent produces evidence-backed claims about their BUSINESS, _not events_ - orchestrator can do this easily. it already hands off per-file intent, evidence, and assumptions. the orchestrator delivers the evidence chain
3. MIRROR STAGE - this is the new interactive bit that gives you the feeling of using the [Mirror of Erised](https://harrypotter.fandom.com/wiki/Mirror_of_Erised). we present each claim that allows the user to confirm/correct/steer, etc. a lot of cool things with this: we get to catch wrong inferences and every correct _itself_ is a signals. those corrections give us more gold than we can get from rooting through their repo
4. deliverable - no more pile of events. we deliver events, dashboards, signals, scouts, etc. but in the context of what the user cares about: their business. we map the claims to PostHog tools that help them:
    - "your activation looks like publishing an article" turns into we think you care about "where does the publish flow silently fail? _to understand this concern, we will instrument error tracking and session replay for you on that form_
    - now wizard deliverables are mapped to actual business things users care about. they're happy to say "yeah I'll turn on that tool" because they now see without the tool, they can't get that data

requests-for-comments/2026-07-24-wizard-v3.md
sarahxsanders: important: we need to define what we currently don't like about the self-driving wizard program, then address the gaps

requests-for-comments/2026-07-24-wizard-v3.md
sarahxsanders: also to give @daniloc's localhost idea some credit - when I use Claude these days it LOVES to give me an HTML artifact. I think this is becoming a pattern humans like coming from agents. how do we mirror that with notebooks if we go that route for delivery?

daniloc: > deliver the feeling of being known

bingo

re: notebooks, I think the real virtue of them as the target here is they're persistent and shareable. You can get a teammate onboard later on if the notebook exists far better than a transient localhost artifact.

you're gonna have less expressiveness compared to building a new thing from scratch but I think the upside more than justifies it

sarahxsanders: Image

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+- Event schemas
+- Using PostHog through a warpper (often the preferred shape)
+- Group analytics
+- Reverse proxies
+
+Once again, these were great compared to an empty project and manual hookups. 
+
+These are significantly less useful if you look at what someone with Opus 5 or Fable can do with our docs. 
+
+## What makes a Wizard V3? (What we need to build next)
```
edwinyjlim: other than the self-driving output, i feel like all of these goals are direct outcomes of creating more granular, composable context to feed the wizard orchestrator

- UX and UI for custom input
- Custom context packages and tools for BYOA (bring your own agent)
- Better quality from better and more compact context

we can experiment with outputs all day once we have all the microskills we need

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+  - What questions are you trying to answer with your product analytics integration? 
+  - Do you want to connect data warehouse sources? 
+  - Do you want self driving? (Do you know what it is???)
+  - Do you want some dashboards to start with?
+  - Do you want us to install PostHog Desktop?
+- (Agent) We move on to instrument what you agreed to, probably as you say yes or no to our tools. (3 minutes)
+- (Agent) Then, once we're done, we ask to clean up after ourselves, review our changes, and leave context for your agent. This takes can take 5+ minutes. That's fine, leave us running, we'll handle the rest.
+
+### 3. Option to defer to the user's agent
+We cannot really compete with Fable class models. That's fine. What if we offer that as an option? A basic installation.
```
edwinyjlim: I think we're underestimating how much auth gives us an edge for a just-born, brand new user account

every little handoff after signup makes a difference even if it's just grabbing a PostHog project token or pre-populating a notebook 

but creating a custom context goodie bag for your agent would be cool

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+### 1. Because we can
+All the prep work to make this Wizard redesign possible is already done. 
+
+I have strong belief that the orchestrated DAG based approach with mini agents can work. I have strong belief that the models are cheap enough and getting cheaper. We can afford to experiment and iterate quickly.
+
+Also because, our ICP has shifted so much that we can afford to experiment. We can afford to drop some users on the floor and ship broken things.
+
+### 2. Because we need to
+
+Opus 5 is so cheap and so incredibly good at integrating PostHog. Give it the set of context we have and it will out perform the Wizard V2 in every way imagineable.
```
edwinyjlim: i don't agree with the framing of urgency here

of course cutting edge models could do the wizard's job if we provided them with the same context. the context is the hard part! 

frontier model convergence is going to be something every agent will have to deal with. the wizard will stay relevant as long as it delivers a delightful experience, which may not come from performance or speed

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+   - Initialize and apply environment variables
+   - Instrument basic identify
+   - instrument all of the base 4 products like Product Analytics, Web Analytics, Session Replay, AIO, and Error Tracking
+   - Enable all the features self-driving tries to enable.
+   - Write a handoff report for their agent locally, handoff to a handbook in PostHog
+    - Suggest captures, funnels, dashboards, insights, etc.
+    - Leave behind context for the agent to use the PostHog MCP
+    - Suggest data sources to connect
+    - Leave commands + links to skills to do things like add new data sources, upload source maps, etc.
+    - **Key removals**: 15 capture sites, dashboards, insights. This isn't relevant, leave the tools for their agents to drive these.
```
edwinyjlim: maybe i'm not getting the pessimism around these artifacts? do we think custom events with semantic names and insights aren't helpful PostHog outputs at all?

I get the speed advantage, but i feel like they're still important enough to generate at least 1-3 of them

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,872 @@
...
+
+Not doing a small set of place holder insights and events.
+
+We're at the verge of slow obscelescence.
+
+The targeted has shifted too far away from what we design the V2 Wizard to do. 
+
+The Wizard adapts, or we become obsolete.
+
+## What we're shipping
```
edwinyjlim: cool with pretty much all of these. they're either improving wizard programs or chaining them

requests-for-comments/2026-07-24-wizard-v3.md
daniloc: "here's a mirror" EXACTLY

requests-for-comments/2026-07-24-wizard-v3.md
charlescook-ph: from a website perspective, I would _love_ to send people to a cloud run as the main CTA, because 'copy some text into your terminal' while ok for now is not the best way to get folks to signup (I suspect)

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: Two things, quality as in meaningful, something they had review and input towards. Not wider, but more accurate to the products and scope they need (as in not giving them a tool they didn't ask for and not integrating the wrong part of the repo)

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: Compared to when I first tested orchestrator with Sonnet on a single repo. The overhead of the orchestrator itself was the big blocker

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: Chatbot is the wrong vision. It should be a highly scrutinized question that is asked and offered. Not really ever free form, multiple choice if you will. I think free form inputs are inherently bad, always, for a first time user. It's more about: Here's a set of things I think you should care about, select all that applies

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: I think two things:

* Initial PR should be minimal and extremely correct and reviewable
* We should surface (long term) follow up PRs incrementally. The Wizard becomes agent(s) that can be executed for smaller goals and tasks
* Once we're in their repo, it is much easier to ship iteratively
* Have justification (we're missing _these signals_ and thus we need this PR)

But note that we don't need all of this at once, we can focus on that first point of initial PR

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: I think there's two parts here:

* We decompose to mini skills, something the orchestrator can work with, and we make it trivial to reorganize what an agent does
* I think we want to turn on all the signals by default but we no longer need to give custom instrumentation, insights, and dashboards
* We signal maxxx and make that experience extremely linear

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: So auto capture vs. explicit tool instrumentation is the big difference here.

And the goal then becomes suggesting scouts AND the custom instrumentation required for that scout becomes much clearer

gewenyu99: Fable is actually running the app, clearer edge cases like CSP and getting around security fencing

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: Yes

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: I don't think the assumption is wrong. Holding on to the assumption and not challenging ourselves to rewrite the thing ASAP when things change _is_ the issue here

gewenyu99: I like this a lot.

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: The main gap is the integration and or integration amending. We're not adding an integration (minimal one) to unlock self-driving quickly, and we're not able to add additional instrumentation shaped to unlock the scouts

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: Yes, checkboxes, always, never freeform prompts. Those are evil

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: True, but I think this just changes our emphasis a bit.

We need to push quality forward and emphasize the unique, fun experience we want to create.

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: I think it's fine if they pick this? I don't see this distracting from our goal if we do a good job with context. Which we must

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: Really good intuition. I also think the context shape can be more and more like what we know are good docs. GPT 5.6 is really good at reasoning through limited ambiguity now.

Examples may be less and less necessary and more and more like shape anchors

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: Yes, these are aspirational and for the future. These are some ideas.

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: Maybe @edwinyjlim is right and we should use charm, who knows lol

requests-for-comments/2026-07-24-wizard-v3.md
gewenyu99: They're only useful if we suggested them and they want them. I think many of these are kinda bad. They should at least be suggested as a list for the user to pick.

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,113 @@
...
+  - What's required to answer these questions?
+    - Set up groups?
+    - Cohorts?
+    - How do we identify?
+    - What features need to be enabled?
+    - Where do we need explicit custom capture?
+- (Agent) We instrument what your questions require to answer — every capture exists because it feeds a question or scout you ticked (3 minutes)
+- (Agent) Then we clean up after ourselves, review our changes, and leave context for your agent. This can take 5+ minutes. That's fine, leave us running, we'll handle the rest.
+
+**Every one of these is a checkbox. Always. Never a freeform prompt — those are evil.** Chatbot is the wrong vision. A first time user doesn't know what to type at us, and a text box is an invitation to get it wrong. It should be a highly scrutinized question that we ask and offer: here's a set of things I think you should care about, select all that apply.
```
edwinyjlim, line 80: > **Every one of these is a checkbox. Always. Never a freeform prompt — those are evil.** 

amen

edwinyjlim (approved):
i'm sold on the vision. let's merge

requests-for-comments/2026-07-24-wizard-v2.5.md
```diff
@@ -0,0 +1,195 @@
...
+
+What we should ship now can be broken down into a few parts:
+
+### 1. Rebuild the context mill
+
+The context mill is shaped around the old agents we needed to hand hold. New models are excellent at understanding more freeform context when given a goal. New models can also be broken down into smaller agents and tasks that can be composed together to achieve a goal.
+
+What this means for the context mill:
+- Move "agent prompting" like defining goals back into the Wizard. We tight coupled what to do and how to do because the old models needed to be hand held.
+- Context mill becomes thin again. It should mostly reference docs and package it into  appropriate little bundles for distinct tasks. This is with some limited information like "examples" for agents, commandments (pitfalls for agents not humans).
```
sarahxsanders, line 116: YESSSSSSSSS. thin context mill is needed. it's become a bit of an overwhelm to root around in there

requests-for-comments/2026-07-24-wizard-v2.5.md
```diff
@@ -0,0 +1,195 @@
...
+- Ship docs updates to fit the context mill's needs, not use the context mill to monkey patch bad docs. Those docs are also mostly read by agents anyway.
```
sarahxsanders, line 117: might be worth discussing what the future of the /docs become and how we can better format them to serve as context delivery - if pocket guides become better "docs" for human readers, I think we can re-envision our docs for systems

requests-for-comments/2026-07-24-wizard-v2.5.md
```diff
@@ -0,0 +1,195 @@
...
+- The Wizard-CI MCP server already does this. PostHog Code can drive the agent with it to take snapshots and test on my behalf.
+
+What this looks like:
+
+![Agent operates on the store, the task stream layer syncs it, any UI renders it](../images/2026-07-24-wizard-v3/agent-tui-split.png)
+
+
+### 4. Force MCP install on the outro and cut out the rest
+
+It's Q3 2026 and we're self-driving. Of course they need the MCP. I'm convinced we should ask about Slack and show the tutorial elsewhere.
```
sarahxsanders, line 163: fan of this

requests-for-comments/2026-07-24-wizard-v3.md
sarahxsanders: evilllllll

requests-for-comments/2026-07-24-wizard-v3.md
```diff
@@ -0,0 +1,113 @@
...
+
+That also means we never have a user's free text steering an agent that has write access to their repo, which is a security property I'd like to keep for free.
```
sarahxsanders, line 82: bullish on this no matter how many times people ask for it

sarahxsanders (approved):
this is muchhhhhh better, glad you took some time to rescope. full speed ahead on 2.5

PR #594, opened 2026-08-13
RFC: Adopt relationship-based access control

ReeceJones:
## Summary

_What is this PR proposing or changing?_

## Checklist

### DIFF
--- a/requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
+++ b/requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
@@ -84,7 +84,7 @@ For example, a viewset which returns `Dashboards` would be a PEP. The viewset wo
 
 This decoupled architecture improves scalability, because changes to the access control rules are contained to the PDP. The PEP treats the PDP as a black box, which reduces the blast radius of access control logic changes. Additionally, modern authorization systems rely on domain specific languages (DSLs) and specialized data structures to model access control rules. This imposes a design constraint that forces developers to model access control as concise formal logic rather than application code.
 
-See [ABAC](https://en.wikipedia.org/wiki/Attribute-based_access_control) and [Cedar](https://docs.cedarpolicy.com/) for further reading on this architecture.
+See [ABAC](https://en.wikipedia.org/wiki/Attribute-based_authorization) and [Cedar](https://docs.cedarpolicy.com/) for further reading on this architecture.
 
 ### ReBAC
 
@@ -108,7 +108,7 @@ For example, a user can edit a dashboard if they are a project admin, or they ar
 
 - *Centralized logic* - All logic is colocated in one place, making it easier to maintain.
 - *Audit logs* - With all authorization requests routed through a single interface, we can log authorization decisions for security review.
-- *Simplified API* - Clients are less aware of the access control implementation and can be thinner, which will help with maintainability.
+- *Simplified SDK* - Clients are less aware of the access control implementation and can be thinner, which will help with maintainability.
 - *Formalized rules* - Rules are separated from code, and written down to make it easier to reason about, develop, and maintain.
 - *More authorization surfaces* - Access control can be used for more use cases such as tenant isolation, plan entitlements, sharing, contextual policies.
 - *Own the rules, not the engine* - We own the rules that determine authorization, not the infrastructure code that iterates over DB rows and figures out how to turn that into an authorization decision.
@@ -122,186 +122,186 @@ For example, a user can edit a dashboard if they are a project admin, or they ar
 ## Design 
 *What are the key user experience and technical design decisions / trade-offs?*
 
-We will adopt a ReBAC framework, then expose a minimal Python API for clients to perform authorization checks. In this case the python API will act as the PDP, and clients will act as the PEP.
+We will adopt a ReBAC framework, then expose a minimal Python SDK for clients to perform authorization checks. In this case the python SDK will act as the PDP, and clients will act as the PEP.
 
-### Python API
+### Python SDK
 
-I propose the following approximate python API. This python API should be exposed from the `products/access_control` product folder, and be the only way that access control checks are performed by application code outside of `products/access_control`.
+I propose the following approximate Python SDK.
 
 ```python
-from products.access_control.backend.facade import access_control
+from posthog import authorization
 
 # Check if a user can perform an action on a resource
-can_access = access_control.check(subject=..., resource=..., action=...)
+can_access = authorization.check(subject=..., resource=..., action=...)
 
-# Same as `access_control.check(...)`, but performs multiple checks in bulk
-can_access = access_control.check(
+# Same as `authorization.check(...)`, but performs multiple checks in bulk
+can_access = authorization.check(
     [
-        access_control.CheckRequest(subject=..., resource=..., action=...),
-        access_control.CheckRequest(subject=..., resource=..., action=...),
+        authorization.CheckRequest(subject=..., resource=..., action=...),
+        authorization.CheckRequest(subject=..., resource=..., action=...),
     ]
 )
 
 # Enumerate a user's permissions for a resource
-permissions = access_control.permissions(subject=..., resource=...)
+permissions = authorization.permissions(subject=..., resource=...)
 
 # Enumerate resources a user can access
-accessible_resources = access_control.list_resources(subject=..., resource=..., action=...)
+accessible_resources = authorization.list_resources(subject=..., resource=..., action=...)
 
 # Enumerate users who can access a resource
-authorized_users = access_control.list_subjects(resource=..., action=..., subject=...)
+authorized_users = authorization.list_subjects(resource=..., action=..., subject=...)
 
 # Generic utility for upserting and deleting relationships in bulk
-access_control.write(
+authorization.write(
     [
-        access_control.AddRelationshipRequest(subject=..., resource=..., action=...),
-        access_control.DeleteRelationshipRequest(subject=..., resource=..., action=...),
+        authorization.AddRelationshipRequest(subject=..., resource=..., action=...),
+        authorization.DeleteRelationshipRequest(subject=..., resource=..., action=...),
     ]
 )
 
 # Special function to inspect the graph. Returns the existence of the (subject, resource, action) tuple in the graph, NOT whether the user has access
 # Should only be used by APIs that explicitly need to check for specific relations like sharing UIs, for example.
 # Prefixed with `dangerous_` to avoid accidental use
-access_control.dangerous_query_relation(subject=..., resource=..., action=...)
+authorization.dangerous_query_relation(subject=..., resource=..., action=...)
 
 # Debug function for PostHog admins. If decision is to allow, returns context for what relations were involved in that decision being made.
-access_control.dangerous_explain(access_control.CheckRequest())
+authorization.dangerous_explain(authorization.CheckRequest())
 ```
 
-**For example, here are what some common access control operations might look like using this API.**
+**For example, here are what some common access control operations might look like using this SDK.**
 
 ```python
-from products.access_control.backend.facade import access_control
+from posthog import authorization
 
 # check if user can access dashboard
-can_access = access_control.check(
-    subject=access_control.Subject(user_id=user.pk), # ex: "user:{user.pk}"
-    resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboard.pk), # ex: "dashboard:{dashboard.pk}"
-    action=access_control.Action.Edit,
+can_access = authorization.check(
+    subject=authorization.Subject(user_id=user.pk), # ex: "user:{user.pk}"
+    resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboard.pk), # ex: "dashboard:{dashboard.pk}"
+    action=authorization.Action.Edit,
 )
 assert can_access is False
 
 # grant user read and write access to dashboard
-access_control.write(
+authorization.write(
     [
-        access_control.AddRelationshipRequest(
-            subject=access_control.Subject(user_id=user.pk),
-            resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboard.pk),
-            action=access_control.Action.View,
+        authorization.AddRelationshipRequest(
+            subject=authorization.Subject(user_id=user.pk),
+            resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboard.pk),
+            action=authorization.Action.View,
         ),
-        access_control.AddRelationshipRequest(
-            subject=access_control.Subject(user_id=user.pk),
-            resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboard.pk),
-            action=access_control.Action.Edit,
+        authorization.AddRelationshipRequest(
+            subject=authorization.Subject(user_id=user.pk),
+            resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboard.pk),
+            action=authorization.Action.Edit,
         ),
     ]
 )
 
 # check if user can access dashboard
-can_access = access_control.check(
-    subject=access_control.Subject(user_id=user.pk),
-    resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboard.pk),
-    action=access_control.Action.Edit,
+can_access = authorization.check(
+    subject=authorization.Subject(user_id=user.pk),
+    resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboard.pk),
+    action=authorization.Action.Edit,
 )
 assert can_access is True
 
 # remove user's edit access from the dashboard
-access_control.write(
-    access_control.DeleteRelationshipRequest(
-        subject=access_control.Subject(user_id=user.pk),
-        resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboard.pk),
-        action=access_control.Action.Edit,
+authorization.write(
+    authorization.DeleteRelationshipRequest(
+        subject=authorization.Subject(user_id=user.pk),
+        resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboard.pk),
+        action=authorization.Action.Edit,
     ),
 )
 
 # let user's list all dashboards in their project
-access_control.write(
+authorization.write(
     [
-        access_control.AddRelationshipRequest(
-            subject=access_control.Subject(user_id=user.pk),
-            resource=access_control.Resource(type=access_control.ResourceType.Dashboard, project_id=project.pk), # ex: "project:{project.pk}:dashboard"
-            action=access_control.Action.List,
+        authorization.AddRelationshipRequest(
+            subject=authorization.Subject(user_id=user.pk),
+            resource=authorization.Resource(type=authorization.ResourceType.Dashboard, project_id=project.pk), # ex: "project:{project.pk}:dashboard"
+            action=authorization.Action.List,
         ),
     ]
 )
 
 # check if a user can list dashboards
-can_access = access_control.check(
-    subject=access_control.Subject(user_id=user.pk),
-    resource=access_control.Resource(type=access_control.ResourceType.Dashboard, project_id=project.pk),
-    action=access_control.Action.List,
+can_access = authorization.check(
+    subject=authorization.Subject(user_id=user.pk),
+    resource=authorization.Resource(type=authorization.ResourceType.Dashboard, project_id=project.pk),
+    action=authorization.Action.List,
 )
 assert can_access is True
 
 # check if a user can create dashboards
-can_access = access_control.check(
-    subject=access_control.Subject(user_id=user.pk),
-    resource=access_control.Resource(type=access_control.ResourceType.Dashboard, project_id=project.pk),
-    action=access_control.Action.Create,
+can_access = authorization.check(
+    subject=authorization.Subject(user_id=user.pk),
+    resource=authorization.Resource(type=authorization.ResourceType.Dashboard, project_id=project.pk),
+    action=authorization.Action.Create,
 )
 assert can_access is False
 
 # check if a user can view a dashboard that they have not been explicitly given access to
-can_access = access_control.check(
-    subject=access_control.Subject(user_id=user.pk),
-    resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=other_dashboard.pk),
-    action=access_control.Action.View,
+can_access = authorization.check(
+    subject=authorization.Subject(user_id=user.pk),
+    resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=other_dashboard.pk),
+    action=authorization.Action.View,
 )
 assert can_access is True
 
 # check if a user can access a bunch of different dashboards
 dashboards = [dashboard, other_dashboard_1, other_dashboard_2]
-can_access = access_control.check(
+can_access = authorization.check(
     [
-        access_control.CheckRequest(
-            subject=access_control.Subject(user_id=user.pk),
-            resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboards[0].pk),
-            action=access_control.Action.View,
+        authorization.CheckRequest(
+            subject=authorization.Subject(user_id=user.pk),
+            resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboards[0].pk),
+            action=authorization.Action.View,
         ),
-        access_control.CheckRequest(
-            subject=access_control.Subject(user_id=user.pk),
-            resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboards[1].pk),
-            action=access_control.Action.View,
+        authorization.CheckRequest(
+            subject=authorization.Subject(user_id=user.pk),
+            resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboards[1].pk),
+            action=authorization.Action.View,
         ),
-        access_control.CheckRequest(
-            subject=access_control.Subject(user_id=user.pk),
-            resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboards[2].pk),
-            action=access_control.Action.View,
+        authorization.CheckRequest(
+            subject=authorization.Subject(user_id=user.pk),
+            resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboards[2].pk),
+            action=authorization.Action.View,
         ),
     ]
 )
 assert can_access == [True, False, False]
 
 # see what a user can do for a particular dashboard
-permissions = access_control.permissions(
-    subject=access_control.Subject(user_id=user.pk),
-    resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboard.pk),
+permissions = authorization.permissions(
+    subject=authorization.Subject(user_id=user.pk),
+    resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboard.pk),
 )
-assert permissions == [access_control.Action.View]
+assert permissions == [authorization.Action.View]
 
 # see what dashboards a user can access
-resources = access_control.list_resources(
-    subject=access_control.Subject(user_id=user.pk),
-    resource=access_control.Resource(type=access_control.ResourceType.Dashboard, project_id=project.pk),
-    action=access_control.Action.View,
+resources = authorization.list_resources(
+    subject=authorization.Subject(user_id=user.pk),
+    resource=authorization.Resource(type=authorization.ResourceType.Dashboard, project_id=project.pk),
+    action=authorization.Action.View,
 )
-assert resources == [access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboard.pk)]
+assert resources == [authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboard.pk)]
 
 # see who can access a dashboard
-subjects = access_control.list_subjects(
-    subject=access_control.Subject(type=access_control.SubjectType.User),
-    resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboard.pk),
-    action=access_control.Action.View,
+subjects = authorization.list_subjects(
+    subject=authorization.Subject(type=authorization.SubjectType.User),
+    resource=authorization.Resource(type=authorization.ResourceType.Dashboard, id=dashboard.pk),
+    action=authorization.Action.View,
 )
-assert subjects == [access_control.Subject(user_id=user.pk)]
+assert subjects == [authorization.Subject(user_id=user.pk)]
 
 # Bring together role-based access control, data warehouse table access control, and relationship based access control into one beautiful access control graph
-access_control.write(
+authorization.write(
     [
-        access_control.AddRelationshipRequest(
-            subject=access_control.Subject(role=role, project_id=project.pk), # ex: "project:{project.pk}:role:{role.pk}"
+        authorization.AddRelationshipRequest(
+            subject=authorization.Subject(role=role, project_id=project.pk), # ex: "project:{project.pk}:role:{role.pk}"
             resource=DataWarehouseTableResource(project_id=project.pk, table="postgres.stripe_invoices"), # ex: "project:{project.pk}:data_warehouse_table:postgres.stripe_invoices"
-            action=access_control.Action.View,
+            action=authorization.Action.View,
         ),
     ]
 )
@@ -309,9 +309,9 @@ access_control.write(
 
 ### Gotcha: Django-isms
 
-The mega-brained reader will know that there is a lot of access control checks and enforcement which happens in Django-land. With the new Python API we will re-implement the many Django permission enforcement and utility classes to use the new Python API under the hood, instead of directly evaluating access control rules from the database.
+The mega-brained reader will know that there is a lot of access control checks and enforcement which happens in Django-land. With the new Python SDK we will re-implement the many Django permission enforcement and utility classes to use the new Python SDK under the hood, instead of directly evaluating access control rules from the database.
 
-For example, `AccessControlViewSetMixin` would need to be updated to use the new python API instead of implementing the access control checks directly.
+For example, `AccessControlViewSetMixin` would need to be updated to use the new python SDK instead of implementing the access control checks directly.
 
 ### Gotcha: list operations
 List operations introduce an additional challenge because our existing access control system enforces object-level restrictions directly in the queryset. This means unauthorized objects are omitted from list results entirely. We will need to be careful to maintain this pre-existing behavior while migrating. Some queries are easy to migrate, only filtering objects that match a list of specific ids, whereas others are more complicated, joining on access control tables.
@@ -326,7 +326,7 @@ As we've scaled up the access coverage over more products, the data warehouse, p
 
 We already have lots of hierarchy in our rules: project admin overrides, creator escape hatch, object-level overrides, roles, etc. that are all dependent on their relationships between each other. ABAC is better suited for flat access control structures that are dependent on inline object attributes. For us, ReBAC is a better fit since it directly models relations.
 
-**Move existing code behind new Python API, but don't adopt ReBAC**
+**Move existing code behind new Python SDK, but don't adopt ReBAC**
 
 In this case, we are still spending a lot of time maintaining the code which evaluates the access control rules. Each new use case requires lots of new Python code that intermixes different concerns and is dificult to refactor.
 
@@ -336,11 +336,13 @@ If we adopt a standard ReBAC engine, then code examples on the internet makes it
 we want to spend the bulk of our development effort on the authorization rules, not the plumbing that enables that.
 
 ### Rollout?
-The python API will control rollout of ReBAC backed access control. Initially, we will port the existing access control code into the access control product folder, and expose it through the new python API. Then, once we start implementing it to be backed by ReBAC, we will set up a flag that determines whether the legacy access control code or the new ReBAC model should be used for authorization evaluation.
+The python SDK will control rollout of ReBAC backed access control. Initially, we will port the existing access control code into the access control product folder, and expose it through the new python SDK. Then, once we start implementing it to be backed by ReBAC, we will set up a flag that determines whether the legacy access control code or the new ReBAC model should be used for authorization evaluation.
+
+This flag will have multiple variants: `off`, `shadow-log`, `rebac`. In `off`, only the existing Python logic is used. In `shadow-log` the existing Python is used as the authoritative answer, but a request is made to the ReBAC service in parallel and that result is logged against the result returned by the Python implementation. In `rebac`, the ReBAC service's response is used as the authoritative answer, and the Python implementation's response is logged.
 
 ### What about the data model?
 
-Until we have a reason to replace the existing data model, we will avoid doing so. In any implementation, we need a source of truth for access control policies. All that changes is these policies are replicated into the ReBAC engine.
+In the first iteration we will maintain the existing data model by implementing a dual write system for object permissions. However, once the migration is complete we may re-evaluate this and opt to use the ReBAC service as the canonical source of truth for authorization policies, including those that we display in the UI and allow customers to configure.
 
 ### Most specific access control?
 
@@ -350,10 +352,14 @@ We will block this initiative on most-specific access control being shipped. Mai
 
 No, I think we should avoid this, and remain cloud-native since this is core infra.
 
+### Who would own this?
+
+Platform features would be responsible for maintaining, monitoring, scaling, and on-call response.
+
 ## Sprints
 
 1. Research ReBAC frameworks, benchmark them, and select final option
-2. Implement python control API using existing access control code
+2. Implement python authorization SDK using existing access control code
 3. Implement access control decision logging
 4. Set up ReBAC infra in prod, update local dev stack, update self-hosted stuff
 5. Implement ReBAC backed access control behind a feature flag, add ReBAC decisions to access control logging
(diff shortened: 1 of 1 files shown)


requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+- *ReBAC graph could become inconsistent with DB state* - We need to make sure CRUD operations to DB objects are synchronized with the ReBAC graph, and we need to develop tools to respond to potential incidents where ReBAC graph state drifts and becomes inconsistent with the DB.
+
+## Design 
+*What are the key user experience and technical design decisions / trade-offs?*
+
+We will adopt a ReBAC framework, then expose a minimal Python API for clients to perform authorization checks. In this case the python API will act as the PDP, and clients will act as the PEP.
+
+### Python API
+
+I propose the following approximate python API. This python API should be exposed from the `products/access_control` product folder, and be the only way that access control checks are performed by application code outside of `products/access_control`.
```
rafaeelaudibert: This should not be something exported from `products/` IMO. This is something higher-level that should be coming from the `posthog` folder, or some other higher-level folder.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+# Special function to inspect the graph. Returns the existence of the (subject, resource, action) tuple in the graph, NOT whether the user has access
+# Should only be used by APIs that explicitly need to check for specific relations like sharing UIs, for example.
+# Prefixed with `dangerous_` to avoid accidental use
+access_control.dangerous_query_relation(subject=..., resource=..., action=...)
+
+# Debug function for PostHog admins. If decision is to allow, returns context for what relations were involved in that decision being made.
+access_control.dangerous_explain(access_control.CheckRequest())
+```
+
+**For example, here are what some common access control operations might look like using this API.**
```
rafaeelaudibert: I know you reference this below, but we do need to make sure we make this feel Django-native, allowing it to easily interact with DRF via middlewares and decorators

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+- *Simplified API* - Clients are less aware of the access control implementation and can be thinner, which will help with maintainability.
+- *Formalized rules* - Rules are separated from code, and written down to make it easier to reason about, develop, and maintain.
+- *More authorization surfaces* - Access control can be used for more use cases such as tenant isolation, plan entitlements, sharing, contextual policies.
+- *Own the rules, not the engine* - We own the rules that determine authorization, not the infrastructure code that iterates over DB rows and figures out how to turn that into an authorization decision.
+
+**Cons**
+
+- *More complex architecture* - Requires an additional service deployed to the cluster, with its own data store.
+- *Potential for higher latency* - It is possible that there is more network or computational overhead for each authorization request.
+- *ReBAC graph could become inconsistent with DB state* - We need to make sure CRUD operations to DB objects are synchronized with the ReBAC graph, and we need to develop tools to respond to potential incidents where ReBAC graph state drifts and becomes inconsistent with the DB.
```
benjackwhite, line 122: This is the one i'm most worried about with a distributed system. We would need to somehow have this service deeply baked into all of our test suites as we won't have any standard protection

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
benjackwhite: What does the python API offer? Isn't the whole point of a service is that the service is just called directly? So we would just have some sort of shared helper code for calling it, not another service that could itself be slow/ a bottleneck / an extra unnecessary jump?

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+        ),
+    ]
+)
+```
+
+### Gotcha: Django-isms
+
+The mega-brained reader will know that there is a lot of access control checks and enforcement which happens in Django-land. With the new Python API we will re-implement the many Django permission enforcement and utility classes to use the new Python API under the hood, instead of directly evaluating access control rules from the database.
+
+For example, `AccessControlViewSetMixin` would need to be updated to use the new python API instead of implementing the access control checks directly.
```
benjackwhite: It will _all_ have to do _both_ for quite a while... having built the original access control stuff I can tell you what a headache that was... 

But we should plan for this being a slow rollout which will take a long time and figure out how to run both in parallel to be able to gain confidence in the new system

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+### Gotcha: list operations
+List operations introduce an additional challenge because our existing access control system enforces object-level restrictions directly in the queryset. This means unauthorized objects are omitted from list results entirely. We will need to be careful to maintain this pre-existing behavior while migrating. Some queries are easy to migrate, only filtering objects that match a list of specific ids, whereas others are more complicated, joining on access control tables.
+
+### Why now?
+
+As we've scaled up the access coverage over more products, the data warehouse, properties, and new AI products, the current access control model has not scaled well. It has slowed down development, polluted some business logic with authorization code, and become difficult to reason about.
```
benjackwhite, line 323: How does this new service concept avoid this? Not disagreeing with the direction but I didn't see how the new service will avoid the issues listed

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+In this case, we are still spending a lot of time maintaining the code which evaluates the access control rules. Each new use case requires lots of new Python code that intermixes different concerns and is dificult to refactor.
+
+**Build the ReBAC engine ourselves**
+
+If we adopt a standard ReBAC engine, then code examples on the internet makes it easier for LLMs to use. Additionally,
+we want to spend the bulk of our development effort on the authorization rules, not the plumbing that enables that.
+
+### Rollout?
+The python API will control rollout of ReBAC backed access control. Initially, we will port the existing access control code into the access control product folder, and expose it through the new python API. Then, once we start implementing it to be backed by ReBAC, we will set up a flag that determines whether the legacy access control code or the new ReBAC model should be used for authorization evaluation.
```
benjackwhite: This is the bit i'm least sure about. Feels like its not going to work or at least will short term introduce a bunch of lag as the existing system cant scale as well as the new one (otherwise what would be the point of the new one?)

Why not just extend the existing classes to operate with _both_ the old and the new in read only mode so we can validate it works and then have a flagged way to gradually switch between. 

The API bit is confusing (i'm reading API as a running http service) that I think feels most contentious so far

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: @rafaeelaudibert Fair push. I had suggested to put it under its own `products/` folder to keep it isolated from other code so that it doesn't get hacked up and mixed together with other business logic (which is a big problem right now with access control). Putting it under a `product/` folder gives some protection from internal library code getting exposed via the tach CI checks. Happy to put it in a different place if we can setup similar rules there to keep the robots from doing silly things like exposing library internals.

@benjackwhite The python API is meant to be a standardized interface that is better than we have now, prevents devs (specifically agents) from exposing authz internals, and also manages some house-keeping things like activity logging and low-level details of the ReBAC service.

Also, I suspect there may be some confusion here reading another comment of yours. I am not suggesting adding a new microservice with a REST API that proxies requests between the monorepo python code and the standalone ReBAC service. The Python API is python code that monorepo python code can call and the Python API is what makes the request to the ReBAC service. The ReBAC service would be a containerized service whose code we do not maintain.

For example, product analytics makes a function call to the python API to see if the user can create a new insight, the python API sends a network request to OpenFGA to see if the user can create a new insight. Along the way the python API does activity logging and abstracts away some of more low-level details of OpenFGA like constructing object keys and authenticating with OpenFGA.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: Yup, fully agree. Was going to list all of the Django utilities here in the RFC, but the list ended up being really big and I felt like it would have distracted from the point of the RFC by burying it in small implementation details.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: We can run it in a container when running tests. For example, I believe we already do this with postgres, redis, temporal, etc..

OpenFGA, SpiceDB, etc. are distributed as cloud-native container images which should be easy to spin up in a dev environment or test environment.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: We are expecting this to be a multi-month rollout and not something that we switch on instantly. I mention this later in the doc, but plan on shadow logging the results from the new system, and doing a phased gradual rollout once we have built confidence that the new system agrees with the old.

The Python API will help here since it will be the single place that is used for authz checks, rather than having a separate API for old and new. That way call sites are not aware of which implementation is being used.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: Totally fair and something I debated myself before writing the RFC.

Yes, we could refactor to expose a cleaner API for authz checks, but not adopt ReBAC. However In this case while we fix the code icks, we:
1.  Still have to work around the inflexible Python code responsible for turning access control policies into authorization decisions.
2. The python code is very imperative, making it difficult to update.
3. It is focused on a narrow area of authorization, which is the access control feature we provide customers. It doesn't work for answering general authorization questions, which has meant that certain areas like PHAI Tasks have built their own authorization mechanisms.

Adding a ReBAC serivce we would use a highly compact DSL to turn authorization policies into decisions instead of python code. This has some benefits:
1. We broaden the scope of what authorization questions access control can answer
2. We can ship faster, with smaller changes, because adding new authorization policies or modifying existing ones doesn't mean we have to update the underlying engine at the same time.
3. Concerns are better separated which makes it easier to parse our authorization policies. For example, an agent doesn't need to look at a ton of python code which intermixes other concerns like db queries, caching, management apis, etc.. The agent would just need to look at the DSL files containing the authorization rules, which themselves would be quite compact.

tldr: just refactoring to have a better python API would only solve "polluted some business logic with authorization code".

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: Hmmm, maybe I forgot to add this to the RFC, but my intention was to run them both in parallel until we have enough confidence that we can switch over by doing a gradual rollout. 

I will update this section to include:
1. We run them in parallel and shadow log the results of the ReBAC decisions to verify that they align with our existing system.
2. We add an additional flag variant that stops running them in parallel and only uses the existing system. This is so that in case the ReBAC service hits scaling issues it doesn't bring down the existing access control mechanism.

As for the API bit, I am not proposing we introduce a new microservice that the monorepo sends HTTP requests to and then that microservice sends requests elsewhere. By python API, I mean a pure python library that exposes a small set of functions that call the ReBAC service (OpenFGA, SpiceDB, Permify, etc.).

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
feliperalmeida, line 122: This is probably the hardest part of this implementation - how to implement a reliable "dual write" system. The ReBAC server may go down, network issues, etc. What would happen if the ReBAC graph is out-of-sync?

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+    ]
+)
+
+# check if user can access dashboard
+can_access = access_control.check(
+    subject=access_control.Subject(user_id=user.pk),
+    resource=access_control.Resource(type=access_control.ResourceType.Dashboard, id=dashboard.pk),
+    action=access_control.Action.Edit,
+)
+assert can_access is True
```
feliperalmeida, line 208: Here we're assuming `write` was a sync operation, which might be hard to implement in a dual write system.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+            resource=DataWarehouseTableResource(project_id=project.pk, table="postgres.stripe_invoices"), # ex: "project:{project.pk}:data_warehouse_table:postgres.stripe_invoices"
+            action=access_control.Action.View,
+        ),
+    ]
+)
+```
+
+### Gotcha: Django-isms
+
+The mega-brained reader will know that there is a lot of access control checks and enforcement which happens in Django-land. With the new Python API we will re-implement the many Django permission enforcement and utility classes to use the new Python API under the hood, instead of directly evaluating access control rules from the database.
```
feliperalmeida: Will this be integrated to the Django ORM? E.g. when there's a `save` or `delete` operation, we might need to sync to the ReBAC graph. How is this going to happen?

feliperalmeida (commented):
I love this idea and I think this is the way going forward. Thanks @ReeceJones for putting this together.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: A user could be granted permission to access an object they don't have access to, or be denied access to an object they do have access to. In an ideal world we don't have dual writes, and the ReBAC graph is the source of truth for authorization policies. However, that would require deprecating the existing data model which is (probably?) doable and preferred, but beyond the scope of this PR because it would be another big migration. Trying to keep this manageable for now.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: Yeah... there are edge cases here, but I think it will be manageable.

There are system design things we can do here like transactional outbox to ensure OpenFGA writes are wrapped in the Postgres transaction, as well as some implementation specific things depending on the ReBAC service to ensure consistency.

If you want to do more reading here on the consistency side of things I suggest you look into causal consistency tokens (Zanzibar, SpiceDB), query consistency (OpenFGA), and contextual tuples (OpenFGA, others?). Its a gross simplification, but causal consistency tokens are returned by `write()` and ensure that all queries using that token return a view of the state that is at least as up to date as the `write()` command that issued it, and query consistency lets you bypass caching and consistency controls and force the query results to be fully consistent.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: Yeah we'll need to do this. Likely it will be a post-create, post-delete hook that writes something like the following tuples to the ReBAC graph.

`(user:123, dashboard:123, "creator")`
`(project:123, dashboard:123, "parent")`

As hinted before, the actual mechanism for this will likely be a little more complicated than doing this as a sync request. Probably following the transactional outbox pattern + using whatever consistency controls the ReBAC service gives us.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
benjackwhite: That makes more sense - I always prefer SDK over API for this just becuase there is no way of confusing an SDK for a running http service

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
benjackwhite: > ensure OpenFGA writes are wrapped in the Postgres transaction

They support distributed transactions? I think this is a very hard problem that I'd be surprised if they have just solved out of the gate. We should just be aware and move forward but not underestimate this

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
benjackwhite: Maybe I just want to make clear that you aren't underestimating the severity of this system given it will be in the hot path of literally every single API call. 

That's fine but it means the maintainers of this need to treat it that way. The RFC is very solution based which is fine but this would be a foreign system with its own scaling challenges that I would be asking "who maintains this, improves it, monitors it, is on call for it etc". Just want to make sure that is considered as there is a whiff of "infra will handle it" - which they will not ;)

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+This decoupled architecture improves scalability, because changes to the access control rules are contained to the PDP. The PEP treats the PDP as a black box, which reduces the blast radius of access control logic changes. Additionally, modern authorization systems rely on domain specific languages (DSLs) and specialized data structures to model access control rules. This imposes a design constraint that forces developers to model access control as concise formal logic rather than application code.
+
+See [ABAC](https://en.wikipedia.org/wiki/Attribute-based_access_control) and [Cedar](https://docs.cedarpolicy.com/) for further reading on this architecture.
+
+### ReBAC
+
+ReBAC solves the problem of access control by modeling access as a graph. A user can perform an action on an object if they have a relation, direct or indirect, to that object. This is a natural fit for us, because we already model access control through relations.
```
a-lider, line 91: Did a bit of research on the history of Zanzibar. It was built because hundreds of services needed to share access decisions across products (a Drive doc embedded in Gmail, Photos shared into Calendar). Each product had its own version of access control, its own backend and an isolated database. And the graph was way more complex than object -> resource with nested groups and cross-application hierarchies.

In our case all products use Django and one Postgres, one team_id scope. I don't think mapping it into a graph and adding a layer of complexity to sync it is worth it at this stage.

a-lider: What problem would ReBac solve beyond centralizing the access rules?

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
a-lider: I worry this ends up way harder to maintain and extend than what we have now wIth all the scaling problems and infra. But that's one of the main goals here.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+
+### There is no general purpose authorization library
+
+Right now authorization is handled different ways in different contexts, but there is no cohesive throughline. For example, tenant isolation for organization data is handled completely differently than checking if a user is allowed to create new dashboards.
+
+In the age of agents this also means we cannot easily instruct agents to follow a specific authorization pattern. Authorization is so fractured and context-dependent that we cannot reliably tell an agent how to implement authorization when building new features. This has obvious security implications.
+
+### Auditability
+
+Right now if a user is having trouble accessing something, we have no way of easily figuring out why. Our best bet is to dump their access control rules from the database, feed it to an agent, and ask it to figure out why it isn't working. This is because access control evaluation happens in many different places, with different behaviors, rather than a single common code path.
```
a-lider, line 48: Partially solved this one in https://github.com/PostHog/posthog/pull/76045

`UserAccessControl` returns the source of the resolved access now, not just the level. I've only used it for the UI so far (to show somthing like "Based on a member override"), but we can add more logging.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,362 @@
...
+**Cons**
+
+- *More complex architecture* - Requires an additional service deployed to the cluster, with its own data store.
+- *Potential for higher latency* - It is possible that there is more network or computational overhead for each authorization request.
+- *ReBAC graph could become inconsistent with DB state* - We need to make sure CRUD operations to DB objects are synchronized with the ReBAC graph, and we need to develop tools to respond to potential incidents where ReBAC graph state drifts and becomes inconsistent with the DB.
+
+## Design 
+*What are the key user experience and technical design decisions / trade-offs?*
+
+We will adopt a ReBAC framework, then expose a minimal Python API for clients to perform authorization checks. In this case the python API will act as the PDP, and clients will act as the PEP.
```
a-lider: I agree with the direction, and I'm happy to help with decoupling access control. We just don't necessarily need a remote engine for that IMO. We can keep the current data model with `UserAccessControl` as the resolver (still need to think through the read-only MCP and public sharing cases you mentioned, but that seems less complex).

a-lider: I took a stab at it. Here's a draft PR with roughly the API shape you proposed, but with the existing resolver behind it instead of ReBAC: https://github.com/PostHog/posthog/pull/86028

It's `access_control.check(action, resource)` plus `require`, `decide` and `filter`, living in `products/access_control/backend/facade`. 
3 call sites moved over as examples (hogql table access, property restrictions and dashboard copy_tile)

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: This isn't about spinning off authorization into a  standalone service so it is accessible from a non-django context. This is about simplifying authorization

I mention this in the RFC, but ReBAC solves the fundamental problem that we have right now where our authorization model is not extendible to new product areas and does not cover the general concept of authorization. Moving to ReBAC, we make authorization an infrastructure problem rather than a code problem, and move 1-step up so we have greater control and management of authorization policy.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: This only solves it on the user frontend, and it assumes that the code displaying the access level explanation is correctly.

1. It does not give us historical logs to see if a user was granted or denied access to a resource (useful for security incidents, debugging users' past authorization issues for example)
2. We have had quite a few issues where the access control code was incorrectly interpretting access control rules, which conflicted with our docs and what the UI showed. Even after adding excessive regression tests and property based testing, I'm still quite skeptical that we can ever be confident that our logic is formally correct, because issues still occasionally slip through.

Also, this does not address the primary point I was making here: our access control (and authorization) logic is so complex and sprawled throughout the codebase that we have to rely on agents to understand it. It would be much better if the rules were centralized in one place, outside of the python code, so we could read the authorization rules without the needing the context of db queries, biz logic, the access control data model, etc.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: I think the PR is actually a good example why we should do ReBAC.

It has 4 different methods (3 unique if you exclude `check_each`) for determining if a user is authorized to perform an action. It also assumes that the subject will always be a user (not an API key for example). And it doesn't address the broader authorization question. It only focuses on the access control feature (as in configurable rules for paid platform addon customers). 

Plus, it seems to further propagate the weird behavior our our access control system. `_ACTION_RUNG` is a weird hack to map actions to access levels, it still relies on injecting access control into django queries, there is one-off logic for property access control in the facade, etc..

What I'm proposing here isn't a simple "lets rebuild the code because it looks ugly". I want us to take authorization more seriously. I think we should do that by evolving our ownership of "access control" (platform addon feature) to "authorization" (developer api) by thinking of it as infrastructure rather than a feature we offer customers.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: Thanks, I agree. I do think we (platform features) should own it. I have no expectations that infra would be responsible for it beyond obvious things like shared k8s tooling. For performance, I think it is manageable (Google reports p95 latency of 10ms for Zanzibar) to with some work.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: Ah, yeah I see how the terminology could be confusing. I will update the RFC to be more clear about it, and opt to use SDK instead of API for the terminology.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Proposal
+
+### Vision
+
+A cultural shift at PostHog where authorization is a primary concern, and there is one interface that developers across the company use for authorization checks.
+
+### Proposal
+
+I propose we adopt relationship-based access control (ReBAC) and repackage access control as a general purpose authorization library for PostHog developers within the monorepo.
```
yasen-posthog, line 64: > for PostHog developers within the monorepo

I know we talked about an integrated module, but if we go that route we may want to put some m2m auth flows so "external" services (ie billing) can also access that. Or a microservice 🙅

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,368 @@
...
+
+### Success criteria
+
+1. No customer's access is impacted by the change.
+2. All authorization requests are pass through a single interface.
```
yasen-posthog: I'd rephrase as "all auth decisions", cuz requests will most likely still be intercepted by some sort of middleware/mixin? But more important point being that all ReBAC evaluation + per-product business logic is evaluated at a central place.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
yasen-posthog: Plus the corner we've put ourselves in by having this lack of extensibility stops us from **significantly** simplifying our access rules and making them easier to understand AND easier to build on. Essentially we need to work on our foundational layer so when we're done, we have more freedom to more easily build the things that our clients actually want (eg guest mode, temporary access rules, eg access hold).

So a couple of example for policies you write in Cedar for example
```
// 1. editor of this specific dashboard
permit (principal, action == Action::"editDashboard", resource)
when { principal in resource.editors };

// 2. editor for dashboards in this project   (the resource tier)
permit (principal, action == Action::"editDashboard", resource)
when { principal in resource.tier.editors };

// 3. project admin
permit (principal, action == Action::"editDashboard", resource)
when { principal in resource.tier.project.admins };

// 4. org admin
permit (principal, action == Action::"editDashboard", resource)
when { principal in resource.tier.project.org.admins };
```

Thus the rule if an insight should be rendered on a dashboard is expressed like that
```
permit (principal, action == Action::"renderInsight", resource)
when {
  principal in resource.dashboard.viewers ||
  principal in resource.dashboard.editors ||
  principal in resource.dashboard.tier.editors ||
  principal in resource.dashboard.tier.project.members ||
  principal in resource.dashboard.tier.project.admins ||
  principal in resource.dashboard.tier.project.org.admins
};
```

So you essentially have 2 layers:
- A `schema` of how access flows in your system, ie the rules I shared above. We write those in a special file and they are committed to code.
- A set of `grants` that live in the db (similar to what we have today as AC records).

Therefore the call site is super simple, something like
```
verdict = access.can("renderInsight", insight)
```

This decouples PEP from PDP and allows you to change the whole behavior (if you need to) with just a single change to the described schema (eg you decide admins don't have implied access to all insights so we you just remove `principal in resource.dashboard.tier.project.admins` from the policy.

Hope this gives you some concrete idea of what the future might look like as it can be quite an abstract topic where the immediate gains are not super obvious.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,368 @@
...
+
+### Implementations
+
+[Zanzibar](https://storage.googleapis.com/gweb-research2023-media/pubtools/5068.pdf) is Google's in-house ReBAC implementation. It is used across hundreds of their products, including popular products like Google Docs. Zanzibar is the primary source of inspiration for nearly all modern ReBAC implementations.
+
+[OpenFGA](https://openfga.dev/) is a ReBAC implementation that is a CNCF incubation project, maintained by Okta and Grafana. It seems to be the most popular ReBAC implementation right now due to its cloud-native approach.
+
+[SpiceDB](https://authzed.com/spicedb) is an authorization database with support for ReBAC configuration. It is [open source](https://github.com/authzed/spicedb), under the Apache 2.0 License, and maintained by [AuthZed](https://authzed.com) who sell authorization as a service.
+
+[Permify](https://permify.co/) is an [open source](https://github.com/Permify/permify) ReBAC implementation that, like OpenFGA, aims to be cloud-native. It is  licensed under the APGL-3 license and maintained by [FusionAuth](https://fusionauth.io/) who sell authorization as a service.
```
yasen-posthog, line 103: I'd also add Cedar + its py bindings `cedar-py`. Allows us to bundle it in the main repo.

OpenFGA looked great but it's a closed Go binary and designed to be used as a separate microservice (at least that's what i gathered from my research).

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
yasen-posthog: I agree having this as a separate service will make things too complex for us and I have a feeling we need to bake V2 into the existing monorepo, because otherwise we'll have not one hard challenge (changing the access paradigm) but two (paradigm + infra and ensuring ~99.99% uptime).

We can still abstract access to eg Cedar-py via an SDK, so we can futher down the line think about switching over form a local evaluation to the remote (service) eval.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,368 @@
...
+# Enumerate resources a user can access
+accessible_resources = authorization.list_resources(subject=..., resource=..., action=...)
+
+# Enumerate users who can access a resource
+authorized_users = authorization.list_subjects(resource=..., action=..., subject=...)
+
+# Generic utility for upserting and deleting relationships in bulk
+authorization.write(
+    [
+        authorization.AddRelationshipRequest(subject=..., resource=..., action=...),
```
yasen-posthog, line 159: I believe the accepted term is `grants`, not relationships?

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,368 @@
...
+```
+
+**For example, here are what some common access control operations might look like using this SDK.**
+
+```python
+from posthog import authorization
+
+# check if user can access dashboard
+can_access = authorization.check(
+    subject=authorization.Subject(user_id=user.pk), # ex: "user:{user.pk}"
```
yasen-posthog, line 180: Can we think about making the current user implicit unless specified? It will be the same user.pk 99% of time and it will make the checks more succinct.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,368 @@
...
+
+This flag will have multiple variants: `off`, `shadow-log`, `rebac`. In `off`, only the existing Python logic is used. In `shadow-log` the existing Python is used as the authoritative answer, but a request is made to the ReBAC service in parallel and that result is logged against the result returned by the Python implementation. In `rebac`, the ReBAC service's response is used as the authoritative answer, and the Python implementation's response is logged.
+
+### What about the data model?
+
+In the first iteration we will maintain the existing data model by implementing a dual write system for object permissions. However, once the migration is complete we may re-evaluate this and opt to use the ReBAC service as the canonical source of truth for authorization policies, including those that we display in the UI and allow customers to configure.
```
yasen-posthog, line 347: Is the `dual write` you suggest something like a shadow write to the ReBAC side at the point where a client changes their (v1) AC records?

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
yasen-posthog, line 1: Great first push! I left a few comments, I think the most important being the need to try and contain v2 within the scope of the existing monorepo.

Also, I like this a solution-oriented RFC but I also think we need to think through all the conceptual conundrums we have (sharing, org owner/admin levels, non-linear access levels, etc) and see how these will play out using ReBAC and try to foresee any complications. A good example is the one you pointed with implied query filtering in list ops.

I do have a wishlist of features for V2 and it would be nice to see if these will be viable. Sharing my rambling/notes directly:

* It needs to be able to treat agents like users. Actually maybe using the current querying user as the default but clamping down by the AgentUser permissions?
* Should be able to visualize clearly what each member/role access there is, with flow diagrams in a visual way.
* Should be able to take in dynamically introduced runtime temporary restrictions
    * Eg if you suspect someone of wrongdoing, you can do read-only access or no access for 24 hours (automatically imposed/automatically expiring)
* Should be able to take new levels seamlessly and the levels should not be linearly ordered
    * Eg sharing doesnt fit with the no access/read/write model, but we need to have it in place.
* Should be able to express org owner/admin as AC rules. As a consequence there should be an internal non-entitlement-gated bit for free and low-tier orgs.
* Should be represented by the same models that we have in place? Or completely new ones and we switchover?
* Showing before/after when you're changing access rules. E.g if I remove this AC rule from Alice, what are the access implications for her (preview before save).
* We need to be able to detect drift between policy definitions (in-code) and grants (in-db).

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: Cedar is not ReBAC, so I think that is a separate proposal. Cedar is more suited for RBAC and ABAC patterns.

Our code already follows a ReBAC-like paradigm, so I think it is actually quite a bit more complex to adopt Cedar  that you may expect, because we will need to still manage all of the data fetching, attribute building, and push that into the Cedar context.

A separate service is no free lunch, but it is more aligned with our current authorization paradigm, and adopted successfully, simplifies the architecture since we don't need to manage as much scaffolding code to rebuild the authorization model for each request.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: m2m auth would be a good thing to add down the line, but I don't see the need for a microservice. Primarily because the django monolith is the primary consumer of authorization logic, and we don't expect this to change.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: I've added Cedar so that we can evaluate it during the PoC phase. However, I don't think that Cedar is without tradeoffs, as the complexity is shifted to building the Cedar entities, which could involve fetching a lot of entities from the database.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: ReBAC uses the term `relationships`

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: This is a rough sketch of what the SDK will look like not the final version, so sure.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: Yes. When a client modifies their AC rules, we do one write to our database modifying, and another write to the ReBAC service.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
ReeceJones: > It needs to be able to treat agents like users. Actually maybe using the current querying user as the default but clamping down by the AgentUser permissions?

This is why subject is not defined as a user. Subject could be a m2m token, project, impersonated user, api token, agent-scoped oauth token, etc.

> Should be able to visualize clearly what each member/role access there is, with flow diagrams in a visual way.

This is what the `authz.explain(...) function does

> Should be able to take in dynamically introduced runtime temporary restrictions

The implementations mentioned support contextual policies like this.

> Should be able to take new levels seamlessly and the levels should not be linearly ordered

ReBAC and really any sort of authorization policy engine solves this by resolving access level using rules, not by doing the integer comparison like we do.

> Should be able to express org owner/admin as AC rules

I agree. This is a goal of mine. Especially since it feels like this authorization right now is mostly enforced through static code analysis (semgrep IDOR checks). 

> Showing before/after when you're changing access rules

I think this is somewhat out of scope, and I'm a little hesitant about it. Specifically, because it means iterating over all org/project-scoped resources and evaluating the authorization rules against them. In any possible implementation I think this would be hard and costly to maintain.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,379 @@
...
+
+In the age of agents this also means we cannot easily instruct agents to follow a specific authorization pattern. Authorization is so fractured and context-dependent that we cannot reliably tell an agent how to implement authorization when building new features. This has obvious security implications.
+
+### Auditability
+
+Right now if a user is having trouble accessing something, we have no way of easily figuring out why. Our best bet is to dump their access control rules from the database, feed it to an agent, and ask it to figure out why it isn't working. This is because access control evaluation happens in many different places, with different behaviors, rather than a single common code path.
+
+Additionally, we do not have an audit trail that documents access control decisions. This means we cannot retroactively determine if a user was granted or denied access to a resource, let alone understand why. As we and our customers continue to scale, this will be important to document for security incident response.
```
Piccirello, line 50: Huge +1 on having an audit log to replay decisions later

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,379 @@
...
+
+For example, a viewset which returns `Dashboards` would be a PEP. The viewset would ask the PDP if the authenticated user can perform the API action. The PDP returns a boolean response, yes or no, and if denied, the PEP is responsible for terminating the request and responding with the correct HTTP response.
+
+This decoupled architecture improves scalability, because changes to the access control rules are contained to the PDP. The PEP treats the PDP as a black box, which reduces the blast radius of access control logic changes. Additionally, modern authorization systems rely on domain specific languages (DSLs) and specialized data structures to model access control rules. This imposes a design constraint that forces developers to model access control as concise formal logic rather than application code.
+
+See [ABAC](https://en.wikipedia.org/wiki/Attribute-based_authorization) and [Cedar](https://docs.cedarpolicy.com/) for further reading on this architecture.
```
Piccirello, line 87: Another good high level article on the terminology https://idpro.org/authorization-terminology-is-a-mess-lets-fix-it/

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
```diff
@@ -0,0 +1,379 @@
...
+
+As we've scaled up the access coverage over more products, the data warehouse, properties, and new AI products, the current access control model has not scaled well. It has slowed down development, polluted some business logic with authorization code, and become difficult to reason about.
+
+### Alternatives considered
+
+**ABAC**
+
+We already have lots of hierarchy in our rules: project admin overrides, creator escape hatch, object-level overrides, roles, etc. that are all dependent on their relationships between each other. ABAC is better suited for flat access control structures that are dependent on inline object attributes. For us, ReBAC is a better fit since it directly models relations.
+
+**Move existing code behind new Python SDK, but don't adopt ReBAC**
```
Piccirello, line 331: Given the shared concerns around the size of the changes proposed in this RFC, is there a world in which a new python SDK is introduced first, and then ReBAC is added later? That way all logic can be centralized (and maybe all decision making moved to a separate service) before cutting over to a new authZ system.

Edit: I see that this is explicitly called out in the Rollout section below. Perhaps this isn't clear enough? To me it solves some of the concerns shared around the complexity of this migration. And if the ReBAC piece falters for some reason, having all authorization logic colocated would still be a huge win.

Piccirello (commented):
This would be a massive win for reducing developer error as our product changes and for our ability to investigate during an incident.

requests-for-comments/2026-08-13-adopt-relationship-based-access-control.md
yasen-posthog: Yeh it kinda boils down to what we choose on the complexity and stability axes. Although Cedar is not ReBAC per-se (they mention that they cover it in their docs) it allows us to reuse our existing db records/models... the downside being the [db store] -> [policy engine] boilerplate and fetching we need to do. However I see this as a way to have a shadow implementation of v2 (without the dual-write) which we can run in parallel with v1 in order to detect issues, before we go live. On the other hand it's the more conservative approach and long term we'll most probably end up with a separate service anyway, if we want a full-blown ReBAC system and want to completely decouple the monolith from any access decisions. So on that axes this is complexity 5 (this is the big ?, is it 5 or more) and stability 10 I'd say.

On the other hand there is the service OpenFGA (or similar) route which has the dual-write and availability questions to answer. I see that as the better architectural end game, but the complexity with actually running and supporting a separate service is something like 8 and stability is like 4, cuz we're bound to have some downtime until we iron out the whole thing in production.

I'd keep this topic open for discussion at the offsite, but I think we need to be realistic about the risks.

